PR #39, opened 2019-11-21
Support returning pandas dataframe from sql exec

Shelnutt2:
If the user does not specify a output array and the query produces results we now will return the results in a json array to the user. By default this json bytes will be converted to a pandas dataframe.
    
The goal is to allow returning relatively smaller result sets, such as an aggregate without having to store in a output array.

This adds pandas as a new dependency.


Shelnutt2:
> I don't know if there are any limits to pandas json serialization, but this seems neat.

On a size front, uses should not use this for large result sets, performance will be bad even if pandas allows it. Other limitations might exist in pandas but this works well for now. Also I've added an option of `raw_results` to allow returning the json  bytes directly in case a user has some query which pandas is not working for.

This relies on a server side change that is pending and under review. however it is saw to merge now because the status code of 200 (HTTP_OK) isn't currently returned by the server so the pandas code path will not be hit in production currently.

PR #47, opened 2019-12-14
Implement multi-range apply

ihnorton:


tiledb/cloud/cloudarray.py
```diff
@@ -144,3 +163,32 @@ def apply(self, func, subarray, attrs=None, layout=None):
...
         return cloudpickle.loads(res)
+
+    def _calculate_ranges(self, subarray):
+        ranges = list()
+        if isinstance(subarray, MultiIndexRanges):
+            for dim_idx, dim_rng in enumerate(subarray.ranges):
+                dim = self.schema.domain.dim(dim_idx)
+                for idx, rng in enumerate(dim_rng):
+                    start = build_dimension_coordinate(dim.dtype, rng[0])
+                    end = build_dimension_coordinate(dim.dtype, rng[1] + 1)
```
Shelnutt2: Should we be adding a +1 here?

ihnorton: nope, thanks -- left over from testing on prod when it was endpoint-exclusive

ihnorton: adding some tests for apply now

ihnorton:
@stavrospapadopoulos please see the simplified syntax in the new tests, here:

https://github.com/TileDB-Inc/TileDB-Cloud-Py/blob/44eadefcf6d181ebf414a3900699a51056dc494c/tests/test_basic.py#L31-L41

PR #53, opened 2020-01-03
Add retry for urllib3 for 503s

Shelnutt2:
This will enable automatic retries for cases where the server doesn't
have enough resources.

We retry 10 times for a total of 127 seconds. 127 seconds was selected
because in worst case scenarios it takes up to 2 minutes to launch resources from cold starts

Backoff retry times (seconds):
0.25
0.5
1.0
2.0
4.0
8.0
16.0
32.0
64.0

### DIFF
--- a/tiledb/cloud/config.py
+++ b/tiledb/cloud/config.py
@@ -2,6 +2,7 @@ import os.path, sys
 from pathlib import Path
 import json
 from . import rest_api
+from urllib3 import Retry
 
 config = rest_api.configuration.Configuration()
 default_config_file = Path.joinpath(Path.home(), ".tiledb", "cloud.json")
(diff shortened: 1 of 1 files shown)


ihnorton:
Is there any output of what is happening during the retries? I would be surprised to have a simple call hang for >2 min without any indication that retries are happening in the background.

Shelnutt2:
> Is there any output of what is happening during the retries? I would be surprised to have a simple call hang for >2 min without any indication that retries are happening in the background.

No there is no output. I don't think the retry functionality has the ability for us to make it print/callback in between retries. I agree the silent hanging is not ideal but I don't have a better solution unless we role our own retry.

ihnorton:
Added story in CH for now.

PR #55, opened 2020-01-16
Add support for compressed zlib results (only) from apply

ihnorton:


ihnorton:
woot mac CI is fast now w/ wheels 😄 (except 3.5, I'll probably turn off mac CI there)

PR #60, opened 2020-02-08
Async API for sql exec and udf apply

Shelnutt2:
Add async functions and api for sql/udfs

New functions are `tiledb.cloud.sql.exec_async` and  `tiledb.cloud.CloudArray.apply_async`

This reworks the client to use one global object for storing the api classes. This is needed to avoid error with deleting the code generated api clients (i.e. `rest_api.SqlApi`) when an async request is still going.

This also adds new wrapper classes of `SQLResult` and `UDFResult` which implements a single function `get`, so we can handle the deserialization. The classes are derived from the multiprocessing.pool.ApplyResult, which is what the code generated async request returns.

TODOs:
-    [x] Support setting worker pool size
-    [x] Support async UDFs

Example usage:

```
result = tiledb.cloud.sql.exec_async(query="select * from `tiledb://TileDB-Inc/quickstart_dense`")
print(result.get())
```

```
def median(df):
    import numpy
    return numpy.median(df["a"])

with tiledb.DenseArray("tiledb://TileDB-Inc/quickstart_dense", ctx=tiledb.cloud.Ctx()) as A:
  res = A.apply_async(median, [(1,4), (1,4)])

print(res.get())
```

Users can set the urllib3/request worker pool size with `tiledb.cloud.config.config.connection_pool_maxsize=x`. Default pool size is `multiprocessing.cpu_count() * 5`

### DIFF
--- a/tiledb/cloud/array.py
+++ b/tiledb/cloud/array.py
@@ -1,12 +1,50 @@
-import urllib
 from . import rest_api
 from . import config
 from . import client
-from . import tasks
 from . import cloudarray
 from . import tiledb_cloud_error
 from .rest_api import ApiException as GenApiException
 from .rest_api import rest
+
+import zlib
+import multiprocessing
+import cloudpickle
+
+
+class UDFResult(multiprocessing.pool.ApplyResult):
+    def __init__(self, response):
+        self.response = response
+
+    def get(self, timeout=None):
+        try:
+            response = rest.RESTResponse(self.response.get(timeout=timeout))
+
+            global last_udf_task_id
+            last_udf_task_id = response.getheader(client.TASK_ID_HEADER)
+
+            res = response.data
+
+        except GenApiException as exc:
+            raise tiledb_cloud_error.check_udf_exc(exc) from None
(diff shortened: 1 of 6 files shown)


Shelnutt2:
@ihnorton I'm putting this up to get initial feedback on the design and just using the code generated apis async request capabilities.

tiledb/cloud/array.py
```diff
@@ -1,13 +1,49 @@
...
+    def get(self, timeout=None):
+        try:
+            response = rest.RESTResponse(self.response.get(timeout=timeout))
+
+            global last_udf_task_id
+            last_udf_task_id = response.getheader(client.TASK_ID_HEADER)
+
+            res = response.data
+
+        except GenApiException as exc:
```
ihnorton: `response.get` will raise `multiprocessing.TimeoutError` if it fails, so need to check that here as well.

Shelnutt2:
@ihnorton this is the final design based on your feedback and our conversations. It looks like the osx python 3.5 is failing with a wheels problem that is unrelated.

PR #61, opened 2020-02-13
Support new function for generic UDFs

Shelnutt2:
Support new function for generic UDFs, will be rebased once #60 is merged.  Docs will be updated once we cut a new release.

```
import tiledb.cloud

def test(msg="hello world"):
  return msg

# namespace optional just like sql.exec
tiledb.cloud.udf.exec(func=test, argument="new parameter", namespace="demo")

result = tiledb.cloud.udf.exec_async(func=test, argument="new parameter")
print(result.get())
```


ihnorton:
LGTM. Linux failures look like they are due to backend not deployed? We don't have wheels for macOS on Python 3.5, so I think that matrix should just be removed.

Shelnutt2:
> LGTM. Linux failures look like they are due to backend changes not deployed? We don't have wheels for macOS on Python 3.5, so I think that matrix should just be removed.

New pr is up to remove osx python3.5 from testing, #63 . The failures are due to backend not being deployed. Once all the PRs are approved everywhere I'll deploy it out. I won't merge this one until the tests pass.

PR #62, opened 2020-02-13
Add support for apply on array by URI

Shelnutt2:
New functions of `tiledb.cloud.array.apply` and `tiledb.cloud.array.apply_async` are added to allow applying a udf to an array without opening it. The `CloudArray.apply_async` has been updated to use these new functions. We had to make some minor protocol tweaks to avoid needing the dimension datatypes.

~Will rebase after #61 is merged.~


tiledb/cloud/array.py
```diff
@@ -182,3 +186,177 @@ def array_activity(uri):
...
+
+    **Example**
+    >>> import tiledb, tiledb.cloud, numpy
+    >>> def median(df):
+    ...   return numpy.median(df["a"])
+    >>> # Open the array then run the UDF
+    >>> tiledb.cloud.array.apply_async("tiledb://TileDB-Inc/quickstart_dense", median, [(0,5), (0,5)], attrs=["a", "b", "c"]).get()
+    2.0
+
+    :param func: user function to run
```
ihnorton: Params are usually before examples (not sure if the docstring parser will even work in this order)

tiledb/cloud/array.py
```diff
@@ -182,3 +186,177 @@ def array_activity(uri):
...
+    """
+
+    (namespace, array_name) = split_uri(uri)
+    api_instance = client.client.udf_api
+
+    if not callable(func):
+        raise TypeError("First argument to `apply` must be callable!")
+
+    pickledUDF = cloudpickle.dumps(func, protocol=udf.tiledb_cloud_protocol)
+    pickledUDF = base64.b64encode(pickledUDF).decode("ascii")
```
ihnorton: I probably asked about this before, but it still looks weird (call `.decode('ascii')`) -- is this because the generated API doesn't accept bytes?

Shelnutt2: Yes unfortunately it does not accept bytes, somewhere in the middle it tries to perform some operations which isn't supported on bytes. I don't have the exact details at the moment but can reproduce if you want.

tiledb/cloud/array.py
Shelnutt2: I will update this and a few other places in a separate PR now

tiledb/cloud/array.py
ihnorton: No need, understood.

PR #64, opened 2020-02-14
Add support for registering a UDF and invocating by name

Shelnutt2:
This adds support for registering UDFs for later invocation. A handful of new functions are added to facilitate registering, updating, listing and running registered udfs.

Example new usage is:
```
# Register
tiledb.cloud.udf.register_generic_udf(func=load_csv_sparse, name="sparse_csv_ingestor", namespace="demo")

# call exec by registered name instead of by passing func
res = tiledb.cloud.udf.exec(name="demo/sparse_csv_ingestor", arguments=params)
```

will rebase after #62 is merged.


tiledb/cloud/array.py
```diff
@@ -277,11 +278,15 @@ def apply_async(
     (namespace, array_name) = split_uri(uri)
     api_instance = client.client.udf_api
 
-    if not callable(func):
+    if func is not None and not callable(func):
         raise TypeError("First argument to `apply` must be callable!")
```
ihnorton: "`func` argument to ..."

tiledb/cloud/udf.py
```diff
@@ -98,8 +114,288 @@ def exec(
...
+        pickledUDF = cloudpickle.dumps(func, protocol=tiledb_cloud_protocol)
+        pickledUDF = base64.b64encode(pickledUDF).decode("ascii")
+
+        udf_model = rest_api.models.UDFRegistration(
+            name=name,
+            language=rest_api.models.UDFLanguage.PYTHON,
+            version="{}.{}.{}".format(
+                sys.version_info.major, sys.version_info.minor, sys.version_info.micro,
+            ),
+            image_name=image_name,
```
ihnorton: If `image_name` is None does it just use the default for `UDFLanguage.PYTHON`?

tiledb/cloud/rest_api/test/test_udf_api.py
```diff
@@ -28,6 +28,36 @@ def setUp(self):
     def tearDown(self):
         pass
 
+    def test_get_ud_fs(self):
```
ihnorton: `ud_fs` -> `udfs`?

tiledb/cloud/udf.py
Shelnutt2: Yes, server side None/empty string is converted to the string "default" which we then parse are default for the language and language version.

tiledb/cloud/rest_api/test/test_udf_api.py
Shelnutt2: This is from the autogenerated code, there is something funky in the parse of UDFs, it converts it to `udf_fs` instead of `udfs` or `u_d_f_s`.

PR #66, opened 2020-02-20
Add ability to save raw text of udf

Shelnutt2:
Add ability to save raw text of udf. This is disableable with the do_not_include_source_lines parameter.

### DIFF
--- a/tiledb/cloud/array.py
+++ b/tiledb/cloud/array.py
@@ -6,6 +6,7 @@ from . import tiledb_cloud_error
 from .rest_api import ApiException as GenApiException
 from .rest_api import rest
 from . import udf
+from . import utils
 
 import zlib
 import multiprocessing
@@ -254,6 +255,7 @@ def apply_async(
     layout=None,
     image_name=None,
     http_compressor="deflate",
+    include_source_lines=True,
 ):
     """
     Apply a user defined function to an array asynchronous
@@ -264,6 +266,7 @@ def apply_async(
     :param layout: tiledb query layout
     :param image_name: udf image name to use, useful for testing beta features
     :param http_compressor: set http compressor for results
+    :param include_source_lines: disables sending sources lines of function along with udf
     :return: UDFResult object which is a future containing the results of the UDF
 
     **Example**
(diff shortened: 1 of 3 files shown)


tiledb/cloud/array.py
```diff
@@ -254,6 +255,7 @@ def apply_async(
     layout=None,
     image_name=None,
     http_compressor="deflate",
+    do_not_include_source_lines=False,
```
ihnorton: -> `include_source_lines=True` and avoid the double negative below?

tiledb/cloud/array.py
```diff
@@ -284,10 +287,23 @@ def apply_async(
         raise TypeError("name argument to `apply` must be set if no function is passed")
 
     pickledUDF = None
+    source_lines = None
     if func is not None:
         pickledUDF = cloudpickle.dumps(func, protocol=udf.tiledb_cloud_protocol)
         pickledUDF = base64.b64encode(pickledUDF).decode("ascii")
 
+        if not do_not_include_source_lines:
```
ihnorton: (change as suggested in :258?)

tiledb/cloud/udf.py
```diff
@@ -232,6 +294,19 @@ def update_udf(func, name, namespace=None, image_name=None, type=None):
         pickledUDF = cloudpickle.dumps(func, protocol=tiledb_cloud_protocol)
         pickledUDF = base64.b64encode(pickledUDF).decode("ascii")
 
+        source_lines = None
+        if not do_not_include_source_lines:
```
ihnorton: 4th call I think -- suggest to make this a function?

Shelnutt2:
@ihnorton I've made the suggested changes, please let me know if you like it better. The function was a good idea, was repeating the same lines in too many places.

tiledb/cloud/utils.py
```diff
@@ -0,0 +1,24 @@
+import inspect
+import logging
+
+logger = logging.getLogger("tiledb.cloud")
+
+
+def getsourcelines(func):
+    if func is not None and not callable(func):
```
ihnorton: oh, btw this can be written as `if not callable(func)` (`callable(None)` -> `False`)

PR #72, opened 2020-04-16
Add support for rendering visualization graph

Shelnutt2:
Two tree formats are supported, one using pydot if the user has dot/graphviz available.
The second using a python method to compute the layout.

Graphs are build using plotly and can use normal figures or widgets if inside a notebook.

Graph figures can be auto-updated in realtime using callback on the status.

A future PR will introduce support for our own `TileDB-Jupyter-Plot-Widget`, for now we use plotly.

### DIFF
--- a/tiledb/cloud/dag/visualization.py
+++ b/tiledb/cloud/dag/visualization.py
@@ -1,0 +1,231 @@
+import networkx as nx
+import random
+
+from . import dag
+
+
+def build_graph_node_details(nodes):
+    """
+    :param nodes: List of nodes to get status of
+    :return: tuple of node_colors and node_text
+    """
+    # Loop over statuses to set color and label.
+    # If you rerun this cell and the one below you can update the graph
+    node_colors = []
+    node_text = []
+    for node in nodes:
+        status = node.status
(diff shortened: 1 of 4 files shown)


tiledb/cloud/dag/visualization.py
```diff
@@ -0,0 +1,230 @@
...
+            text=node_text,
+            selector=dict(mode="markers"),
+        )
+
+
+def hierarchy_pos(
+    G, root=None, width=1.0, vert_gap=0.2, vert_loc=0, leaf_vs_root_factor=0.5
+):
+    """
+    Taken from https://epidemicsonnetworks.readthedocs.io/en/latest/_modules/EoN/auxiliary.html#hierarchy_pos
```
ihnorton, line 59: Should explicitly mention that it is MIT licensed for future clarity: https://epidemicsonnetworks.readthedocs.io/en/latest/_downloads/8e9c8138fef49ddba8102fa7799c29d7/license.txt

Shelnutt2: Done, thanks for point it out.

PR #89, opened 2020-09-17
Support setting namespace for all tasks in DAG

Shelnutt2:
A DAG can now override the namespace a node/task is run in. The override can happen in a Delayed object or in the DAG directly. This allows easier setting of the namespace all tasks should run in.

This also fixes a minor issue with overriding namespaces in the `Delayed` API itself.

Example usage:

```
from tiledb.cloud.compute import Delayed
import numpy

# Wrap numpy median in a delayed object
x = Delayed(numpy.median, name="x")

# It can be called like a normal function to set the parameters
# Note at this point the function does not get executed since it
# is delayed type
x([1,2,3,4,5])

y = Delayed(numpy.median, name="y")
y([1, 2, x])

# To force execution and get the result call `compute()`
y.compute(namespace="seth")
```

Also basic namespace in a `Delayed` object:
```
from tiledb.cloud.compute import Delayed
import numpy

# Wrap numpy median in a delayed object
x = Delayed(numpy.median, name="x", namespace="seth")

# It can be called like a normal function to set the parameters
# Note at this point the function does not get executed since it
# is delayed type
x([1,2,3,4,5])

# To force execution and get the result call `compute()`
x.compute()
```

### DIFF
--- a/tiledb/cloud/compute/delayed.py
+++ b/tiledb/cloud/compute/delayed.py
@@ -8,9 +8,11 @@ import numbers
 
 
 class DelayedBase(Node):
-    def __init__(self, func, *args, name=None, dag=None, **kwargs):
+    def __init__(self, func, *args, name=None, dag=None, local_mode=False, **kwargs):
         self.timeout = None
-        super().__init__(func, *args, name=name, dag=dag, **kwargs)
+        super().__init__(
+            func, *args, name=name, dag=dag, local_mode=local_mode, **kwargs
+        )
 
     def set_timeout(self, timeout):
         if timeout is not None and not isinstance(timeout, numbers.Number):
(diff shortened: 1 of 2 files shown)


tiledb/cloud/compute/delayed.py
```diff
@@ -69,26 +75,26 @@ def visualize(self, notebook=True, auto_update=True, force_plotly=True):
...
-        self.local = local
         if func_exec is not None and not callable(func_exec):
             raise TypeError("func_exec argument to `Node` must be callable!")
 
         if not local:
-            super().__init__(udf_exec, func_exec, *args, **kwargs)
+            super().__init__(udf_exec, func_exec, *args, local_mode=local, **kwargs)
         else:
-            super().__init__(func_exec, *args, **kwargs)
+            super().__init__(func_exec, *args, local_mode=local ** kwargs)
```
ihnorton, line 84: missing comma

ihnorton (approved):
CI failure cause is noted in comment, otherwise LGTM

ihnorton:
Hmm, actually mac should have failed too but passed here?

ihnorton:
Oh, nevermind the tests section is only enabled on linux.

PR #92, opened 2020-09-30
Support s3 renaming of notebooks

antalakas:


tiledb/cloud/notebook.py
```diff
@@ -0,0 +1,46 @@
+from . import rest_api
+from . import client
+from . import tiledb_cloud_error
+from .rest_api import ApiException as GenApiException
+from .rest_api import rest
+
+import urllib
+
+def split_uri(uri):
```
Shelnutt2, line 9: Don't we have this in another package?

antalakas: We import it from array now

Shelnutt2 (approved):
Looks like you just need to format with black then LGTM

PR #95, opened 2020-10-04
 Add support for async_req to all functions 

Shelnutt2:
Now any function can be run as an async request by passing the `async_req` parameter. This allows clients to perform listing and other api requests async if they wish. Default is still for synchronous requests.


antalakas (approved):
lgtm

PR #97, opened 2020-10-08
Set APIClient threadpool default to 2*num_cores

Shelnutt2:
For async support right now we were only running one at a time. This was an oversight because the APIClient overrides the multiprocess pool's default of all cores. So we have to set the APIClient's thread_pool size to all cores.

`APIClient`'s setting of 1:
https://github.com/TileDB-Inc/TileDB-Cloud-Py/blob/aea645215d1113713ad60ea08801fd39c043c912/tiledb/cloud/rest_api/api_client.py#L72


antalakas (approved):

ihnorton (commented):
💯

PR #103, opened 2020-10-23
Addition of Pylint Checks to Azure CI

nguyenv:


nguyenv:
It is current failing on Azure CI for reasons unrelated to this change. There is an issue with too many open SQL connections on Linux.

PR #106, opened 2020-11-06
Enable activation of UDF V1

antalakas:


tiledb/cloud/array.py
```diff
@@ -408,9 +408,6 @@ def apply_async(
         if http_compressor is not None:
             kwargs["accept_encoding"] = http_compressor
 
-        if v2:
-            kwargs["v2"] = True
```
ihnorton, line 412: I think we still need to add the argument into the kwargs so it is sent to the server. You could do `kwargs["v2"] = v2` unconditionally here, or add it to [line 407](https://github.com/TileDB-Inc/TileDB-Cloud-Py/pull/106/files#diff-44a1d24bf2f22228aac1de6886859f11da5a92622de7c2b1575c4d53c5c1a38fR407):

```
kwargs = {"_preload_content": False, "async_req": True, "v2": v2}
```

antalakas: Done, thanks

ihnorton (approved):

PR #107, opened 2020-11-12
Use layout=None if unset; add translation for 'U'->'unordered'

ihnorton:
I believe this also needs server-side change (per @Shelnutt2). Towards addressing this error:
```
"Error submitting query: [TileDB::CellSlabIter] Error: Unsupported subarray layout; the iterator supports only row-major and column-major layouts"}
```

The second commit is the real one, first is just reformatting.


ihnorton:
(I just sync'd my `black` version with the one on CI, there were apparently some rule differences from 19->20)

PR #112, opened 2020-12-02
We now require urllib3 >= 1.26

Shelnutt2:
We now require urllib3 >= 1.26


ihnorton (approved):
Is this due to the API regeneration?

nguyenv (approved):

Shelnutt2:
> Is this due to the API regeneration?

Yes it is.

PR #113, opened 2020-12-09
Align with array wrapped UDFs

antalakas:


tiledb/cloud/udf.py
```diff
@@ -436,95 +436,3 @@ def info(namespace=None, name=None, async_req=False):
...
-        api_instance = client.client.udf_api
-
-        return api_instance.get_udf_info_list(
-            namespace=namespace, search=search, async_req=async_req
-        )
-    except GenApiException as exc:
-        raise tiledb_cloud_error.check_exc(exc) from None
-
-
-def share(name=None, namespace=None, async_req=False):
```
Shelnutt2, line 459: I thought we left the share api intact?

antalakas: We removed it expecting the user to share the array right?

antalakas: Restored

Shelnutt2 (approved):

PR #114, opened 2020-12-09
Add support for datetime objects in UDF ranges

Shelnutt2:
This now supports converting UDF ranges from datetimes to epoch based int64s


ihnorton (approved):
Maybe throw an error if the types are mismatched? (or can they also be `None`?). Otherwise LGTM.

nguyenv (approved):

PR #116, opened 2020-12-17
Add delete for registered udfs and optional update fields

antalakas:


tiledb/cloud/udf.py
```diff
@@ -515,3 +533,38 @@ def unshare(name=None, namespace=None, async_req=False):
...
     except GenApiException as exc:
         raise tiledb_cloud_error.check_exc(exc) from None
+
+
+"""
+Delete a registered udf
+"""
+
+
+def delete(name=None, namespace=None, async_req=False):
```
Shelnutt2, line 543: Lets force the `name` and `namespace` to be set. Don't set it to `=None` in the function definition

Shelnutt2 (changes requested):

Shelnutt2 (approved):

PR #121, opened 2021-02-04
Retry support for Delayed tasks

gsakkis:
Draft proposal for transparent retrying of Delayed tasks. The actual retrying logic (when to stop, how long to wait between retries, etc.) is provided by [tenacity](https://tenacity.readthedocs.io/en/latest/). Docs and tests to be added if there are no objections to the idea.

Usage:
```py
import logging
import random
import time
from tiledb.cloud.compute import Delayed

from tenacity import (
    Retrying,
    wait_fixed,
    stop_after_attempt,
    before_log,
    before_sleep_log,
)


logging.basicConfig(level=logging.INFO, format="%(asctime)s: %(message)s")


def f(x, p):
    v = random.random()
    if v < p:
        raise ValueError(f"Value less than {p}: {v}")
    time.sleep(v * 10)
    return v ** x


retrying = Retrying(
    stop=stop_after_attempt(3),
    wait=wait_fixed(2),
    before=before_log(logging, logging.INFO),
    before_sleep=before_sleep_log(logging, logging.INFO),
)

nodes = [Delayed(f, name=f"node_{i}", retrying=retrying)(i, 0.4) for i in range(3)]
print(Delayed(sum, local=True)(nodes).compute())
```
Sample output:
```
2021-02-04 14:07:42,161: Starting call to 'node_1', this is the 1st time calling it.
2021-02-04 14:07:42,171: Starting call to 'node_2', this is the 1st time calling it.
2021-02-04 14:07:42,266: Starting call to 'node_0', this is the 1st time calling it.
2021-02-04 14:07:47,219: Retrying node_0 in 2.0 seconds as it raised TileDBCloudError: Docker logs: Running UDF.
Error executing UDF:
------------------- UDF Backtrace -------------------
  File "/opt/python_generic_udf_entry.py", line 486, in main
    result = udf_function(*(udf_args_parsed[0]))
  File "retry_demo.py", line 21, in f
ValueError: Value less than 0.4: 0.23205460161491887

------------------- End Backtrace -------------------
Error(s): docker container exited with non-zero code: 1. - Code: 4010.
2021-02-04 14:07:49,221: Starting call to 'node_0', this is the 2nd time calling it.
2021-02-04 14:07:54,510: Retrying node_0 in 2.0 seconds as it raised TileDBCloudError: Docker logs: Running UDF.
Error executing UDF:
------------------- UDF Backtrace -------------------
  File "/opt/python_generic_udf_entry.py", line 486, in main
    result = udf_function(*(udf_args_parsed[0]))
  File "retry_demo.py", line 21, in f
ValueError: Value less than 0.4: 0.024038032335790716

------------------- End Backtrace -------------------
Error(s): docker container exited with non-zero code: 1. - Code: 4010.
2021-02-04 14:07:56,512: Starting call to 'node_0', this is the 3rd time calling it.
1.9224380470427413
```

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -17,6 +17,7 @@ plotly = ">= 4.0.0"
 networkx = ">= 2.0.0"
 pydot = "*"
 tiledb-plot-widget = ">= 0.1.7"
+tenacity = "*"
 
 [dev-packages]
 
--- a/setup.py
+++ b/setup.py
@@ -29,6 +29,7 @@ REQUIRES = [
     "networkx>= 2.0.0",
     "pydot",
     "tiledb-plot-widget>=0.1.7",
+    "tenacity",
 ]
 
 # NOTE: we cannot use an __init__.py file in the tiledb/ directory, because it is supplied
--- a/tiledb/cloud/compute/delayed.py
+++ b/tiledb/cloud/compute/delayed.py
@@ -1,8 +1,5 @@
 from ..dag.dag import Node, DAG
-
-from ..array import apply as array_apply
-from ..sql import exec as sql_exec
-from ..udf import exec as udf_exec
+from .. import array, sql, udf, retry_task
 
 import numbers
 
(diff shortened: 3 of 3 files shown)


ihnorton (commented):
Tenacity [looks useful](https://julien.danjou.info/python-tenacity/) and seems like a lightweight dependency.

One concern (esp with the user supplying their own retry object) is that IMO we need to make sure it is scoped to the correct types of exceptions. We don't want a user accidentally flooding retry requests due to (their own) logic errors, but only for execution failures on the system side, right?

gsakkis:
I'm not sure if it's possible to determine what the "correct" exception types are in general. A `TypeError` most likely is a user error, a `ValueError` may or may not be. I think for preventing flooding it would make more sense to limit the number of retries? Perhaps this can be enforced on the server side (since all retries of send the same `task_id` of the initial failed task).

Shelnutt2:
I also have a concern about any automatic retry support. There of course will always be transient failures no matter what we do, but I think its more likely for there to be user code errors and the user might end up rerunning a task that costs them which is going to always fail because of a programing mistake in their UDF.

I think you can provide similar benefit with automatic retry by allowing the graph to retry a failed node, and then if it succeeds to resume the graph execution and restart all cancelled tasks.

ihnorton:
If we do end up needing a hook like this, my concern was just to make sure it is hidden enough to prevent someone casually running up a bill by accident.

> I'm not sure if it's possible to determine what the "correct" exception types are in general

We can have a pretty good idea whether an error happens "inside" user code (the code itself throws an exception) or "outside" the user code (for example if there is a problem in executor setup, or if container is killed due to resource limits) because we have catch block(s) around the user code execution. However, I suspect the differences are not yet exposed in a way that is useful for this decision on the client side right now (eg TileDB-Cloud specific, enumerated error types).

gsakkis:
> We can have a pretty good idea whether an error happens "inside" user code (the code itself throws an exception) or "outside" the user code (for example if there is a problem in executor setup, or if container is killed due to resource limits) because we have catch block(s) around the user code execution. 

Assuming we can distinguish between inside and outside user code, is the argument that we shouldn't attempt to support retries for the inside case at all, i.e. let "inside" error handling be the responsibility of the user function ?

PR #125, opened 2021-02-09
Fix task failure handling

gsakkis:
- Fix tests for failed tasks: `.compute()` raises an exception so all subsequent code was being skipped
- Fix the hanging `DelayedBase.compute()` when a task fails
- Cancel all non-started tasks on `DAG.cancel()` and fix the cancellation tests

### DIFF
--- a/tests/test_dag.py
+++ b/tests/test_dag.py
@@ -190,18 +190,21 @@ class DAGFailureTest(unittest.TestCase):
 class DAGCancelTest(unittest.TestCase):
     def test_dag_cancel(self):
         d = dag.DAG()
-
         node = d.add_node(time.sleep, 1)
         node_2 = d.add_node(np.mean, [1, 1])
+        node_2.depends_on(node)
 
         d.compute()
-
         # Cancel DAG
         d.cancel()
 
-        self.assertEqual(node.result(), None)
-        self.assertEqual(node.status, dag.Status.CANCELLED)
         self.assertEqual(d.status, dag.Status.CANCELLED)
+
+        self.assertEqual(node.status, dag.Status.CANCELLED)
+        self.assertEqual(node.result(), None)
+
+        self.assertEqual(node_2.status, dag.Status.CANCELLED)
+        self.assertEqual(node_2.result(), None)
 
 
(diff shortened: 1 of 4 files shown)


tests/test_dag.py
```diff
@@ -146,67 +146,66 @@ def test_end_nodes_dag(self):
 
 class DAGFailureTest(unittest.TestCase):
     def test_dag_failure(self):
-        with self.assertRaises(TypeError):
```
Shelnutt2, line 149: So the behavior change here is the `compute` no longer raises the exception?

Shelnutt2: @gsakkis can we make `compute` still raise the exception? Else now a user has no way to know there was a failure, per the test here `d.compute()` completes and returns without any signally to the user that it task graph actually failed.

gsakkis: @Shelnutt2 `Dag.compute` is currently async, it just kicks off the dag computation without waiting for it to finish (and therefore know if an exception is raised). What we can do is raise the exception in `Dag.wait`. We can also make `Dag.compute` (optionally) synchronous by having it call `wait`; in fact that's what [DelayedBase.compute](https://github.com/TileDB-Inc/TileDB-Cloud-Py/pull/125/files#diff-2e84d094d407cf2aedc028330223fe426d6fef815326cf7f4d4cb445a25fa49cR41) does.

Shelnutt2: I was intending `Delayed. compute()` and `Dag.wait()`. Doing it in `Dag.wait()` should be enough to propagate it up.

gsakkis: Moved the reraise logic to `DAG.wait` at https://github.com/TileDB-Inc/TileDB-Cloud-Py/pull/125/commits/1f72ac9e83dfa2d849f7ccb9463c9d363ffe6450

Shelnutt2: Thanks! I'll test and we can get this merged soon.

Shelnutt2 (approved):
I've tested a variety of scenarios and it works great. Thanks for the much needed improvements here!

PR #132, opened 2021-03-18
Add support for multi-array UDFs

antalakas:
Usage:

```
def median(numpy_ordered_dictionary):
    return np.median(numpy_ordered_dictionary[0]["a"]) + np.median(numpy_ordered_dictionary[1]["a"])

array_list = array.ArrayList()
array_list.add(dense_array, [(1, 4), (1, 4)], ["a"])
array_list.add(sparse_array, [(1, 2), (1, 4)], ["a"])
namespace="namespace"
res = array.exec_multi_array_udf(median, array_list, namespace)
print("Median Multi UDF:\n{}\n".format(res))
```


tiledb/cloud/array.py
```diff
@@ -563,3 +562,165 @@ def apply(
...
+        converted_layout = "row-major"
+    elif layout.upper() == "C":
+        converted_layout = "col-major"
+    elif layout.upper() == "G":
+        converted_layout = "global-order"
+    elif layout.upper() == "U":
+        converted_layout = "unordered"
+
+    arrays = [] # list[UDFArrayDetails]
+    for uri, a_d in array_details.items():
```
Shelnutt2, line 634: Lets not use tuples here, its very easy to make mistakes and hard for the user to reason with. I think either we should expose the `rest_api.models.UDFArrayDetails` or have our own high level class for the user to build off of so its easier to see which parameters are for which and view `help()` etc.

antalakas: Added high level class `ArrayList` the user can use to build param list

tiledb/cloud/array.py
```diff
@@ -563,3 +602,150 @@ def apply(
...
+    >>> import numpy as np
+    >>> from tiledb.cloud import array
+    >>> import tiledb.cloud
+    >>> def median(numpy_ordered_dictionary):
+    ...     return np.median(numpy_ordered_dictionary["a"])
+
+    >>> dense_array = "tiledb://TileDB-Inc/quickstart_dense"
+    >>> sparse_array = "tiledb://TileDB-Inc/quickstart_sparse"
+    >>> array_details = dict([
+    >>>     (dense_array, ([(1, 4), (1, 4)], ["a"])),
```
Shelnutt2, line 641: This example should be updated for the new parameter

tiledb/cloud/array.py
```diff
@@ -563,3 +602,150 @@ def apply(
...
+    include_source_lines=True,
+    task_name=None,
+    result_format=rest_api.models.UDFResultType.NATIVE,
+    result_format_version=None,
+    **kwargs
+):
+    """
+    Apply a user defined function to multiple arrays
+    :param func: user function to run
+    :param array_details: dict from array name (tiledb format) to tuple (ranges, attributes)
```
Shelnutt2, line 623: Adjust the description to match the new class

tiledb/cloud/array.py
```diff
@@ -563,3 +602,150 @@ def apply(
...
+        raise TypeError("name argument to `apply` must be set if no function is passed")
+
+    pickledUDF = None
+    source_lines = None
+    if func is not None:
+        source_lines = utils.getsourcelines(func) if include_source_lines else None
+        pickledUDF = cloudpickle.dumps(func, protocol=udf.tiledb_cloud_protocol)
+        pickledUDF = base64.b64encode(pickledUDF).decode("ascii")
+
+    if array_list is None:
```
Shelnutt2, line 663: Lets also check the type of the array_list passed

tiledb/cloud/array.py
```diff
@@ -563,3 +602,150 @@ def apply(
...
+        raise TypeError("arrays need to have passed in order to call multi array udfs")
+
+    # Get the list of arrays
+    arrays = array_list.get()
+
+    if len(arrays) == 0:
+        raise TypeError("arrays need to have passed in order to call multi array udfs")
+
+    # Get the namespace from the first array in list
```
Shelnutt2, line 672: This is incorrect. The namespace to upload the task too should use the same logic as the generic UDF

Shelnutt2 (changes requested):
A few minor changes else this is looking good!

tiledb/cloud/array.py
antalakas: Updated

tiledb/cloud/array.py
antalakas: Adjusted

tiledb/cloud/array.py
antalakas: The type of `array_list` is checked

tiledb/cloud/array.py
antalakas: The `namespace` is passed to parameter as an argument

Shelnutt2 (approved):
Thanks this is awesome addition!

antalakas:
> Thanks this is awesome addition!
👍

PR #139, opened 2021-05-04
client: pass permissions as a str rather than list

thetorpedodog:


tiledb/cloud/client.py
```diff
@@ -5,8 +5,9 @@
 from . import config
 from . import tiledb_cloud_error
 from .rest_api import ApiException as GenApiException
-import urllib
 import os
+import typing as t
```
ihnorton, line 9: Please use mypy style explicit imports.

ihnorton: vscode's python extension can add them automatically (for better or sometimes worse)

ihnorton (changes requested):
Style nit, but otherwise LGTM.

tiledb/cloud/client.py
thetorpedodog: Done.

ihnorton: Sorry, I was unclear there -- by explicit imports I meant: `from typing import Union, Optional, ...`. That's the mypy convention and what we've adopted in other repos.

thetorpedodog: ah jeez, I had not been staging my new changes when I ran `git commit --amend` so it was happily pushing no changes and just updating the commit time.

now **acutally** fixed.

ihnorton: Thank you!

ihnorton (approved):

tiledb/cloud/client.py
```diff
@@ -5,8 +5,8 @@
 from . import config
 from . import tiledb_cloud_error
 from .rest_api import ApiException as GenApiException
-import urllib
```
ihnorton, line 8: I think this was unintentionally dropped here, which is causing the CI failure.

thetorpedodog: Whoops! Fixed.

ihnorton (approved):

Shelnutt2 (approved):

PR #140, opened 2021-05-04
Add tiledb.cloud.version Module

nguyenv:
* Similar to TileDB-Py's `tiledb.version`, `tiledb.cloud.version` is a
  module containing two attributes: the `version` string and `version_tuple`


Shelnutt2 (approved):
Thanks!

thetorpedodog:
This change adds the import to `tiledb.cloud`, but it looks like the actual `version.py` file got left behind! (I too have forgotten to `git add` many times...)

Shelnutt2:
@thetorpedodog the version file is autogenerated, we exclude it from being added to the repo. We use `setuptools_scm` to auto-generate it based on the git tag.

PR #143, opened 2021-05-17
Default Serverless SQL Queries to Arrow Return Type

Shelnutt2:
This PR switches the default return type of serverless SQL queries to apache arrow instead of JSON. The user is given the option to toggle the format they wish.

This PR also adjusts `UDFResult`/`SQLResults` classes to have a common base `TaskResult`. This allows for all the functionality around serialization and results to be handled in a common base class. As an effect now SQLResult gets support for apache arrow result type in the same way UDFs do.


thetorpedodog:
Closing this as obsoleted by #155.

PR #144, opened 2021-05-17
WIP: fix incorrect datetime ranges in UDF apply

ihnorton:
If the user supplies a np.datetime64 start/end range with a different time resolution than the
underlying dimension in the schema, then currently we cast directly to int64 which will
not be correct when applied server-side. In this patch we add passthrough for the schema,
and then apply the corresponding dimension dtype before getting the raw int64 value.
Untested, pending fixes for unrelated issue server-side.

### DIFF
--- a/tests/test_array_udf.py
+++ b/tests/test_array_udf.py
@@ -1,0 +1,48 @@
+import tiledb, tiledb.cloud
+import sys, os, platform, unittest, tempfile
+import numpy, numpy as np
+
+from tiledb import Dim, Domain, Attr, ArraySchema
+from tiledb.cloud import client
+from tiledb.cloud import array
+from tiledb.cloud import tasks
+from tiledb.cloud import tiledb_cloud_error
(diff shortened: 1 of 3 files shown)


ihnorton:
Marked as draft because:
- unverified, pending server-side issues
Two other issues, which might be tricky/require refactoring:
- needs test fix
- there is another use of `parse_ranges` in ArrayList, but not sure how we can get an `ArraySchema` reference there

ihnorton:
This fixes the motivating issue, but two things need to be addressed:
- the test is WIP
  - needs a tmp directory with `tiledb://` address, or we can write a permanent array to test reads
  - needs to actually check the returned values for several different indexing modes (currently stubbed)
- need to figure out what to do with the ArrayList class which also uses `parse_ranges` — will need refactoring there if we have to also pass a schema through

ihnorton:
Tabling for now.

PR #145, opened 2021-05-19
Don't split array name for tasks

Shelnutt2:
The server expects it in `tiledb://` format and splits it there.

### DIFF
--- a/tiledb/cloud/tasks.py
+++ b/tiledb/cloud/tasks.py
@@ -1,6 +1,6 @@
 from . import client
 from . import tiledb_cloud_error
-from .array import split_uri
+from . import config
 from . import sql
 from . import array
 from .rest_api import ApiException as GenApiException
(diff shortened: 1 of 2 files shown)


tiledb/cloud/tasks.py
```diff
@@ -52,6 +52,12 @@ def tasks(
     """
     api_instance = client.client.tasks_api
 
+    if namespace is None:
```
thetorpedodog, line 55: "" is not a valid namespace, correct? If so, this can be `if not namespace`.

tiledb/cloud/tasks.py
```diff
@@ -52,6 +52,12 @@ def tasks(
...
+        if config.user is None:
+            config.user = client.user_profile()
```
thetorpedodog, line 57: This is a kind of unexpected mutation of global state—I think a function to do this get-or-default probably belongs in config.py, so that it's more predictable what is going to happen there.

tiledb/cloud/tasks.py
```diff
@@ -52,6 +52,12 @@ def tasks(
...
+
+        namespace = config.user.usernam
```
thetorpedodog, line 59: username?

thetorpedodog (commented):
Sorry for having missed this review for so long! If I don't get to any code review within like half a day please bump me.

PR #146, opened 2021-06-01
utils: fix exception logging and do cleanup in getsourcelines.

thetorpedodog:

PR #147, opened 2021-06-01
Enable use of registered UDFs by name from Delayed classes.

thetorpedodog:
This series of changes enables users to use registered UDFs in Delayed classes in a series of steps:

1. Enable users to pass the name of the function to the backend UDF functions via the `func` parameter, so they can write:
    ```python
    # pass a callable
    udf.exec(my_real_function, param1, named=param2)
    # pass a UDF name
    udf.exec('me/my_registered_udf', param1, named=param2)
    ```
    
    This comprises the first two commits.

2. Propagate support through the `delayed` module, and add tests. This is the final commit.

This change is currently diffbase’d off of #146, which does substantial cleanup and testing that makes this half of the change easier.

Much of this change is also concerned with cleaning up the registration functions as well. If it would make it easier to review, I am amenable to splitting the first stage of this process away from the second stage and into its own pull request (and could even split stage one into array and UDF halves).

### DIFF
--- a/tests/test_delayed.py
+++ b/tests/test_delayed.py
@@ -1,10 +1,15 @@
+import os
 import time
 import unittest
-import os
 
 import numpy as np
-from tiledb.cloud.compute import Delayed, DelayedSQL, DelayedArrayUDF, Status
 import tiledb.cloud
+from tiledb.cloud import testonly
+from tiledb.cloud.compute import Delayed
+from tiledb.cloud.compute import DelayedArrayUDF
+from tiledb.cloud.compute import DelayedSQL
+from tiledb.cloud.compute import Status
+
 
 tiledb.cloud.login(
     token=os.environ["TILEDB_CLOUD_HELPER_VAR"],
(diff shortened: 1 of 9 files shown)


thetorpedodog:
bump. @Shelnutt2 I added you for the review on this because the bug came from you but feel free to bump it to somebody else if they can better handle it

tests/test_delayed.py
```diff
@@ -152,6 +157,28 @@ def test_array_apply(self):
...
+        uri = "tiledb://TileDB-inc/quickstart_sparse"
+        with tiledb.open(uri, ctx=tiledb.cloud.Ctx()) as A:
+            orig = A[:]
+
+        def sum_a(x):
+            import numpy
+
+            return numpy.sum(x["a"])
+
+        with testonly.register_udf(sum_a) as sum_a_name:
```
Shelnutt2, line 171: This is handy!

Shelnutt2 (approved):

PR #148, opened 2021-06-02
Copy signature of async array and UDF functions onto sync versions.

thetorpedodog:
This change introduces `signature_of`, which copies the function
signature from a source to the decorated function, and applies it to
the synchronous functions in array.py and udf.py.

This avoids potential issues with parameter order or defaults falling
out of sync between the synchronous and the async versions, while still
allowing interactive users to access tab-completion. The docstrings
of the synchronous versions link directly to the async versions.

I have manually tested this both in IPython and Jupyter and tab
completion and other contextual help continues to work there.

Also adds tests for `tiledb.cloud.utils`.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,0 +1,37 @@
+import importlib
+import os
+import os.path
+import sys
+import tempfile
+import textwrap
+import unittest
+
+from tiledb.cloud import utils
+
+
+class SourceLinesTest(unittest.TestCase):
+    def test_this_function(self):
+        me = utils.getsourcelines(self.test_this_function)
+        self.assertIsInstance(me, str)
+
+    def test_missing_function(self):
+        with tempfile.TemporaryDirectory() as tmpdir:
+            testfile = os.path.join(tmpdir, "bogus_module.py")
+            with open(testfile, "w") as out:
+                out.write(
+                    textwrap.dedent(
+                        """\
+                        def unimportant_function(a):
+                            pass
+                        """
+                    )
+                )
+            sys.path.insert(0, tmpdir)
+            try:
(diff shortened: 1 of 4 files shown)


ihnorton:
@thetorpedodog, how does this interact with documentation generation? We have generated docs (https://tiledb-inc-tiledb-cloud-py.readthedocs-hosted.com/en/latest/#) -- although it looks like several aspects are broken since last time I looked at it (ch7907).

thetorpedodog:
The generated documentation will include a hyperlink to the base function (as included in the updated docstrings). This is a tradeoff, because the docs are no longer Right There, but right now some of the docstrings are out-of-date (missing params, etc.) and my read of it is that having a single source which we can keep up to date is better than having two places that we have to remember to manually copy-and-paste every time we make a documentation change. (And as a reader, it might even be better to just have the "this works exactly the same as [link to Y]" so that it's clear that there is no difference.)

Sphinx generates docs by importing the module, so if we *really* wanted to make sure they were both present and stayed up-to-date, we could do some find-and-replace docstring-copying magic in the decorator, but I don't think that would actually have any advantages over saying "this works exactly the same as [other thing]".

thetorpedodog:
This may also make more sense in the context of #147, where I do significant work on annotating function signatures and updating docstrings. I could see some basic code generation as an alternative (read in the async signature and docstring, spit out a sync function; add a test to ensure that they stay up to date) but I am not sure that provides a meaningful benefit to users. (This, of course, is my opinion; I'd be fine with doing it if you think it is worth it.)

ihnorton (approved):

nguyenv (approved):

PR #150, opened 2021-06-30
Update autogenerated API files (and minor script fixes).

thetorpedodog:
Pulling this out to separate codegen changes from actual functionality changes.


thetorpedodog:
haha whoops some non-generated changes need to happen at the same time as the generated ones. back into the drafts with you

thetorpedodog:
ok we are back

ihnorton (approved):
LGTM, but noting that when I've worked on updates like this in the past, I try to keep the generated updates isolated in separate commits for two reasons: (1) review-ability and (2) so that git history is somewhat more manageable. You could split it up into three commits like this:

```
git reset HEAD~1 --mixed
git add update_generated.sh
git commit -m "Bump update_generated.sh"
git add tiledb/cloud/rest_api/
git commit -m "Check in generated files"
git diff
git commit -a -m "Update HL API files to use new generated output"
```

thetorpedodog:
Keeping the script update as a separate commit makes sense (obviously, we don't re-run the script instantly, so it being not precisely what its output is is fine). The reason I kept the codegen and UDFResultType-to-ResultFormat changes together was the desire to have each step of the commit be working code, rather than have a commit that you switch to and it doesn’t work altogether.

PR #152, opened 2021-07-06
Fix build problems (add wheel, switch Ubuntu version)

thetorpedodog:
Per the Packaging Guide, the new and cool way to build wheels is with
the `build` package, which will also build your source distribution.

https://packaging.python.org/guides/distributing-packages-using-setuptools/#wheels

### DIFF
--- a/azure-pipelines.yaml
+++ b/azure-pipelines.yaml
@@ -36,7 +36,7 @@ jobs:
       versionSpec: '$(python.version)'
       architecture: 'x64'
 
-  - script: python -m pip install --upgrade 'numpy==1.17.*' pytest pylint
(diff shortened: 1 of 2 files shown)


thetorpedodog:
Change of plans. We're now going to add wheel to the build dependency list because the new `build` module doesn't seem to be picking up on the versioning.

thetorpedodog:
@ihnorton kind of at a loss here. I added setuptools/-scm and wheel to the requirements list, which fixed the "build wheel" part of the build process, but doing so then broke the py3.7 test at runtime with a different error than it had before???

ihnorton (approved):
LGTM as a temporary measure; issue opened to fix the compatibility issue.

PR #155, opened 2021-07-26
Fully support Apache Arrow as a Serverless SQL response format

thetorpedodog:
This is a continuation of #143, but it felt weird to push a bunch of changes onto somebody else’s code review so here we are. Much of what was blocking that change was actually server-side issues but fixes for those are forthcoming.

Each of the original two changes has been slightly patched so that the project should be fully working at each commit. My changes to each are as follows:

- 3c180ee56d8dde92320e54c037d786f615f43876 (was df68652c6e5931cd38432aceaa512aa1a10758e6): Use the right arrow package (also reordered to be first in the sequence)
- fce6add3edc296c18d5a4cfd2d031964f7893e12 (was 04c693c6f63bf9b10d02febdb857b60d60a92382): Update TaskResult to correctly set `task_id`.

I also applied a final cleanup. Once the server-side changes for those are deployed, automated testing should pass and this should be good to go.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -17,6 +17,7 @@ plotly = ">= 4.0.0"
 networkx = ">= 2.0.0"
 pydot = "*"
 tiledb-plot-widget = ">= 0.1.7"
+pyarrow = "3.0.0"
 
 [dev-packages]
 
--- a/setup.py
+++ b/setup.py
@@ -29,6 +29,7 @@ REQUIRES = [
     "networkx>= 2.0.0",
     "pydot",
     "tiledb-plot-widget>=0.1.7",
+    "pyarrow==3.0.0",
 ]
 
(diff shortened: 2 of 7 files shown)


thetorpedodog:
The test errors that remain here appear to be because the server is serializing data that does not directly come from a column as a (byte)string. For instance:

```
In [3]: uri = "tiledb://TileDB-inc/quickstart_sparse"

In [4]: sql.exec("select SUM(`a`) as a from `{}`".format(uri))
Out[4]: 
      a
0  b'6'
```

I am confident that the client is doing the right thing here, but we can’t make this the default until we fix the server side, which may end up being quite involved :(

thetorpedodog:
It works! We’re ready to go!

Shelnutt2 (approved):
Great work!

PR #156, opened 2021-08-16
Add ResultFormat alias back to __init__.py and sort imports.

thetorpedodog:
- Adds ResultFormat alias to (hopefully) restore backwards compat
  with some existing code.
- Creates __all__ to avoid to lint-disable.
- Organizes imports.

### DIFF
--- a/tiledb/cloud/__init__.py
+++ b/tiledb/cloud/__init__.py
@@ -1,42 +1,63 @@
 # This file imports specifically to re-export.
-# flake8: noqa: F401
-
-from .client import (
-    Config,
-    Ctx,
-    list_arrays,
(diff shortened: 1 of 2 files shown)


Shelnutt2 (approved):

thetorpedodog:
hmm. these failures suggest that there is a problem on the server side, probably related to some recent context-handling work.

antalakas (approved):

ihnorton (commented):
Not really a fan of the java style import-per-line, myself... Otherwise LGTM.

thetorpedodog:
This now works on a staging instance. We should be able to rerun and get correct test results once it is deployed.

PR #157, opened 2021-08-24
Update generator script to use a local binary, and regenerate

thetorpedodog:
- ~Updates to use OpenAPI Generator v5.2.1 (legacy Python generator).~
- Uses a locally-stored JAR rather than downloading and spinning up
  an entire container.
- Minor tweaks to how generation works to make it easier and safer:
  - Automatically figure out output path.
  - Automatically remove things, but only after everything else
    has succeeded
  - Use a fake package name so already-replaced sites are unambiguous.
  - Work around a generator bug so manual edits are no longer needed.

---

This is split up into the generator script update and the generate change. The best way to review the second half is probably to just verify that I have not done anything sneaky after by running `./update-generated.sh` and noting that nothing has changed.

~Functionally, nothing should change between the old and new versions of the generator, other than some minor internal improvements; all our external code should work exactly the same.~


thetorpedodog:
ok so it turns out that the python code generated by the legacy generator in v5.1 is not compatible with that generated by v4.3.

one thing at a time then.

ihnorton (changes requested):
What is the motivation for changing this up and checking in the .jar? Git is not very smart, which means that every time we update the .jar, that gets stored in the history. This package checkout is already getting hefty.

thetorpedodog:
> What is the motivation for changing this up and checking in the .jar? Git is not very smart, which means that every time we update the .jar, that gets stored in the history. This package checkout is already getting hefty.

The motivation is to make generation quicker and easier, by running it locally rather than requiring spinning up an entire container, and avoiding the need for a sudo prompt in the middle of the generation process.

Including the `.jar` seemed like the natural thing for me at the time—I am accustomed to having tools checked in to avoid external dependencies in a world where checkout size was never a concern. I have updated the change to download and cache the jar instead.

thetorpedodog:
Also separately it would be nice to update the generated API to use the new Python generator, but that seems like a Lot of work at the moment.

tiledb/cloud/rest_api/test/test_user_api.py
```diff
@@ -14,7 +14,7 @@
 
 import unittest
 
-import rest_api
+import tiledb.cloud.rest_api
```
ihnorton, line 17: it looks like something is not right with the substitutions

ihnorton:
The other issue with a jar is platform and java version differences -- the container environment was fairly reproducible. It's ok with me since we don't have many developers on this, but please get @Shelnutt2 sign-off as well.

thetorpedodog:
> The other issue with a jar is platform and java version differences -- the container environment was fairly reproducible. It's ok with me since we don't have many developers on this, but please get @Shelnutt2 sign-off as well.

The JAR includes all its dependencies (confirmed by running on a new VM with a fresh install of Java), and the generated code will be identical with any supported Java version (>8, which was released in 2014).

thetorpedodog:
re-regenerated the API on [one change back from HEAD](https://github.com/TileDB-Inc/TileDB-Cloud-API-Spec/tree/d931dd68498340f4c330373f00d912c3f11b967b) since the latest commit removes `result_format_version`, which is an incompatible change that I want to keep separate

Shelnutt2:
> > The other issue with a jar is platform and java version differences -- the container environment was fairly reproducible. It's ok with me since we don't have many developers on this, but please get @Shelnutt2 sign-off as well.
> 
> The JAR includes all its dependencies (confirmed by running on a new VM with a fresh install of Java), and the generated code will be identical with any supported Java version (>8, which was released in 2014).

If only jvm actually worked as describe.. it does seem here that this is compatible and fine with java11 and java8 jre's from openjdk and iced-tea. I'm fine with this. If we find we prefer docker we can just go back, or setup a custom image. I'd rather go with what makes @thetorpedodog development easier at the moment.

ihnorton (approved):

PR #161, opened 2021-08-30
Notebook uploader

johnkerl:
Ready for review, with the following caveat:

* We might decide how much of the auto-increment logic to port (or not to port) from `TileDB-Cloud-Jupyter-Contents`.
*  If destination exists, three options:
    * Reject the write with 'already exists'. 
    * Auto-increment (as done in Jupyter)
    * Overwrite current version (should not be the default; perhaps should be supported as an option)
* Today's code does the first option and can be merged if everyone's OK with that for now.

# Manual test scenario

* Use `download-notebook.py` (below) to get valid JSON
* Use `upload-notebooy.py`
* Click through in the TileDB Cloud console and verify all tabs work for the newly created notebook

```
$ cat download-notebook.py
#!/usr/bin/env python

# ================================================================
# Setup:
# * https://console.dev.tiledb.io
# * Set my storage path and storage credential name in Settings
#
# Note: cannot be invoked from within the base directory of either the
# TileDB-Py or TileDB-Cloud-Py repos -- or any directory with a subdirectory
# named `tiledb` -- since `import tiledb` / `import tiledb.cloud` clashes
# with the existence of that subdirectory.
# ================================================================

import sys
import tiledb.cloud

config = tiledb.Config()
config["rest.username"] = "johnkerl"
config["rest.password"] = "redacted"
# API endpoint used by https://console.dev.tiledb.io:
config["rest.server_address"] = "api.dev.tiledb.io"

# This is the array-URI format in TileDB Cloud:
tiledb_uri              = "tiledb://seth/Voila-Covid-Demo"
storage_credential_name = "my-sandbox-creds"
ipynb_file_name         = "/Users/johnkerl/Desktop/PLAN/foo.ipynb"

if len(sys.argv) == 3:
    tiledb_uri      = sys.argv[1]
    ipynb_file_name = sys.argv[2]

tiledb.cloud.download_ipnyb_file_name_from_cloud(
    tiledb_uri,
    storage_credential_name,
    ipynb_file_name,
)

print("Wrote "+ipynb_file_name)
```

```
$ cat upload-notebook.py
#!/usr/bin/env python

# ================================================================
# Setup:
# * https://console.dev.tiledb.io
# * Set my storage path and storage credential name in Settings
#
# Note: cannot be invoked from within the base directory of either the
# TileDB-Py or TileDB-Cloud-Py repos -- or any directory with a subdirectory
# named `tiledb` -- since `import tiledb` / `import tiledb.cloud` clashes
# with the existence of that subdirectory.
# ================================================================

import sys
import tiledb.cloud

tiledb.cloud.login(
    host="https://api.dev.tiledb.io/v1",
    username="johnkerl",
    password="redacted",
)

ipynb_file_name         = "/Users/johnkerl/Desktop/PLAN/foo.ipynb"
namespace               = "johnkerl"
array_name              = "testing-upload-014"
storage_path            = "s3://tiledb-johnkerl"
storage_credential_name = "my-sandbox-creds"

if len(sys.argv) == 3:
    ipynb_file_name = sys.argv[1]
    array_name      = sys.argv[2]

tiledb_uri = tiledb.cloud.upload_ipnyb_file_name_to_cloud(
    ipynb_file_name,
    namespace,
    array_name,
    storage_path,
    storage_credential_name,
)

print("tiledb_uri="+tiledb_uri)
```

Be sure to run this for the cases when the destination array does not already exist, and does already exist.

# Tracking

CH 7290


tiledb/cloud/__init__.py
```diff
@@ -21,6 +21,10 @@
 from .client import organizations
 from .client import user_profile
 from .notebook import rename_notebook
+from .notebook import upload_ipnyb_file_name_to_cloud
```
johnkerl, line 24: Nothing needed here to export `tiledb.cloud.notebook.upload_ipnyb_file_name_to_cloud`. These are to let people access `upload_ipnyb_file_name_to_cloud` from `import tiledb.cloud` rather than `import tiledb.cloud.notebook`.

tiledb/cloud/notebook.py
```diff
@@ -35,3 +54,288 @@ def rename_notebook(
...
+    :param str character_encoding: Optional character encoding of the file; defaults to utf-8 which is ASCII-compatible.
+    :return: contents of the notebook file as a string, nominally in JSON format.
+    """
+    ctx = tiledb.cloud.Ctx({
+        "rest.creation_access_credentials_name": storage_credential_name,
+    })
+    with tiledb.open(tiledb_uri, 'r', ctx=ctx) as A:
+        size = A.meta["file_size"]
+        data = A.query(attrs=["contents"])[slice(0, size)]
+        json = data["contents"].tostring().decode("utf-8", "backslashreplace")
```
thetorpedodog, line 102: This should probably be `tobytes()`, since `tostring()` is a deprecated alias.

I also don’t think we want a special error handler here; if the data we get back is invalid UTF-8, that is a problem that we probably don’t want to paper over.

tiledb/cloud/notebook.py
```diff
@@ -35,3 +54,288 @@ def rename_notebook(
...
+    tiledb_uri: str,
+    storage_credential_name: str,
+    ipynb_file_name: str,
+    *,
+    character_encoding: Optional[str] = 'utf-8',
+) -> None:
+    """
+    Downloads a notebook file from TileDB Cloud to local disk.
+    :param str tiledb_uri: such as "TileDB-Inc/quickstart_dense".
+    :param str storage_credential_name: such as "janedoe-creds", typically from the user's account settings.
```
thetorpedodog, line 69: I am open to correction from other reviewers but, for easier reading as a user, I prefer to wrap the docstrings at 80 cols and continue the descriptions on the next line with a two-space indent:

```
    :param storage_credential_name: yada yada yada
      yada yada
```

tiledb/cloud/notebook.py
```diff
@@ -35,3 +54,288 @@ def rename_notebook(
...
+            uri=tiledb_uri,
+            file_type=tiledb.cloud.rest_api.models.FileType.NOTEBOOK,
+            # If file_properties is empty, don't send anything at all.
+            file_properties=file_properties or None,
+        )
+
+        return True, tiledb_uri, array_name
+    except tiledb.TileDBError as e:
+        raise e
+        if "Error while listing with prefix" in str(e):
```
thetorpedodog, line 304: This code is unreachable.

tiledb/cloud/notebook.py
```diff
@@ -35,3 +54,288 @@ def rename_notebook(
...
+            # It is possible to land here if user sets wrong default S3
+            # credentials with respect to default S3 path.
+            raise Exception.web.HTTPError(
+                400, f"Error creating file: {e}. Are your credentials valid?"
+            ) from e
+
+    except Exception as e:
+        raise e
```
thetorpedodog, line 312: This probably shouldn't be here anyway (no point in having a catch clause that immediately re-raises and does nothing else) but in the case where you *do* want to re-raise an exception (e.g. after filtering something out), you should use the bare `raise`, which will bubble the exception up without rewriting its call stack etc.

thetorpedodog (commented):
lots of little formatting things as well (mixes of `"` and `'`, stray spaces, etc.) that Black should fix automatically too

tiledb/cloud/notebook.py
```diff
@@ -1,9 +1,28 @@
...
+
 from . import rest_api
 from . import client
 from . import tiledb_cloud_error
 from . import array
 from .rest_api import ApiException as GenApiException
 from .rest_api import rest
+import tiledb
+
+from typing import Optional, Tuple
```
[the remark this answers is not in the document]
johnkerl: vim + my conscience, LOL -- thanks! :)

tiledb/cloud/notebook.py
```diff
@@ -35,3 +54,288 @@ def rename_notebook(
...
     except GenApiException as exc:
         raise tiledb_cloud_error.check_exc(exc) from None
+
+
+def download_ipnyb_file_name_from_cloud(
+    tiledb_uri: str,
+    storage_credential_name: str,
+    ipynb_file_name: str,
+    *,
+    character_encoding: Optional[str] = 'utf-8',
```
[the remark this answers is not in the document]
johnkerl: awesome py-fu, thank you! :)

tiledb/cloud/__init__.py
johnkerl: thanks! :)

johnkerl:
@thetorpedodog thanks for all the great feedback -- ready for round two! :)

tiledb/cloud/notebook.py
```diff
@@ -35,3 +51,296 @@ def rename_notebook(
...
+
+
+def download_ipnyb_file_name_from_cloud(
+    tiledb_uri: str,
+    storage_credential_name: str,
+    ipynb_file_name: str,
+) -> None:
+    """
+    Downloads a notebook file from TileDB Cloud to local disk.
+    :param tiledb_uri: such as "TileDB-Inc/quickstart_dense".
```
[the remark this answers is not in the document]
johnkerl: My typo in the docstring; should have been `tiledb://TileDB-Inc/quickstart_dense`. (Fixed.)

tiledb/cloud/notebook.py
```diff
@@ -35,3 +51,296 @@ def rename_notebook(
...
+    """
+    ctx = tiledb.cloud.Ctx(
+        {
+            "rest.creation_access_credentials_name": storage_credential_name,
+        }
+    )
+    with tiledb.open(tiledb_uri, "r", ctx=ctx) as arr:
+        size = arr.meta["file_size"]
+        data = arr.query(attrs=["contents"])[slice(0, size)]
+        json = data["contents"].tobytes().decode("utf-8", "backslashreplace")
```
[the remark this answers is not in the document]
johnkerl: good catch, thanks!

tiledb/cloud/notebook.py
```diff
@@ -35,3 +51,296 @@ def rename_notebook(
...
+
+        return True, tiledb_uri, array_name
+    except tiledb.TileDBError as e:
+        if "Error while listing with prefix" in str(e):
+            # It is possible to land here if user sets wrong default S3
+            # credentials with respect to default S3 path.
+            raise Exception.web.HTTPError(
+                400, f"Error creating file: {e}. Are your credentials valid?"
+            ) from e
+        else:  # already-exists raises here
```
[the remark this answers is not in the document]
johnkerl: i defer to your style preference ;) -- i do like if/else; i find them more clear to the reader on a visual scan, even if not needed by the machine.

tiledb/cloud/notebook.py
```diff
@@ -35,3 +51,296 @@ def rename_notebook(
...
+        tiledb.cloud.array.update_info(uri=tiledb_uri, array_name=array_name)
+
+        tiledb.cloud.array.update_file_properties(
+            uri=tiledb_uri,
+            file_type=tiledb.cloud.rest_api.models.FileType.NOTEBOOK,
+            # If file_properties is empty, don't send anything at all.
+            file_properties=file_properties or None,
+        )
+
+        return True, tiledb_uri, array_name
```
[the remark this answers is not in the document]
johnkerl: thanks @thetorpedodog ! this was a holdover from the idea of implementing auto-increment (which apparently we aren't interested in doing on this PR). here i was thinking of separating out different ways of failing. but without the auto-increment, good call, this is just a needless remnant of an unimplemented idea.

johnkerl (commented):
@thetorpedodog ready for round three! :)

tiledb/cloud/notebook.py
```diff
@@ -14,9 +27,11 @@ def rename_notebook(
 ):
     """
     Update an array's info
-    :param str namespace: optional username or organization array should be registered under. If unset will default to the user
+    :param str namespace: optional username or organization array should be
```
thetorpedodog, line 33: `str` no longer needed since that information appears in the signature

tiledb/cloud/notebook.py
```diff
@@ -35,3 +50,290 @@ def rename_notebook(
...
+      user's account settings.
+    :param array_name : name to be seen in the UI, such as "testing-upload"
+    :param namespace: such as "janedoe".
+    :param storage_credential_name: such as "janedoe-creds", typically from the
+      user's account settings.
+    :return: TileDB array name, such as "tiledb://janedoe/testing-upload".
+    """
+
+    ctx = tiledb.cloud.Ctx(
+        {
```
thetorpedodog, line 183: weird that black gave you this. will it let you do

```python
ctx = tiledb.cloud.Ctx(
    {"rest.creation_access_credentials_name": storage_credential_name},
)
```
?

tiledb/cloud/notebook.py
```diff
@@ -35,3 +50,290 @@ def rename_notebook(
...
+        except tiledb.TileDBError as e:
+            if "Error while listing with prefix" in str(e):
+                # It is possible to land here if user sets wrong default S3
+                # credentials with respect to default S3 path.
+                raise ValueError(
+                    f"Error creating file: {e}. Are your credentials valid?"
+                ) from e
+            if "Cannot create array" in str(e) and "already exists" in str(e):
+                raise ValueError(
+                    f"Error creating file: {array_name} already exists in namespace {namespace}."
```
thetorpedodog, line 257: recommend throwing an `!r` in these formats so that the user gets nice quotation marks.

tiledb/cloud/notebook.py
```diff
@@ -35,3 +50,290 @@ def rename_notebook(
...
+                )
+        # Retry other exceptions
```
thetorpedodog, line 259: I would bump this in by four spaces so that it is in the except block and note that it is "retry other TileDBErrors", because other exception types will still bubble up.

tiledb/cloud/notebook.py
```diff
@@ -35,3 +50,290 @@ def rename_notebook(
...
+
+    contents_as_array = numpy.array(bytearray(ipynb_file_contents, CHARACTER_ENCODING))
+
+    with tiledb.open(tiledb_uri, mode="w", ctx=ctx) as arr:
+        arr[0 : len(contents_as_array)] = {"contents": contents_as_array}
+        arr.meta["file_size"] = len(contents_as_array)
+        arr.meta["type"] = file_type = tiledb.cloud.rest_api.models.FileType.NOTEBOOK
+        arr.meta["format"] = "json"
+
+    return
```
thetorpedodog, line 342: no need for `return` at the end of a void function.

tiledb/cloud/notebook.py
```diff
@@ -35,3 +50,290 @@ def rename_notebook(
...
+        ],
+        ctx=ctx,
+    )
+
+    # Goal: tiledb://my_username/s3://my_bucket/my_array
+    # https://docs.tiledb.com/cloud/how-to/arrays/create-arrays
+    tiledb_uri_s3 = "tiledb://" + posixpath.join(namespace, storage_path, array_name)
+
+    # Create the (empty) array on disk.
+    tiledb.DenseArray.create(tiledb_uri_s3, schema)
```
thetorpedodog, line 294: Just use `tiledb.Array.create` here; in nearly all cases you should never have to directly reference `DenseArray` or `SparseArray` directly. It knows which one to create based on whether the schema has `sparse=False` or `sparse=True`.

tiledb/cloud/notebook.py
```diff
@@ -1,10 +1,23 @@
...
+"""
+
+import posixpath
+import time
+from typing import Tuple
+
+import numpy
+
+import tiledb
+from tiledb.cloud import array, client, rest_api, tiledb_cloud_error
```
thetorpedodog: `isort .`

tiledb/cloud/notebook.py
```diff
@@ -35,3 +50,290 @@ def rename_notebook(
...
+    tiledb_uri_s3 = "tiledb://" + posixpath.join(namespace, storage_path, array_name)
+
+    # Create the (empty) array on disk.
+    tiledb.DenseArray.create(tiledb_uri_s3, schema)
+    tiledb_uri = "tiledb://" + posixpath.join(namespace, array_name)
+    time.sleep(0.25)
+
+    file_properties = {}
+
+    tiledb.cloud.array.update_info(uri=tiledb_uri, array_name=array_name)
```
thetorpedodog, line 300: Since you've `from tiledb.cloud import array`, you should use `array.update_info` here.

tiledb/cloud/notebook.py
```diff
@@ -35,3 +53,290 @@ def rename_notebook(
...
+                raise ValueError(
+                    f"Error creating file: {e}. Are your credentials valid?"
+                ) from e
+            if "Cannot create array" in str(e) and "already exists" in str(e):
+                raise ValueError(
+                    f"Error creating file: {array_name} already exists in namespace {namespace}."
+                )
+        # Retry other exceptions
+        tries -= 1
+    raise ValueError(f"Out of retries uploading {array_name} to namespace {namespace}.")
```
thetorpedodog, line 261: It would probably be best to let the last exception bubble out, probably by moving the final exception check inside the loop `except`:

```python
try:
  # ...
except TileDBException as e:
  if ...:
    ...
  if retries == 0:
    raise
```

On that last `raise`, we may specifically want to do the `check_exc` stuff that we do elsewhere in the codebase. I am not *super* familiar with the semantics or reason behind that so I have to defer to another reviewer for that.

thetorpedodog (commented):
Nothing major left; just a few nits. Some of these comments may be out of date because as I was reviewing you pushed a new version which fixed some of them, but I have no idea how to find the old ones and get rid of them.

thetorpedodog:
Also, annoyingly, it looks like your test failures are due to 502 errors :(

tiledb/cloud/notebook.py
johnkerl: this is not a function i touched -- it was the function that was here when i started. and its doc string is out of data. and that info does not appear in the array signature.

i'll (try to) clean this up while i'm here, even though i didn't introduce it.

tiledb/cloud/notebook.py
johnkerl: yup, black is good with either

tiledb/cloud/notebook.py
johnkerl: TIL! :)

tiledb/cloud/notebook.py
johnkerl: i have done so many times and i am now doing so on every commit.

* what formatting would you prefer?
* does isort give you something different?

tiledb/cloud/notebook.py
johnkerl: TIL! :)

tiledb/cloud/notebook.py
thetorpedodog: This was one of those comments where one of your follow-ons ended up fixing the problem and the comment became outdated; what appears in efdc478a30413c37d3ba466eab77c64621e497c4 is The Correct Format.

tiledb/cloud/notebook.py
johnkerl: i prefer to do that on a follow-on PR -- i've added that to ch7290 -- thanks!

tiledb/cloud/notebook.py
```diff
@@ -1,37 +1,338 @@
...
+                namespace,
+                dom,
+                ctx,
+            )
+            return (tiledb_uri, array_name)
+        except tiledb.TileDBError as e:
+            if "Error while listing with prefix" in str(e):
+                # It is possible to land here if user sets wrong default S3
+                # credentials with respect to default S3 path.
+                raise ValueError(
```
thetorpedodog, line 249:
```suggestion
                raise tiledb_cloud_error.TileDBCloudError(
```

(likewise below; this is not an appropriate situation for a `ValueError`; the value is perfectly valid, but it does not work given the environment. for comparison, `os.mkdir("")` might be a `ValueError`, since making an empty directory makes no sense, but `os.mkdir("/path/to/somewhere/that/already/exists")` is *not* a ValueError.)

thetorpedodog (approved):
This looks good! In a follow-up it might be good to add an automated test where we upload a notebook from a file, then download it to a different file, and perform a diff. That will also require a call to find the current user’s default storage path (since, ideally, we do not want to hardcode in our own unit test user’s default path).

If you want to wait for comments from others you may do so but (apart from my suggestion below) this is good to go imo.

In the interest of keeping an easier-to-understand change history I would recommend the “squash and merge” option when submitting.

tiledb/cloud/notebook.py
johnkerl: > In the interest of keeping an easier-to-understand change history I would recommend the “squash and merge” option when submitting.

100% my preference as well @thetorpedodog -- thanks! :)

PR #162, opened 2021-08-30
nonfunctional fixes to dag.py.

thetorpedodog:
- Changes to importing only modules and sorts imports.
- Removes redundant imports in visualization functions.
- Renames `__function`s to `_function`.
- Autogenerates `Status`'s `__str__`.
- Renames some Capitalizes Variables to lowercase.

---

This is mostly the prelude to my next change, because I wanted to keep the simple cleanup changes separate.

### DIFF
--- a/tiledb/cloud/dag/dag.py
+++ b/tiledb/cloud/dag/dag.py
@@ -1,44 +1,19 @@
+import json
 import numbers
 import time
 import uuid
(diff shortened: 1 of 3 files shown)


thetorpedodog:
pushing this back to the drafts because there are more maddening things going on with downstream stuff.

PR #163, opened 2021-08-30
Add support for stored parameter replacement

thetorpedodog:
This change pulls out `_replace_nodes_with_results` to the root level, then makes it more generic to support stored parameters.

Since the tree-walking process for replacing stored parameters is fundamentally the same as that for replacing Nodes, this extracts out `_ResultReplacer`, which implements the tree-walking algorithm, and implements replacing results and stored parameters in terms of that function. In a future change, it will also be used to implement replacing Nodes with `StoredParam`s (though this will require some work on the graph side to better keep track of node results).

---

As usual, this should be less scary than it looks, since much of it is (newly-added) test code.

### DIFF
--- a/tests/test_dag.py
+++ b/tests/test_dag.py
@@ -1,14 +1,20 @@
+import base64
 import collections
 import collections.abc as cabc
 import os
 import pickle
 import time
 import unittest
+import uuid
+import cloudpickle
 
 import numpy as np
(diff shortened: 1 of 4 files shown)


thetorpedodog:
this is currently based on #162 and will be re-pointed to master once that is in.

PR #165, opened 2021-08-31
Make imports consistent and clean up dag.py.

thetorpedodog:
After lots of puzzling import breakage over the past several changes, I found what I think is the culprit: a `dag.py` ⇄ `visualization.py` dependency cycle. This change first fixes that, and then goes about the task of making imports consistent.

- Sets up isort with the default black configuration + a one-import-per-line rule. I chose this for several reasons, among them that it means that you don't end up with something like:
- 
- Changes all relative imports to absolute imports (i.e. no more `from . import whatever`), since those are recommended against and in some style guides forbidden. Using absolute imports makes it always obvious where an import comes from without having to know the exact path you are already on.
- Lastly, does some tidying inside `dag.py`, to make it easier for the actual change I want to do.


thetorpedodog:
```
/opt/hostedtoolcache/Python/3.6.14/x64/lib/python3.6/site-packages/tiledb/cloud/dag/dag.py:9: in <module>
    import tiledb.cloud.dag.status as st
E   AttributeError: module 'tiledb' has no attribute 'cloud'
```

i am going to become the joker

thetorpedodog:
> ```
> /opt/hostedtoolcache/Python/3.6.14/x64/lib/python3.6/site-packages/tiledb/cloud/dag/dag.py:9: in <module>
>     import tiledb.cloud.dag.status as st
> E   AttributeError: module 'tiledb' has no attribute 'cloud'
> ```
> 
> i am going to become the joker

For future reference when some poor sap runs into this again, this is a [change between Python 3.6 and 3.7](https://docs.python.org/3/whatsnew/3.7.html#other-language-changes), fixing [Python bug #30024](https://bugs.python.org/issue30024).

ihnorton (approved):

thetorpedodog:
rebased atop #172 to rerun tests with less concurrency. now will not submit until after that is in.

thetorpedodog:
I think for now we need to just ignore transient server errors beyond our control, because they are happening continually, and GitHub does not offer the option to rerun a single entry of the test matrix (or even a single pipeline within one file). Once we stabilize the server we can make all test passes mandatory.

For now I think a good metric is provided that one runner for each Python version passes, and do a quick check of the failing test to ensure that the failure is transient (e.g. a 502).

PR #169, opened 2021-09-01
Convert from Azure Pipelines to GitHub actions

thetorpedodog:
Pretty self-explanatory from the commits and the description, yeah?

[ch10037]


Shelnutt2:
> Pretty self-explanatory from the commits and the description, yeah?
> 
> [ch10037]

bigger change is finally removing the dead pipfile! :tada:

.github/workflows/tiledb-cloud-py.yaml
```diff
@@ -0,0 +1,51 @@
+name: TileDB-Cloud-Py autobuild
+
+on:
+  push:
+    tags: [ "v*" ]
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
```
Shelnutt2, line 8: Do you want to run on all pull requests instead of only master? That way if you do a PR into a different branch it runs?

thetorpedodog: Good idea. Changed.

thetorpedodog:
This is, in theory, good to go. Unfortunately it looks a lot like the mysterious hangs are still happening, though now in a different place: after running everything in test_basic.py, and also only sometimes. This is an improvement at least?

My feelings are that, for now, we should drop the requirement that the whole test matrix passes so that we’re not completely blocked by test environment problems. We in the company will just have to run `pytest` locally in both py3.6 and py3.7 and note that we have done so in a comment. If the pipelines *do* finish, they will still give us some extra confidence. (@ihnorton, @nguyenv, @Shelnutt2: thoughts?)

I’m going to smash that “re-run jobs” button, eat lunch, and see if it has decided to behave once I get back.

ihnorton:
3.6 and 3.7 are the only versions supported by UDFs, so that's not a viable solution IMO. I would suggest reverting #157 on a branch to see if the issues are caused by those changes vs potential issues w/ newly-released versions of dependencies.

thetorpedodog:
figured it out. it was `test_tasks` the whole time. Azure Pipelines and the lack of output was just a red herring.

Added 2667c7352b0f67aee03a78ca2cb70eacfcccdb72 to the start of the change history and all the tests seem to be happily running.

On the plus side, getting off of Azure Pipelines and into the better-integrated GitHub Actions is still a net benefit.

thetorpedodog:
last attempt failed due to things that were Not Our (in the narrow sense of “this library’s”) Fault:

```
E           HTTP response headers: HTTPHeaderDict({'Date': 'Wed, 01 Sep 2021 18:50:29 GMT', 'Content-Type': 'text/html', 'Content-Length': '150', 'Connection': 'keep-alive'})
E           HTTP response body: <html>
E           <head><title>502 Bad Gateway</title></head>
E           <body>
E           <center><h1>502 Bad Gateway</h1></center>
E           <hr><center>nginx</center>
E           </body>
E           </html>
```

ihnorton:
> On the plus side, getting off of Azure Pipelines and into the better-integrated GitHub Actions is still a net benefit.

💯

thetorpedodog:
Everything passes! This is ready to go.

ihnorton:
What is the reason for the version-specific `requirements` files with so many pins? This is a customer-facing package, so ideally we should test against un-pinned versions because that's what new installs will  (usually) see, unless there is a known issue.

ihnorton:
Alternatively, I suppose we could have another matrix entry without pins - trying to understand the goal.

thetorpedodog:
> What is the reason for the version-specific `requirements` files with so many pins? This is a customer-facing package, so ideally we should test against un-pinned versions because that's what new installs will (usually) see, unless there is a known issue.

The `requirements-pyX.Y.txt` files were each generated by `pip freeze` so that we get a repeatable install if we use the requirements file (thus insulating us from test failures due to dependency changes). `pip install tiledb-cloud` (or anything which depends upon `tiledb-cloud`) will still fetch the latest versions of everything.

Added a matrix dimension for fresh install vs. pinned deps.

ihnorton (approved):
LGTM

thetorpedodog:
Removed outdated “expected” status checks in the repository settings; I’ll add the new ones once they start appearing in search (presumably sometime after this merge).

ihnorton:
One intermediate option as follow-up is to use the pins for PRs in general but add a nightly task which runs against (and possibly updates?) unpinned versions of everything on pypi. @nguyenv added that recently for tiledb-py to test against core ‘dev’ on a nightly basis. It is a nice balance of stability for PRs along with proactive warning when something breaks in a dep (https://github.com/TileDB-Inc/TileDB-Py/blob/dev/.github/workflows/daily-test-build.yml)

thetorpedodog:
I like that idea a lot! Once I get out of the stack of tasks I have going for myself I will look into it. Thank you!

PR #172, opened 2021-09-01
Remove "fresh" dependency strategy from the matrix.

thetorpedodog:
This reverts to essentially the version of the pipeline YAML file from
before we added the "fresh checkout" version. It seems like running too
many tests concurrently is causing us trouble where the server gives us
502s, so this is an attempt to reduce it. Limiting the number of runners
means that we would have to wait for two (or more) tests to complete
one after the other. In the context of a test pipeline, pinned
dependencies are more useful than a fresh install.

### DIFF
--- a/.github/workflows/tiledb-cloud-py.yaml
+++ b/.github/workflows/tiledb-cloud-py.yaml
@@ -24,18 +24,12 @@ jobs:
           - ubuntu-20.04
           - macos-10.15
         python-version: ["3.7", "3.6"]
-        dependencies:
-          - pinned
-          - fresh
         include:
           - os: ubuntu-20.04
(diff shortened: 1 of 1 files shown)


thetorpedodog:
I think for a fresh install we will have to go with the strategy from https://github.com/TileDB-Inc/TileDB-Cloud-Py/pull/169#issuecomment-910716803, because right now trying it on every PR appears to be causing too much flakiness.

thetorpedodog:
ha ha ha it did not work ha ha ha

ihnorton (approved):

thetorpedodog:
this takes it from “almost never works” to “sometimes works” so it’s going in

PR #173, opened 2021-09-02
Further reduce autotest concurrency and add a nightly build.

thetorpedodog:
This stops running automated on-push tests on Mac OS (for now).
To make up for this loss, it adds a nightly test pipeline that runs
our full matrix of tests, the cross product of:

- Mac OS and Linux
- Python 3.6 and 3.7
- requirements.txt and a fresh `pip install`

These are limited to one run at a time, since time is no object when
running tests at night.

---

thank you to @nguyenv for the original idea and @ihnorton for pointing it out to me!

### DIFF
--- a/.github/workflows/commit-check.yaml
+++ b/.github/workflows/commit-check.yaml
@@ -1,4 +1,4 @@
-name: TileDB-Cloud-Py autobuild
+name: commit-check
 
 on:
   push:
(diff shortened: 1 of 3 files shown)


thetorpedodog:
Added a second change to (hopefully?) build a wheel for uploading to PyPI after running the full battery of tests. It requires a `PYPI_API_TOKEN` secret to be set on the repository once committed.

Shelnutt2:
> Added a second change to (hopefully?) build a wheel for uploading to PyPI after running the full battery of tests. It requires a `PYPI_API_TOKEN` secret to be set on the repository once committed.

@ihnorton or I can provide one tomorrow.

thetorpedodog:
The errors we were running into were caused by a bug in core that was fixed in https://github.com/TileDB-Inc/TileDB/pull/2451. Because there are no longer test errors, withdrawing this change.

PR #174, opened 2021-09-07
Re-enable the full testing matrix and add PyPI step.

thetorpedodog:
- Re-enables the full testing matrix to run, concurrently.
- Adds nightly build to check for dependency regressions.
- Adds workflow to upload to PyPI when a release is created.

### DIFF
--- a/.github/workflows/tiledb-cloud-py.yaml
+++ b/.github/workflows/tiledb-cloud-py.yaml
@@ -1,11 +1,15 @@
-name: TileDB-Cloud-Py autobuild
+name: autobuild
 
 on:
-  push:
-    tags: [ "v*" ]
-    branches: [ master ]
   pull_request:
     branches: [ "*" ]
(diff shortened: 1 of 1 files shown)


thetorpedodog:
Bringing this back but running FRESH deps on NIGHTLY builds (to check for dependency problems) and PINNED deps on COMMITS (for a consistent environment).

ihnorton (approved):

ihnorton:
Oh - not directly related, but: at some point we should probably move the requirements-py* files under a subdir with any other CI things, and/or add a MANIFEST to exclude them from the final wheel.

thetorpedodog:
> Oh - not directly related, but: at some point we should probably move the requirements-py* files under a subdir with any other CI things, and/or add a MANIFEST to exclude them from the final wheel.

Putting this On The List along with changing the test API key environment variable.

PR #177, opened 2021-09-10
Add step to upload to test PyPI on merges to the main branch.

thetorpedodog:


ihnorton (approved):
Neat! Not sure if it's necessary to run on every merge to main, but probably fine for now -- your call. If we start sending 10s of packages a day to testpypi, then should switch to a time- or tag-based upload trigger.

(cc @nguyenv would be nice to do this in nightlies for tiledb-py -- once we are building releases on GHA)

PR #178, opened 2021-09-10
Enable CodeQL analysis.

thetorpedodog:
Simple CodeQL setup based entirely on the generated sample.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,44 @@
+name: "codeql-analysis"
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ master ]
+  schedule:
+    # Time randomly-selected by the autogenerator to spread load.
+    - cron: '29 17 * * 3'
+
+jobs:
+  analyze:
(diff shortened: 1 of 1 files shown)


thetorpedodog:
As I was clicking through some stuff managing actions, I saw this suggestion. I decided to give it a shot; we can take it or leave it.

ihnorton (approved):
Does this do anything by default? I read the docs for a few minutes and it seems to be all about writing your own queries; do they have a set of built-in queries that are run by default?

Anyway, it doesn't look like it runs until the file is checked in on the main branch, so I'm fine with merging this to see what happens. Assuming theere is some set of default rules, I suspect it may be fairly noisy with the generated REST API code.

thetorpedodog:
[CodeQL includes a default set of analyses](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/about-code-scanning-with-codeql#about-codeql-queries) (“GitHub experts, security researchers, and community contributors write and maintain the default CodeQL queries used for code scanning.”) and they are clean as of right now, which seems good. It also does a dependency-install so that it can theoretically catch new issues introduced in dependencies, which seems useful, particularly on the weekly job if we have no other changes.

PR #181, opened 2021-09-21
Make the primary implementations of UDFs synchronous.

thetorpedodog:
Currently, the canonical implementation of UDFs is to use the generated
API’s asynchronous interface, and the synchronous version wraps that
and gets the result of the `multiprocessing.pool.TaskResult` returned by
the generated code. This will not work for our plans to avoid
round-tripping all task results when running DAGs.

This change instead makes a new `whatever_base` as the canonical base
implementation of a UDF caller, and bases the existing `whatever` and
`whatever_async` functions off of them, maintaining compatibility.
Additionally, we manage our own thread pool for async requests, which
avoids us having to use multiprocessing’s poorly-supported `TaskResult`
interface and gives us better control over how many threads we run.
UDF request management is fully moved into the `Client` object.

Results are returned by a new `results.Response` class, which includes
the important `task_id` information that will be needed to avoid
round-tripping requests in the future, and a `results.AsyncResponse`
class also wraps (a `futures.Future` of) that `Response` to provide
an API that is 100% compatible for callers currently expecting a
`UDFResult` or `SQLResult`.

Additional changes:

- Creates exactly 1 `ApiClient` to better reuse network connections.
- Cleans up API instance creation.

### DIFF
--- a/tests/test_dag.py
+++ b/tests/test_dag.py
@@ -3,6 +3,7 @@ import collections
 import collections.abc as cabc
 import os
 import pickle
+import threading
 import time
 import unittest
 import uuid
@@ -154,6 +155,23 @@ class DAGClassTest(unittest.TestCase):
         end_results = d.end_results_by_name()
         self.assertEqual(1, len(end_results))
         self.assertEqual(node_6.result(), end_results[node_6.name])
+
+    def test_concurrency(self):
+        task_count = 5
+        barrier = threading.Barrier(task_count)
+
+        def rendezvous(val):
+            barrier.wait()
+            return val
+
+        d = dag.DAG()
+        tasks = [
+            d.add_node(rendezvous, f"#{i}", local_mode=True) for i in range(task_count)
+        ]
+        terminal = d.add_node(", ".join, tasks, local_mode=True)
+        d.compute()
+        d.wait(1)  # Avoid locking up forever if we deadlock.
+        self.assertEqual("#0, #1, #2, #3, #4", terminal.result())
 
 
 class DAGFailureTest(unittest.TestCase):
@@ -395,19 +413,18 @@ class DAGCloudApplyTest(unittest.TestCase):
         self.assertEqual(d.status, dag.Status.COMPLETED)
 
 
-status_updates = 0
-done_updates = 0
-
-
 class DAGCallbackTest(unittest.TestCase):
+    def setUp(self) -> None:
+        super().setUp()
+        self.status_updates = 0
+        self.done_updates = 0
+
     def test_simple_dag(self):
         def status_callback_test(dag):
-            global status_updates
-            status_updates += 1
+            self.status_updates += 1
 
         def done_callback_test(dag):
-            global done_updates
-            done_updates += 1
+            self.done_updates += 1
 
         d = dag.DAG(
             update_callback=status_callback_test, done_callback=done_callback_test
@@ -425,10 +442,8 @@ class DAGCallbackTest(unittest.TestCase):
         self.assertEqual(node_1.result(), 2)
         self.assertEqual(node_2.result(), 4)
         self.assertEqual(node_3.result(), 8)
-        global status_updates
-        global done_updates
-        self.assertEqual(status_updates, 5)
-        self.assertEqual(done_updates, 1)
+        self.assertEqual(self.status_updates, 5)
+        self.assertEqual(self.done_updates, 1)
 
 
 class CustomSeq(cabc.MutableSequence):
--- a/tiledb/cloud/client.py
+++ b/tiledb/cloud/client.py
@@ -435,19 +435,47 @@ def send_udf_call(
     api_func: Callable[..., urllib3.HTTPResponse],
     api_kwargs: Dict[str, Any],
     decoder: results.AbstractDecoder,
+    id_callback: Optional[IDCallback] = None,
     *,
-    id_callback: IDCallback,
-    store_results: bool,
+    results_stored: bool,
 ) -> results.Response:
-    """Sends an API request and returns the results in the form of a Response."""
+    """Synchronously sends a request to the given API.
+
+    This handles the boilerplate parts (exception handling, parsing, response
+    construction) of calling one of the generated API functions for UDFs.
+    It runs synchronously and will return a :class:`results.Response`.
+    To run the same function asychronously, use
+    :meth:`Client.wrap_async_base_call` around the function that calls this
+    (by convention, the ``whatever_api_base`` functions).
+
+    This should only be used by callers *inside* this package.
+
+    :param api_func: The UDF API function that we want to call from here.
+        For instance, this might be :meth:`rest_api.SqlApi.run_sql`.
+    :param api_kwargs: The arguments to pass to the API function as a dict.
+        This should only include the parameters you want to send to the server,
+        *not* any of the “meta” parameters that are mixed in with them (e.g.
+        ``_preload_content``; this function will correctly set up the request).
+    :param decoder: The Decoder to use to decode the response.
+    :param id_callback: When the request completes (either by success or
+        failure), this will be called with the UUID from the HTTP response,
+        or None if the UUID could not be parsed.
+    :param results_stored: A boolean indicating whether the results were stored.
+        This does *not affect* the request; the ``store_results`` parameter of
+        whatever API message the call uses must be set, and this must match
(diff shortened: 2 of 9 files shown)


tiledb/cloud/client.py
```diff
@@ -375,148 +372,154 @@ def find_organization_or_user_for_default_charges(user):
...
+    *,
+    id_callback: IDCallback,
+    store_results: bool,
+) -> results.Response:
+    """Sends an API request and returns the results in the form of a Response."""
+    try:
+        http_response = api_func(_preload_content=False, **api_kwargs)
+    except rest_api.ApiException as exc:
+        id_callback(results.extract_task_id(exc))
+        raise tiledb_cloud_error.check_exc(exc) from exc
```
ihnorton, line 447: In the previous code, `raise X from None` was specifically used to avoid including the whole local rest call stack for remote errors; it makes them unreadable. AFAICT `check_exc` will already raise `from exc` when appropriate: https://github.com/TileDB-Inc/TileDB-Cloud-Py/blob/81c07150fd70f60a674794559bc20ae42678048e/tiledb/cloud/tiledb_cloud_error.py#L29

tiledb/cloud/client.py
```diff
@@ -375,148 +372,154 @@ def find_organization_or_user_for_default_charges(user):
...
-                configuration=config.config, pool_threads=self.pool_threads
-            )
-        )
+        self.retry_mode(RetryMode.FORCEFUL)
+
+    def retry_mode(self, mode: RetryOrStr = RetryMode.DEFAULT):
+        """Sets how we should retry requests and updates API instances."""
+        mode = RetryMode.maybe_from(mode)
+        config.config.retries = _RETRY_CONFIGS[mode]
+        client = rest_api.ApiClient(config.config)
```
ihnorton, line 495: Is this going to initialize another (multiprocessing) ThreadPool? https://github.com/TileDB-Inc/TileDB-Cloud-Py/blob/641edd135d7c59c40544a1e8826acfd44e2fe392/tiledb/cloud/rest_api/api_client.py#L112

or is that code lazy and never-called?

tiledb/cloud/client.py
```diff
@@ -375,148 +372,154 @@ def find_organization_or_user_for_default_charges(user):
...
+    store_results: bool,
+) -> results.Response:
+    """Sends an API request and returns the results in the form of a Response."""
+    try:
+        http_response = api_func(_preload_content=False, **api_kwargs)
+    except rest_api.ApiException as exc:
+        id_callback(results.extract_task_id(exc))
+        raise tiledb_cloud_error.check_exc(exc) from exc
+
+    task_id = results.extract_task_id(http_response)
```
ihnorton, line 449: Do we want to catch here? This function will throw if the future throws -- if I'm following correctly, when you get the result then any exception on the "thread" will propagate.

tiledb/cloud/client.py
```diff
@@ -375,148 +372,154 @@ def find_organization_or_user_for_default_charges(user):
...
+        self.udf_api = rest_api.UdfApi(client)
+        self.user_api = rest_api.UserApi(client)
+
+    def set_threads(self, threads: Optional[int] = None):
+        """Updates the number of threads in the async thread pool."""
+        with self._pool_lock:
+            old_pool = getattr(self, "_thread_pool", None)
+            self._thread_pool = futures.ThreadPoolExecutor(threads)
+            if old_pool:
+                old_pool.shutdown()
```
ihnorton, line 511: This function is async right? The docs say that it blocks until all futures are done, so presumably it yields, but that's not explicitly clear.

ihnorton (commented):
Aside from the comments/questions inline, two global concerns:
- do we have a test that launches multiple async calls and ensures they actually ... run async?
- how does the ThreadPoolExecutor future timeout interact with the http stack - eg what happens to the http call if the future times out?

tiledb/cloud/client.py
thetorpedodog: Fixed.

tiledb/cloud/client.py
thetorpedodog: This is the synchronous version, and `extract_task_id` is guaranteed never to throw anything; it will either return None or a UUID.

I’ve expanded the documentation on this function to make it a bit clearer what is going on.

tiledb/cloud/client.py
thetorpedodog: It is, yes, but:

- The threads are only ever actually started when a request is made, so it is lazy in that sense. Since we never use the async functionality built into the API client, they will never be started.
- The previous code created one thread pool for every individual API instance, so we still end up creating fewer never-used thread pools.

tiledb/cloud/client.py
thetorpedodog: Whoops! I should have included `wait=False` here, so that it just tells the old pool to shut down. Thank you!

thetorpedodog (commented):
> do we have a test that launches multiple async calls and ensures they actually ... run async?

At this point, we do not (and never have). I have an idea for how to implement it…after lunch.

> how does the ThreadPoolExecutor future timeout interact with the http stack - eg what happens to the http call if the future times out?

It’s identical to the way the previous version using the `multiprocessing.pool` implementation worked—the `Future`, like an `AsyncResult`, is (with the exception of `.cancel()`) exclusively a one-way communication. In both instances, the `timeout` parameter simply means that the *thread asking for the result* will quit waiting; the *thread sending the request* will continue making the request, and any other thread (or even the same thread, in a later `.result()` call) which has a reference to that Future will be able to ask it for results. (`.cancel()` can’t affect a request already in flight, because it can only [prevent the horse from leaving the barn](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future.set_running_or_notify_cancel).)

tiledb/cloud/client.py
ihnorton: 👍 thanks

tiledb/cloud/client.py
ihnorton: Won't this line throw if there's an exception on the thread? https://github.com/TileDB-Inc/TileDB-Cloud-Py/pull/181/files#diff-ab11ae7abf2b52119b99ba2775df89d3baeaf5e7062dc8cb131d911bc9033c5bR118

x-ref: https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future.result
> If the call raised an exception, this method will raise the same exception.

tests/test_dag.py
```diff
@@ -155,6 +156,23 @@ def test_end_nodes_dag(self):
         self.assertEqual(1, len(end_results))
         self.assertEqual(node_6.result(), end_results[node_6.name])
 
+    def test_concurrency(self):
```
ihnorton, line 159: Thanks!

tiledb/cloud/client.py
thetorpedodog: This *is* the thread; when we make a call without `_async_req=True` (and with `_preload_content=False`), the generated API returns an immediate `urllib3.HTTPResponse`, not a `futures.Future[urllib3.HTTPResponse]`. (Out of all the things I hate about the generated code, the fact that providing different parameters *wildly* changes the type it returns, and in *incredibly* poorly-documented ways, is probably the worst.)

On the line you’re referencing, both `HTTPResponse` and `ApiException` are guaranteed to have a `.headers` field which is (or at least can be treated as) an `Optional[Mapping[str, str]]`.

ihnorton: I see, thanks.

Shelnutt2 (approved):
Great work!

Shelnutt2:
> LGTM, pending CI and @Shelnutt2 clarification on whether there's any need to maintain the `SQLResult`/`UDFResult` object identity of the results (eg as subclasses of `AsyncResponse`).

I don't think we need to maintain the object, I'm fine removing them in this case.

thetorpedodog:
> > LGTM, pending CI and @Shelnutt2 clarification on whether there's any need to maintain the `SQLResult`/`UDFResult` object identity of the results (eg as subclasses of `AsyncResponse`).
> 
> I don't think we need to maintain the object, I'm fine removing them in this case.

Specifically:

- We never pickle them, so there is no need to maintain front/backend compatibility
- Users have no particular reason to instantiate them themselves, so they are unlikely to have any references to those names anywhere else

PR #184, opened 2021-09-23
At long last, actually substitute in stored parameters.

thetorpedodog:
Here it is, the monster change that actually implements stored parameters. A major reason that it's so complicated is the fact that `Node` and company have a very large API surface, so users may access it in any number of good or bad ways.

Most of what else goes here would be essentially a description of what is in the commit messages, so you should probably just read those. It is likely that there are things that are unclear, but at least the tests all pass?

### DIFF
--- a/tiledb/cloud/dag/dag.py
+++ b/tiledb/cloud/dag/dag.py
@@ -3,11 +3,23 @@ import numbers
 import time
 import uuid
 from concurrent import futures
-from typing import Any, Dict, Optional, Tuple
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    FrozenSet,
+    Generic,
+    Optional,
+    Set,
+    Tuple,
+    TypeVar,
+    Union,
+)
 
 import networkx as nx
 
 from tiledb.cloud import array
+from tiledb.cloud import results
 from tiledb.cloud import sql
 from tiledb.cloud import tiledb_cloud_error as tce
 from tiledb.cloud import udf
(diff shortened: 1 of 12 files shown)


ihnorton (commented):
> This setup is unfortunate but is necessary to provide API compatibility
with callers that were expecting to be able to call `some_node.future`.

Could we make `Node.future` an `@property` function instead?

tiledb/cloud/dag/dag.py
```diff
@@ -79,8 +121,10 @@ def depends_on(self, node):
 
     def cancel(self):
         self.status = st.Status.CANCELLED
-        if self.future is not None:
-            self.future.cancel()
+        with self._future_lock:
```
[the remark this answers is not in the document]
thetorpedodog: I think before the introduction of retry logic, this made sense, since the only invariant `_future_lock` was that both `_real_future` and `future` were both set up together and consistently. There was also a time where `_future_lock` was a more comprehensive lifecycle lock, and some of the operations that the callback would have called themselves would have acquired the `_future_lock`, which meant that we needed to release it (indicating that the invariants asserted by the lock held).

However, post–retry logic, that was no longer true and we should have indeed held onto this lock while calling `rft.cancel()`, because otherwise we might swap out the `_real_future` before this could happen.

That said, this entire thing is obsolete because it’s all much simpler now.

thetorpedodog (commented):
> > This setup is unfortunate but is necessary to provide API compatibility
> > with callers that were expecting to be able to call `some_node.future`.
> 
> Could we make `Node.future` an `@property` function instead?

The way it was implemented, no, because we would have needed to keep a handle on all generated futures anyway because of retry logic.

ihnorton:
> That said, this entire thing is obsolete because it’s all much simpler now.

💯

tiledb/cloud/decoders.py
```diff
@@ -0,0 +1,60 @@
...
+from typing import Any, Generic, TypeVar
+
+import cloudpickle
+import pandas
+import pyarrow
+
+from tiledb.cloud import tiledb_cloud_error as tce
+from tiledb.cloud.rest_api import models
+
+_T = TypeVar("_T")
```
ihnorton, line 15: These `TypeVars` are effectively parametrized over the three result types, right? Not sure if it's necessary, but wonder if bounding the typevar would be useful from a documentation standpoint.

tiledb/cloud/results.py
```diff
@@ -127,3 +132,20 @@ def _maybe_uuid(id_str: Optional[str]) -> Optional[uuid.UUID]:
         return uuid.UUID(hex=id_str)
     except ValueError:
         return None
+
+
+# Compatibility aliases.
+# Until a version with decoders.py is deployed to the UDF environment,
```
ihnorton, line 138: We might need to maintain these aliases for longer due to stored UDFs...

tiledb/cloud/dag/dag.py
```diff
@@ -120,37 +133,60 @@ def finished(self):
...
+        assert self.future  # Needed to silence mypy.
         self.future.add_done_callback(self._handle_completed_future)
 
+    def _do_exec(self) -> _T:
+        if self._is_remote:
+            # Stored parameters work only on remote functions.
+            (args, kwargs), seen = _replace_nodes_with_stored_params(
+                (self.args, dict(self.kwargs))
+            )
+            # TODO: We can't use stored params in SQL queries.
```
ihnorton, line 156: (making sure there is a story for this, especially if it might break some existing use-case?)

ihnorton (approved):
Some minor (NFC) questions for my own clarification. :shipit:

ihnorton:
Nice work 💯

tiledb/cloud/decoders.py
thetorpedodog: The parameter is actually intended to be the type from the user’s function. For instance, if you do

```python
def my_func(*args, **kwargs) -> MyType: ...

n = dag.submit_udf(my_func, ...)
# n is a Node[MyType]
# n._real_result should be a Result[MyType].
# n._real_result.decoder will be an AbstractDecoder[MyType].
```

Because of the way `result_format` works, this is not perfect. If `n` is serialized as a Pickle, everything works as expected. However, if `n` is serialized as Arrow, it should be a `pyarrow.Table` and if it’s JSON, it is pretty much an `Any`.

It’s…not impossible, but extremely verbose to try and get this to 100% correctness based upon all the layers of function calls with `*args` and `**kwargs` that we have going on. This 80–90% solution was at least good enough to help me out as I was implementing the internals, so I think it should provide at least *some* value to users.

tiledb/cloud/results.py
thetorpedodog: These never need to be handled directly by user code, so they should not need to stay around. I have a plan to seamlessly remove these before officially releasing this feature that I will detail more on internal channels.

thetorpedodog (commented):
Thanks again! I decided to fix the one bug you pointed out rather than do it in a follow-up, so please take another look.

tiledb/cloud/decoders.py
ihnorton: Ok... TBH this feels a little odd for python, but if the type-checker handles this then I'll trust you that it's useful 😄

ihnorton (approved):
Thanks for the fix! 👍

thetorpedodog:
Rebased with no code changes to get a nicer history (and verify no conflicts) before merging.

PR #188, opened 2021-09-23
Switch to VFS I/O for notebook upload/download

johnkerl:
Ask from @Shelnutt2 on sc10506


tiledb/cloud/notebook.py
```diff
@@ -71,7 +71,8 @@ def download_notebook_to_file(
         tiledb_uri,
         storage_credential_name,
     )
-    with open(ipynb_file_name, "w") as handle:
+    vfs = tiledb.VFS()
```
ihnorton, line 74: Probably need to pass the ctx into the VFS constructor, looks like some previous open calls may have been missing it.

johnkerl: thanks @ihnorton !! can you tell me more about what you mean by "previous" -- ?

ihnorton: Sorry, I missed that these calls are for local download - please disregard.

tiledb/cloud/notebook.py
```diff
@@ -337,7 +339,9 @@ def _write_notebook_to_array(
 
     contents_as_array = numpy.array(bytearray(ipynb_file_contents, CHARACTER_ENCODING))
 
-    with tiledb.open(tiledb_uri, mode="w", ctx=ctx) as arr:
+    vfs = tiledb.VFS(tiledb.cloud.Ctx().config())
+
+    with vfs.open(tiledb_uri, mode="w", ctx=ctx) as arr:
```
ihnorton, line 344: `vfs.open` returns a FileIO object, not an array, so the assignments below will not work -- this needs to be `tiledb.open` to return an array object.

tiledb/cloud/notebook.py
```diff
@@ -67,7 +67,8 @@ def download_notebook_to_file(
     ipynb_file_contents = download_notebook_contents(
         tiledb_uri,
     )
-    with open(ipynb_file_name, "w") as handle:
```
ihnorton, line 70: This is fine as-is, IMO -- sorry about the distraction with the earlier comment. I missed that this is a local FS operation only.

tiledb/cloud/notebook.py
```diff
@@ -135,7 +136,8 @@ def upload_notebook_from_file(
     :return: TileDB array name, such as "tiledb://janedoe/testing-upload".
     """
 
-    with open(ipynb_file_name, "r") as handle:
+    vfs = tiledb.VFS(tiledb.cloud.Ctx().config())
```
ihnorton, line 139: same as above

ihnorton (changes requested):

tiledb/cloud/notebook.py
johnkerl: sorry @ihnorton it looks like i was rushing through this a bit :(

ihnorton (approved):

johnkerl:
thanks @ihnorton ! :)

PR #191, opened 2021-10-05
Downgrade urllib3 to v1.25.11.

thetorpedodog:
An upstream dependency of `scikit-mobility==1.1.2` in the UDF
environment forbids urllib3 v1.26, but the `allowed_methods` parameter
was introduced in that version. We need to downgrade our requirement
and switch back to the deprecated method name.


setup.py
```diff
@@ -18,7 +18,9 @@
 
 REQUIRES = [
     "tiledb>=0.5.0",
-    "urllib3>=1.26",
+    # One of our upstream dependencies needs urllib<1.26.
+    # This should be relaxed once that is fixed.
+    "urllib3==1.25.11",
```
Shelnutt2, line 23: Should this be a wildcard on `1.25` or do we need the specific `1.25.11`?

Shelnutt2 (approved):

setup.py
thetorpedodog: Possibly, but I don't think they will be coming out with any more 1.25s and this is (hopefully) a temporary measure.

gsakkis:
After going down a rabbit hole, it turns out the root cause for `urllib3<1.26` is the `requests==2.22.0` which comes as part of the grandparent image `continuumio/miniconda3:4.8.2`. Possible fixes:
- Require latest/recent `requests` (easier/safer)
- Upgrade the base miniconda3 image (needs more work/testing)

Btw the base udf image may be (strictly speaking) broken too based on this line from the [logs](https://storage.googleapis.com/gitlab-gprd-artifacts/b8/da/b8da20d260161eda98ff165bc169c648e5d49c9cd727e9c929ecfe6914b696d0/2021_10_05/1651052522/1784468415/job.log?response-content-type=text%2Fplain%3B%20charset%3Dutf-8&response-content-disposition=inline&GoogleAccessId=gitlab-object-storage-prd@gitlab-production.iam.gserviceaccount.com&Signature=2Z%2BZL2lhaHwE%2FIITe2ZyT0i6gJuSTMq1HUEoKW0e0ruIFkPCN1Lr5lu%2F0ORX%0AW5DBzuZZyKBVFVRu8KLx%2BGdsZB9P5wd%2BYpX69AKS2ZRFtthkAYBcbXgipKJp%0Aw3ZWp%2F2cNVzfBXNKKAlQ%2BWqLtlgIJpe5CoA8DPqNgoZUM6FYU8RqUzreKdjM%0AvwFsicTyZ1prVX03JfsmJMeP5AeGKVfNRneXGiJyg0TFsXgLL6sxWeWam%2Bie%0Ai%2BYSwL7igPVYHsVdkxq9ZxERsKGJJkS%2B6E0nnsZVNFMk%2Fas8vWxqT34EEMIZ%0AFD0l2Rck8YcQidgtM%2FBW%2BLlTFDQzgPnRyWzvIHUlrg%3D%3D&Expires=1633451642): 
> ERROR: requests 2.22.0 has requirement urllib3!=1.25.0,!=1.25.1,<1.26,>=1.21.1, but you'll have urllib3 1.26.7 which is incompatible.

Though despite this error/warning, it doesn't actually upgrade urllib3.

PR #199, opened 2021-11-03
Expand urllib3 connection pool along with the thread pool.

thetorpedodog:
If a user expanded the size of the client's async threadpool and then
made enough requests to completely fill it, it would result in
`urllib3.connectionpool` logging warnings.  It won't cause any actual
problems, but it is still probably better to expand the pool to avoid
running into that condition.


Shelnutt2 (approved):
Quick fix, thanks!

PR #200, opened 2021-11-03
Update DAGs to not download intermediate results by default.

thetorpedodog:
This enables callers to not actually download the results of a
server-side call with a new `_download_results` parameter to each of the
`whatever_base` functions. If set, it will tell the server not to send
the results of the function when it is complete; just to return headers.
(This is useful even *before* the server has support for not sending
results back—we will get the HTTP response and completely ignore
the contents, meaning that the dataset will not be parsed into memory.)

This requires restructuring of the Node class itself, but maintains
(much of) the same external interface.

The actual future used by the instance internally is privatized and now
wraps a `Result` rather than the object that the UDF call would have
produced. The (not-quite official, but still present) external-facing
`.future` is now a property that materializes a proxy `Future` that
receives the result of the actual Future.

---

There is some weirdness in the way we handle concurrency and cancellation right now but those are no worse than the existing implementation, just slightly different.


thetorpedodog:
Bump on this—this is a bug we’ve had open for a while.

tiledb/cloud/_results/results.py
```diff
@@ -46,14 +43,27 @@ def wrap(cls, func: Callable[..., _T]) -> Callable[..., Result[_T]]:
...
     """A response from running a UDF remotely."""
 
     def get(self) -> _T:
         """Decodes the response from the server."""
         try:
-            return self.decoder.decode(self.body)
+            with self._result_lock:
+                if self._decoded is _SENTINEL:
+                    if self._body is not None:
+                        self._decoded = self.decoder.decode(self._body)
```
ihnorton, line 61: Do we need to keep `_body` after decoding?

tiledb/cloud/dag/dag.py
```diff
@@ -153,57 +158,68 @@ def exec(self, namespace: Optional[str] = None):
...
+            ft.cancel()
+        self._future = ft
+        ft.add_done_callback(self._handle_completed_future)
+
+    def _do_exec(self) -> results.Result[_T]:
+        # We have to make a shallow copy of kwargs here.
+        # Since we modify the kwargs dictionary here before passing it
+        # to the wrapped function, we need to ensure that we don't mutate
+        # our stored kwargs.
+        raw_kwargs = dict(self.kwargs)
```
ihnorton, line 175: Based on the comment above, don't we need a deep copy here?

tiledb/cloud/client.py
```diff
@@ -561,21 +499,17 @@ def set_threads(self, threads: Optional[int] = None):
             self._thread_pool = futures.ThreadPoolExecutor(
                 threads, thread_name_prefix="tiledb-async-"
```
ihnorton, line 500: TIL `thread_name_prefix` 👍

tiledb/cloud/_results/results.py
```diff
@@ -68,14 +78,54 @@ def to_stored_param(self) -> stored_params.StoredParam:
...
 
-    # The HTTP content of the body that was returned.
-    body: bytes
-    # The server-generated UUID of the task.
-    task_id: Optional[uuid.UUID]
-    # The decoder that was used to decode the results.
-    decoder: decoders.AbstractDecoder[_T]
-    # True if the results were stored, false otherwise.
-    results_stored: bool
+    task_id = attr.ib(type=Optional[uuid.UUID])
```
ihnorton, line 81: Mostly out of curiosity, why are you using `.ib` rather than `field` + type annotation here?

tiledb/cloud/_results/results.py
```diff
@@ -68,14 +78,54 @@ def to_stored_param(self) -> stored_params.StoredParam:
...
+
+    If None, the result was not downloaded (yet).
+    """
+
+    @_body.validator
+    def _validate(self, attribute, value):
+        del attribute, value  # unused
+        if self.results_stored and not self.task_id:
+            raise ValueError("task_id must be set for stored results")
+        if not self.results_stored and not self._body:
```
ihnorton, line 101: Is the validator called (1) only in the constructor and (2) after `_body` is actually set on the object?

ihnorton (commented):
AFAICT this is using attrs for two classes, providing:
- automatic `__init__` (also available w/ dataclasses IIUC)
- automatic `__eq__` (not available w/ dataclasses)
- automatic execution of `_body` validator (not available w/ dataclasses)

Are there ways it is implicitly used?

The dependency is fine with me, but please continue to be judicious about dependencies here.

tiledb/cloud/_results/results.py
thetorpedodog: We do not! Good idea.

tiledb/cloud/_results/results.py
thetorpedodog: Mostly because at the time, `attr.ib` was the primary API for attrs. I switched it over to the modern version.

tiledb/cloud/_results/results.py
thetorpedodog: This is called only at initialization time.

tiledb/cloud/dag/dag.py
thetorpedodog: In this case, no; our concerns are only about the arguments we add to the dictionary (`_download_results` and `stored_param_uuids`). I updated the comment to be a bit more precise here.

thetorpedodog (commented):
Good questions!

> AFAICT this is using attrs for two classes, providing:
>
> - automatic __init__ (also available w/ dataclasses IIUC)
> - automatic __eq__ (not available w/ dataclasses)
> - automatic execution of _body validator (not available w/ dataclasses)
>
> Are there ways it is implicitly used?
> 
> The dependency is fine with me, but please continue to be judicious about dependencies here.

Yes, those are the ways we use attrs; it's all explicit. One of the considerations I used when deciding to make this switch was that at least one of our dependencies already itself depends upon attrs, so directly depending upon it does not actually increase our net dependency count.

tiledb/cloud/_results/results.py
ihnorton: What I'm wondering is when `_body` is set here -- does attrs set it first _and then_ call the validator?

tiledb/cloud/dag/dag.py
ihnorton: Got it, ok, thanks.

tiledb/cloud/_results/results.py
thetorpedodog: Oh, yes. Validators are run only after the rest of the `__init__` method has completed, so at that point the class is fully set up, and validation ensures that the setup is correct before it is returned to the caller.

tiledb/cloud/_results/results.py
ihnorton: Alright, your call

tiledb/cloud/_results/results.py
ihnorton: Thanks

ihnorton (approved):
Approving, but please give the tests a once-over and see if there's anywhere else we can add coverage here. It doesn't _seem_ like the scope of changes here is fully reflected in the changes to `test_dag.py`. But that might be due to sufficient coverage already?

thetorpedodog:
> Approving, but please give the tests a once-over and see if there's anywhere else we can add coverage here. It doesn't _seem_ like the scope of changes here is fully reflected in the changes to `test_dag.py`. But that might be due to sufficient coverage already?

We have, to some degree, already covered this with our tests of running server-side, but nothing specifically tested what happened when we tried not to download. Added a test for this.

Thanks for the great review!

PR #203, opened 2021-11-15
Use configuration provided Retry() instance on redirects

dimos-tdb:


antalakas (commented):
Looks good to me, really unsure if there is a better way

tiledb/cloud/pool_manager_wrapper.py
```diff
@@ -0,0 +1,10 @@
+class PoolManagerWrapper:
```
thetorpedodog, line 1: This class is only ever going to be useful in `client.py` and should not be consumed by external users, so it would make sense to define it as `_PoolManagerWrapper` in that file, or rename this module to start with an underscore too.

I would also add a docstring, explaining why this exists and the fact that it only implements the one method of `PoolManager` that the generated API code actually uses.

tiledb/cloud/client.py
```diff
@@ -13,6 +13,7 @@
 from tiledb.cloud import tiledb_cloud_error
 from tiledb.cloud._results import decoders
 from tiledb.cloud._results import results
+from tiledb.cloud.pool_manager_wrapper import PoolManagerWrapper
```
thetorpedodog, line 16: nit: in general this code now uses the convention of "import modules, not things within modules", so `from tiledb.cloud import pool_manager_wrapper` would be the thing to do here (with the consideration of the following comment…)

tiledb/cloud/pool_manager_wrapper.py
```diff
@@ -0,0 +1,10 @@
...
+    def __init__(self, pool_manager):
+        self._pool = pool_manager
+
+    def request(self, method, url, **kwargs):
+        # https://github.com/urllib3/urllib3/issues/2475
+        kwargs["retries"] = self._pool.connection_pool_kw.get("retries")
```
thetorpedodog, line 7: This seems like it should be

```
kwargs.setdefault("retries", ...)
```

so that if the caller were to pass their own `retries`, it would not get clobbered.

dimos-tdb: The caller should not be passing `retries` here though. `retries` are passed in the PoolManager constructor, but not propagated here due to https://github.com/urllib3/urllib3/issues/2475

thetorpedodog: That's just a default; a caller can override the behavior for a given request by passing it here (which is in fact exactly what we are doing here, just that the "override" we are passing to the PoolManager is just what the default *should* have been).

tiledb/cloud/pool_manager_wrapper.py
dimos-tdb: changed

tiledb/cloud/client.py
dimos-tdb: What's the benefit of the suggested change? Can you please elaborate?

tiledb/cloud/pool_manager_wrapper.py
dimos-tdb: I understand how it works. What I mean though is that we should not allow the caller to override this but always use the retries passed to the PoolManager constructor for consistency. Once this is fixed upstream, we can remove this functionality completely.

thetorpedodog: My concern here is that this would replace one unwanted behavior (not using the configured retries) with a different unwanted behavior (discarding call-specific retries). When urllib3 itself is fixed, the behavior will be like the `setdefault` behavior, since the default will be used but the caller will be able to override it. Admittedly, this is not a huge concern right now since we don't use it, but I think the behavior we implement here should be as close to The Right Thing as possible: I don't want us to either depend upon the "forbid overrides" behavior (which is unlike the rest of urllib3) or run into a problem where we *do* want to specify a custom retry policy for a specific request.

tiledb/cloud/client.py
thetorpedodog: We don't use it quite everywhere here, but the idea is to make it easy to tell exactly where any symbol is coming from at its usage site. I picked it up from [the Google style guide](https://google.github.io/styleguide/pyguide.html#22-imports), where the rationale is explained:

> The namespace management convention is simple. The source of each identifier is indicated in a consistent way; `x.Obj` says that object `Obj` is defined in module `x`.

antalakas (approved):
Approving since this is a working solution, improving the code base in the direction given by the specification described in the ticket. Let's address concerns in a follow up PR

PR #204, opened 2021-11-22
Support file create and export

antalakas:
Usage:

```
from tiledb import cloud

created = cloud.create_file("demo", "s3://bucket/bubble.png", "s3://bucket/bubble_tdb")
print(created)

exported = cloud.export_file("demo", created.file_uuid, "s3://bucket/bubble.png")
print(exported)
```


antalakas:
Ref: @snagles

tiledb/cloud/file.py
```diff
@@ -0,0 +1,56 @@
+from tiledb.cloud import client
+from tiledb.cloud import tiledb_cloud_error
+from tiledb.cloud.rest_api import ApiException as GenApiException
+from tiledb.cloud.rest_api import models
+
+
+def create_file(namespace, input_uri, output_uri, async_req=False):
```
thetorpedodog, line 7: Can you add type annotations here?

thetorpedodog (commented):
Sorry for missing this yesterday!

Shelnutt2 (approved):

tiledb/cloud/file.py
antalakas: Sure

tiledb/cloud/file.py
```diff
@@ -0,0 +1,66 @@
...
+from tiledb.cloud.rest_api import ApiException as GenApiException
+from tiledb.cloud.rest_api import models
+
+
+def create_file(
+    namespace: str,
+    input_uri: str,
+    output_uri: str,
+    async_req: bool = False,
+) -> models.FileCreated:
```
thetorpedodog, line 12: I have one minor note on this: Because of the way the `async_req` parameter works, this actually means that the function can return either a `models.FileCreated` *or* a `futures.Future[models.FileCreated]`. (This is one of the things I like least about the generated API—it returns different *types* based upon the value of one of the params, which makes automatic checking difficult.) So technically this should be a `Union[models.FileCreated, futures.Future[models.FileCreated]]`, but other than that this is exactly what it should be.

One other minor gotcha that you might run into: in Python 3.7, futures.Future is not a generic class, so you have to specify it in quotes, so in the end it would look like:

```python
# new imports
from concurrent import futures
from typing import Union

def create_file(
    namespace: str,
    input_uri: str,
    output_uri: str,
    async_req: bool = False,
) -> Union[models.FileCreated, "futures.Future[models.FileCreated]"]:
    ...
```

PR #205, opened 2021-11-30
Support name to register created file with

Shelnutt2:
Support name to register created file with.


antalakas (approved):

tiledb/cloud/file.py
```diff
@@ -8,11 +8,13 @@ def create_file(
     namespace: str,
     input_uri: str,
     output_uri: str,
+    name: str = None,
```
thetorpedodog, line 11:
```suggestion
    name: Optional[str] = None,
```

thetorpedodog (approved):

tiledb/cloud/file.py
Shelnutt2: Thanks!

PR #207, opened 2021-11-30
Add support for local file export

Shelnutt2:
The export_file command now support running the export through TileDB cloud directly to S3 and exporting locally.

### DIFF
--- a/tiledb/cloud/file.py
+++ b/tiledb/cloud/file.py
@@ -1,11 +1,11 @@
-from typing import Optional
+from typing import Tuple, Union, Optional
 
+import tiledb
+from tiledb.cloud import array
 from tiledb.cloud import client
 from tiledb.cloud import tiledb_cloud_error
 from tiledb.cloud.rest_api import ApiException as GenApiException
 from tiledb.cloud.rest_api import models
-from tiledb.cloud import array
-import tiledb
 
 
 def create_file(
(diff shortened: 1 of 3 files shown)


tiledb/cloud/file.py
```diff
@@ -36,6 +39,41 @@ def create_file(
...
+        with tiledb.open(uri, ctx=ctx, timestamp=timestamp) as A:
+            file_size = A.meta["file_size"]
+            # Row major partial export of single attribute
+            iterable = A.query(
+                attrs=["contents"], return_incomplete=True, order="C"
+            ).multi_index[0:file_size]
+
+            fh = vfs.open(output_uri, "wb")
+            for part in iterable:
+                vfs.write(fh, part["contents"])
```
antalakas, line 69: This will write file contents but what about the extension? Don't we need to take into account the mime type for this?

Shelnutt2: For simplicity I was just forcing the user to provide the full local path

antalakas (approved):

tiledb/cloud/file.py
```diff
@@ -36,6 +39,41 @@ def create_file(
         raise tiledb_cloud_error.check_exc(exc) from None
 
 
+def export_file_local(
+    uri: str,
+    output_uri: str,
+    timestamp: Optional[tuple] = None,
```
thetorpedodog, line 45: Based on the docs it looks like this should be:

```suggestion
    timestamp: Union[Tuple[int, int], int, None] = None,
```

(also add `Union` and `Tuple` to the things you’re importing from `typing` above)

tiledb/cloud/file.py
```diff
@@ -53,6 +91,19 @@ def export_file(
 
         api_instance = client.client.file_api
 
+        local = True
+        if (
+            uri.startswith("tiledb://")
```
thetorpedodog, line 96: Would it be better to check if there is no `://` at all or it starts with `file://`?

```python
local = uri.startswith('file://`) or '://' not in uri
```

Shelnutt2: @thetorpedodog we really need to expose a VFS tiledb setting for this since we have this inside the core code. For now I was just looking for a quick fix. I'd be fine with your suggested adjustment! Do you mind pushing the change to this branch?

thetorpedodog: Done and also rebased to fix the conflict.

PR #214, opened 2021-12-08
Cache HTTP redirects

dimos-tdb:


tiledb/cloud/pool_manager_wrapper.py
```diff
@@ -1,17 +1,41 @@
...
 
     It can be used to replace tiledb.cloud.rest_api.rest.RESTClientObject.pool_manager
-    at runtime to work around https://github.com/urllib3/urllib3/issues/2475
+    at runtime in order to cache HTTP redirects and work around
+    https://github.com/urllib3/urllib3/issues/2475
     """
 
     def __init__(self, pool_manager):
         self._pool = pool_manager
+        self._redirect_cache = {}
```
thetorpedodog, line 16: with an added `from typing import Dict` above, add:

```suggestion
        self._redirect_cache: Dict[str, str] = {}
```

Also, can you add a comment explaining exactly what the semantics of the keys and values are? It looks like it's `{url without query string: domain}`.

tiledb/cloud/pool_manager_wrapper.py
```diff
@@ -1,17 +1,41 @@
...
+
+        cached_netloc = self._redirect_cache.get(cacheable_url)
+        if cached_netloc:
+            url = parsed_url._replace(netloc=cached_netloc).geturl()
+
         resp = self._pool.request(method, url, **kwargs)
 
+        for retry in reversed(resp.retries.history):
+            if retry.redirect_location:
+                self._redirect_cache[cacheable_url] = urllib.parse.urlparse(
```
thetorpedodog, line 35: This will probably be easier to read if it's broken into two parts:

```python
dest_domain = urllib.parse.urlparse(retry.redirect_location).netloc
self._redirect_cache[cacheable_url] = dest_domain
```

thetorpedodog (commented):
A couple small requests but otherwise this looks good.

tiledb/cloud/pool_manager_wrapper.py
dimos-tdb: fixed

tiledb/cloud/pool_manager_wrapper.py
dimos-tdb: done

thetorpedodog (approved):

Shelnutt2 (approved):
Looks great!

PR #219, opened 2022-01-04
Support TILEDB_REST_IGNORE_SSL_VALIDATION env var

Shelnutt2:
Support `TILEDB_REST_IGNORE_SSL_VALIDATION` env variable for toggling verify_ssl settings. This is the setting the core library support for disabling ssl validation for curl/REST requests.

The logic here is reverse so we negate the env variable if set.

### DIFF
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -254,24 +254,17 @@ class BasicTests(unittest.TestCase):
             tasks.fetch_results(uuid.uuid4())
 
     def test_timeout(self):
-        def test():
+        def test(_):
             import time
 
             time.sleep(10)
 
         with self.assertRaises(tiledb_cloud_error.TileDBCloudError):
 
-            def test():
-                import time
-
-                time.sleep(10)
-
             with tiledb.open(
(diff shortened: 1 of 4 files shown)


dimos-tdb (approved):

tests/test_basic.py
```diff
@@ -261,17 +261,15 @@ def test():
 
         with self.assertRaises(tiledb_cloud_error.TileDBCloudError):
 
-            def test():
+            def test(data):
```
thetorpedodog, line 264: Since this function is a duplicate of the one above, I would delete this one and add the parameter to the one above (you can call it `_`, since it’s unused).

tiledb/cloud/config.py
```diff
@@ -12,6 +12,13 @@
 default_config_file = Path.joinpath(Path.home(), ".tiledb", "cloud.json")
 
 
+def parse_bool(s: str) -> bool:
+    if s is None:
```
thetorpedodog, line 16: This should never happen per the type annotation. If you expect to get `None` values, then `s` should be an `Optional[str]`.

thetorpedodog (commented):
Few comments, no major problems.

thetorpedodog (approved):

PR #221, opened 2022-01-20
Construct TaskGraphLog based on the structure of a DAG.

thetorpedodog:
This adds the code to create the TaskGraphLog structure based on a DAG.
Includes topological sort of node metadatas and tests of results.

---

This will be a few steps; this is the first, also broken out for easier digestion.

### DIFF
--- a/tests/test_dag.py
+++ b/tests/test_dag.py
@@ -1,6 +1,7 @@
 import base64
 import collections
 import collections.abc as cabc
+import itertools
 import os
 import pickle
 import threading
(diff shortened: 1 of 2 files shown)


tests/test_dag.py
```diff
@@ -167,6 +194,51 @@ def test_multi_dependencies_dag(self):
         self.assertEqual(node_5.result(), 8)
         self.assertEqual(node_6.result(), 24)
 
+        # The order of this assertion is somewhat fragile. It is deterministic
```
johnkerl, line 197: great comment! will be super-helpful if/when it fails in precisely that way & someone with less context comes along to fix it. <3

johnkerl (approved):
LGTM ...

PR #228, opened 2022-01-26
Actually do the one-time DAG setup only *one time*.

thetorpedodog:
Oops.


ihnorton (approved):
How long does this server side object last if eg the client crashes?

thetorpedodog:
The server-side object is just a log entry in the task graph logs table, and (intentionally) lasts indefinitely. It’s the equivalent of an ArrayTask, but for keeping track of the progress of an entire graph.

PR #232, opened 2022-01-28
Reinit API clients after setting thread count.

thetorpedodog:
Also separates external calls from internal setup to avoid reiniting
API clients unnecessarily.


Shelnutt2 (approved):
great!

PR #233, opened 2022-01-28
Enable keepalive pings on sockets to prevent connection drops.

thetorpedodog:
For particularly long-running UDFs or SQL queries, the server may go
minutes or more without responding with any data back to the client.
This means that NATs or load balancers may drop the connection,
resulting in the loss of all the UDF's progress. This change enables
1-minutely TCP keepalive pings on sockets, so idle connections are
kept fresh and do not have an opportunity to go stale.

Short-lived connections and long-lived connections that do not go idle
due to ongoing transfers are unaffected.

---

I have tested this on my Linux VM and verified that it works there using tcpdump. The below screenshot is from a test where I set a 5-second keepalive time and executed `udf.exec(time.sleep, 30)`:

Wireshark screenshot showing keepalive packets in TCP dump


thetorpedodog:
I don’t have a Mac to test this on, but these steps *should* verify that keepalive also works there:

 1. In your checkout, replace `_KEEPALIVE_SECONDS = 60` with `5` or some other value.
 2. In a venv:
    ```plain
    $ pip install --update pip wheel && pip install path/to/TileDB-Cloud
    $ ipython
    In[1]: # login to tiledb cloud if you haven't already
    In[2]: from tiledb.cloud import udf
    In[3]: import time
    ```
 3. In a separate terminal:
    ```plain
    root@your-mac# tcpdump -w /some/output/file.pcap port 443 and host api.cloud.tiledb.com
    ```
    (You could also do probably do this with the Wireshark GUI.)
 4. Back in the venv terminal:
    ```plain
    In[4]: udf.exec(time.sleep, 30)
    ```
    and wait for it to return. `exit()` the `ipython` session.
 5. Interrupt (still ctrl-C on Mac?) the `tcpdump`. Open the pcap file you just created and you should see keepalive packets.

ihnorton:
Do we still need this?

JohnMoutafis:
This PR is not relevant anymore.

PR #234, opened 2022-01-31
Add warnings on task graph log submit failure and other TODOs.

thetorpedodog:
- Raises a warning when task graph log submission fails. This does not
  block execution.
- Remove workarounds from when server did not honor `store_results`
  for SQL queries.


johnkerl (approved):
LGTM

🚢

PR #237, opened 2022-02-24
Fix error with zero-length Arrow response bodies.

thetorpedodog:
For situations where a UDF was successful but for whatever reason
the result returned was a zero-length body, we would mistakenly assume
that the body was not provided, rather than recognizing it as a valid
but zero-length body. We now distinguish between no body and a
zero-length body, and properly handle zero-length Arrow bodies.

We had previously tested against empty Arrow bodies with SQL queries.
SQL queries include a schema in their results, and thus have nonzero
response length. However, empty UDF responses which are written in Arrow
format would not have any write calls at all, meaning the response was
completely empty.

Also adds a test for this situation.

---

Thanks to @ckrilow for finding this bug!

### DIFF
--- a/tiledb/cloud/_results/results.py
+++ b/tiledb/cloud/_results/results.py
@@ -101,7 +101,7 @@ class RemoteResult(Result[_T], Generic[_T]):
         del attribute, value  # unused
         if self.results_stored and not self.task_id:
             raise ValueError("task_id must be set for stored results")
-        if not self.results_stored and not self._body:
+        if not self.results_stored and self._body is None:
             raise ValueError(
                 "no way to access Node results;"
                 " they must be either stored or downloaded"
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -4,6 +4,7 @@ import uuid
 
 import numpy
 import numpy as np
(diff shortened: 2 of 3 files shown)


tiledb/cloud/_results/results.py
```diff
@@ -101,7 +101,7 @@ def _validate(self, attribute, value):
         del attribute, value  # unused
         if self.results_stored and not self.task_id:
             raise ValueError("task_id must be set for stored results")
-        if not self.results_stored and not self._body:
+        if not self.results_stored and self._body is None:
```
johnkerl, line 104: Interesting this is `None` vs `""` -- I do believe you've tested this correctly. Is it worth adding

```
if not self.results_stored and (self._body is None or self._body == ""):
```

?

(just a thought)

johnkerl (approved):
LGTM, with comment

tiledb/cloud/_results/results.py
thetorpedodog: We specifically don’t; this needs to test that the `_body` is not provided (results were not downloaded) versus empty (results were downloaded and were zero-length).

PR #239, opened 2022-03-16
Add very basic value decoder for task graphs.

thetorpedodog:
This is a base implementation of a JSON value decoder to be extended by other implementations in a location-specific manner.

It's the start of improving compatibility and extensibility of parameter-passing across versions and languages.

### DIFF
--- a/tests/taskgraphs/test_codec.py
+++ b/tests/taskgraphs/test_codec.py
@@ -1,0 +1,129 @@
+import datetime
+import json
+import unittest
+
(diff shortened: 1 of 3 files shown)


johnkerl (commented):
This looks syntactically correct but I'd like some narrative in the description field about why this is here and where it's going

thetorpedodog:
This is the start of improving compatibility and extensibility of parameter passing across versions and languages.

johnkerl (approved):
🚢

PR #241, opened 2022-03-21
Update interchange format for less verbosity and nesting.

thetorpedodog:
The previously-devised interchange format required lots of nested
objects, which became particularly acute for nested values.
As a simple example:

    {
      "__tiledb_sentinel__": {
        "some_kind": {"inner": "data"},
      }
    }

This update simplifies it by removing the inner nesting and shortening
the replacement key:

    {
      "__tdbudf__": "some_kind",
      "inner": "data"
    }

The old format has not actually been used by any other code, so changing
will *not* break anything. The new format can encode all data that the
old could.

### DIFF
--- a/tests/taskgraphs/test_codec.py
+++ b/tests/taskgraphs/test_codec.py
@@ -17,69 +17,54 @@ class TestUnescaper(unittest.TestCase):
                 "nested pickle",
                 """
                     {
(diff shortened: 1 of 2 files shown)


johnkerl (approved):
LGTM!

PR #242, opened 2022-03-22
Allow raw installation outside of the context of PyPI.

thetorpedodog:
If the package is installed outside of PyPI or an active checkout
(e.g. via `pip install [some-path]`), a `version.py` file might not be
created. Use a local version number in that case.


Shelnutt2:
@thetorpedodog I've used `python setup.py develop --user` as my go-to for a while.  Are you trying to get it working from the top level repo folder without any install?

antalakas (approved):

thetorpedodog:
Figured out what spurred me to do this. It would cause an error if you pip-installed directly from git:

```
pip install --upgrade git+https://github.com/TileDB-Inc/TileDB-Cloud-Py.git@some-version-hash
```

PR #244, opened 2022-03-23
Be less paitent when waiting for a DAG to complete.

thetorpedodog:
Because we're only waiting on ourselves, it's completely reasonable to
poll for completion fairly frequently. It's extremely cheap and it's
not interfering with anybody else.

The ideal solution would be to use an Event that was set, but getting
that to work would be much more invasive.


thetorpedodog:
speling

Shelnutt2 (approved):

Shelnutt2:
@thetorpedodog should we merge this to include in the next release?

PR #249, opened 2022-04-01
Use TILEDB_REST_TOKEN to store TileDB Cloud login info.

thetorpedodog:
Removes the need for the special `TILEDB_CLOUD_HELPER_VAR` environment
variable and makes it possible to run unit tests that do not use
the network without them failing.


Shelnutt2 (approved):
Been meaning to do this for a long time! :tada:

PR #253, opened 2022-04-05
Create a simple dependency-graph builder.

thetorpedodog:
This creates `depgraph`, which builds dependency graphs and ensures
they are valid. Its limited API is intended to fit well with the limited
case of building DAGs one node at a time with known dependencies, while
allowing manual insertion of dependencies (provided they are legal!).

---

[sc-14031]

### DIFF
--- a/tests/taskgraphs/test_depgraph.py
+++ b/tests/taskgraphs/test_depgraph.py
@@ -1,0 +1,29 @@
+import unittest
+
+from tiledb.cloud.taskgraphs import depgraph
+
+
+class TestDepGraph(unittest.TestCase):
+    def test_it(self):
+        g = depgraph.DepGraph[str]()
+        g.add_new_node("a", "")
+        g.add_new_node("b", "a")
+        g.add_new_node("c", "b")
+        g.add_new_node("d", "a")
+        g.add_new_node("e", "abcd")
+        self.assertEqual(tuple("abcde"), g.topo_sorted)
+        g.add_edge(child="b", parent="d")
+        self.assertEqual(tuple("adbce"), g.topo_sorted)
+        with self.assertRaises(depgraph.CyclicGraphError):
+            g.add_edge(child="a", parent="e")
+        with self.assertRaises(depgraph.CyclicGraphError):
+            g.add_edge(child="b", parent="b")
+        with self.assertRaises(KeyError):
+            g.add_new_node("b", "")
+        with self.assertRaises(KeyError):
+            g.add_new_node("f", "gh")
+        with self.assertRaises(KeyError):
+            g.add_edge(child="a", parent="q")
+
+        g.remove("c")
+        self.assertEqual(tuple("adbe"), g.topo_sorted)
--- a/tests/taskgraphs/test_depgraph.py
+++ b/tests/taskgraphs/test_depgraph.py
@@ -6,24 +6,87 @@ from tiledb.cloud.taskgraphs import depgraph
 class TestDepGraph(unittest.TestCase):
     def test_it(self):
         g = depgraph.DepGraph[str]()
-        g.add_new_node("a", "")
-        g.add_new_node("b", "a")
-        g.add_new_node("c", "b")
-        g.add_new_node("d", "a")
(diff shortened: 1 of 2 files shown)


tests/taskgraphs/test_depgraph.py
```diff
@@ -0,0 +1,29 @@
...
+
+class TestDepGraph(unittest.TestCase):
+    def test_it(self):
+        g = depgraph.DepGraph[str]()
+        g.add_new_node("a", "")
+        g.add_new_node("b", "a")
+        g.add_new_node("c", "b")
+        g.add_new_node("d", "a")
+        g.add_new_node("e", "abcd")
+        self.assertEqual(tuple("abcde"), g.topo_sorted)
```
johnkerl, line 14: If I read this right:

* node names are required to be single-letter?
* node-lists are multi-letter strings?

that is bound to be limiting

if on the other hand it's the case that some `for e in nodelist:` somewhere happens to work OK whether `nodelist` is like `"abc"` for nodes `"a"`, `"b"`, and `"c"`, _or_ if nodelist is like `["a", "b", "c"]` -- because both string and list-of-string are `Iterable` -- then i think this is overly clever and confusing. if this latter case is true let's please spell out `["a", "b", "c"]` -- even if it takes a few more keystrokes right now. It'll be much less confusing forever after.

tiledb/cloud/taskgraphs/depgraph.py
```diff
@@ -0,0 +1,111 @@
...
+        )
+        while q:
+            parent = q.popleft()
+            del in_degrees[parent]
+            for child in self._parent_to_children[parent]:
+                in_degrees[child] -= 1
+                if in_degrees[child] == 0:
+                    q.append(child)
+            output.append(parent)
+        if in_degrees:
```
johnkerl, line 94: in general i don't like implicit truthiness. it's surely 'correct' or 'idiomatic' in some python-fluent perspective. but most of us are polyglots these days. it's important that we code in ways that are clearly correct to the average reader (even though you're a far-above-average writer when it comes to python).

personally i'd prefer `if len(in_degrees) == 0:` but i know there are always concerns about computing sizes of things when is-empty is all that's wanted. i'd be ok with a more explicit `if bool(in_degrees):` here.

p.s. did you know `if (0)` is true in ruby? my favorite example of why implicit truthiness is a Bad Thing. (and don't even get me started on what javascript lets people get away with ...)

tests/taskgraphs/test_depgraph.py
```diff
@@ -0,0 +1,29 @@
+import unittest
+
+from tiledb.cloud.taskgraphs import depgraph
+
+
+class TestDepGraph(unittest.TestCase):
```
johnkerl, line 6: please add test cases for:

* empty graph
* single-node graph

things like this can surface otherwise-non-obvious corner-case issues

even if your code correctly handles these cases today, in your python-expert hands, it might not continue to do so in the future. let's guard-rail it a bit here.

johnkerl (changes requested):

tests/taskgraphs/test_depgraph.py
thetorpedodog: Added tests for these cases.

tiledb/cloud/taskgraphs/depgraph.py
thetorpedodog: I'd prefer to keep using implicit `__nonzero__` here—it’s not only widely used in Python development but is regarded as a [good practice in the Google style guide](https://google.github.io/styleguide/pyguide.html#214-truefalse-evaluations) (if not elsewhere as well).

tests/taskgraphs/test_depgraph.py
thetorpedodog: This isn’t a requirement; it was just me using individual letters as names for the test, and "abcde" (for example) as a sequence of strings. I've expanded these to easier-to-understand names.

johnkerl (approved):
❤️

thetorpedodog:
Thanks for the review!

PR #254, opened 2022-04-05
Create basic types and codec for task graph registration.

thetorpedodog:
This change introduces the basic types that will be used throughout the
task graph registration process, and some basic codecs to transform
Python data back and forth into JSON. The types defined here support
the major encoding needs of a task graph:

- `RegisteredArg`

  "At-rest" encoding, where the nodes are serialized into a
  `RegisteredTaskGraph`. In this case, parameters are stored with
  references to previous nodes in the form of:

      {"__tdbudf__": "node_output", "client_node_id": "..."}

  This allows an executor to reconstruct the graph and substitute in
  stored values when necessary.

- `CallArgStoredParams`

  "Stored parameter" encoding, where nodes are referred to by the
  server-provided execution ID.

      {"__tdbudf__": "stored_param", "task_id": "..."}

  The `task_id` here is the server-generated task ID from performing
  the query. Parameters may also include arrays:

      {"__tdbudf__": "udf_array_details", "udf_array_details": {...}}

  When a UDF is executed, these are sent by the client to the server.

- `CallArg`

  An alternate encoding for when stored parameters are not available.
  Data encoded in this format does not include any `stored_param`
  objects (but still may include `udf_array_details`, since those are
  guaranteed to be loaded by the server). Beyond the
  `udf_array_details`, an `Unescaper` should be able to deserialize
  `CallArg` data on its own.

This builds upon the format first defined in
2c3312e70b76efc21cdd339b03f7fa4b4766cd09.

---

[ch-14031]

### DIFF
--- a/tests/taskgraphs/test_codec.py
+++ b/tests/taskgraphs/test_codec.py
@@ -2,6 +2,9 @@ import datetime
 import json
 import unittest
 
+import pyarrow
+import urllib3
+
 from tiledb.cloud.taskgraphs import _codec
 
 
@@ -127,17 +130,53 @@ class BinaryResultTest(unittest.TestCase):
     def test_binary_result_of(self):
         cases = (
             (b"possession", _codec.BinaryResult("bytes", b"possession")),
-            (
-                frozenset(),
-                _codec.BinaryResult(
-                    "python_pickle",
-                    b"\x80\x04\x95\x04\x00\x00\x00\x00\x00\x00\x00(\x91\x94.",
-                ),
-            ),
+            (set(), _codec.BinaryResult("python_pickle", b"\x80\x04\x8f\x94.")),
+            (pyarrow.Table.from_pydict({}), _codec.BinaryResult("arrow", b"")),
         )
         for inval, expected in cases:
             with self.subTest(inval):
                 self.assertEqual(expected, _codec.BinaryResult.of(inval))
+
+    def test_from_response(self):
+        cases = [
+            dict(
+                mime="application/octet-stream",
+                data=b"raw bytes",
+                want_format="bytes",
+                want_output=b"raw bytes",
+            ),
+            dict(
+                mime="application/json",
+                data=b'{"here": "there"}',
+                want_format="json",
+                want_output={"here": "there"},
+            ),
+            dict(
+                mime="application/vnd.tiledb.python-pickle",
+                data=b"\x80\x04\x95\x04\x00\x00\x00\x00\x00\x00\x00(\x91\x94.",
+                want_format="python_pickle",
+                want_output=frozenset(),
+            ),
+            dict(
+                mime="unknown/mime",
+                data=b"bogus data",
+                want_format="mime:unknown/mime",
+            ),
+        ]
+        for case in cases:
+            with self.subTest(case["mime"]):
+                resp = urllib3.HTTPResponse(
+                    body=case["data"],
+                    headers={"Content-Type": case["mime"]},
+                )
+                actual = _codec.BinaryResult.from_response(resp)
+                self.assertEqual(case["want_format"], actual.format)
+                try:
+                    want_out = case["want_output"]
+                except KeyError:
+                    pass
+                else:
+                    self.assertEqual(want_out, actual.decode())
 
 
 class JSONableTest(unittest.TestCase):
--- a/tests/taskgraphs/test_codec.py
+++ b/tests/taskgraphs/test_codec.py
@@ -5,8 +5,8 @@ import unittest
 from tiledb.cloud.taskgraphs import _codec
 
 
-class TestUnescaper(unittest.TestCase):
-    def test_unescaping(self):
+class EscapingTest(unittest.TestCase):
+    def test_bidirectional(self):
         cases = (
             (
                 "simple",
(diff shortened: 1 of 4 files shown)


thetorpedodog:
NOTE TO SELF: add tests for `is_jsonable_shallow`

tests/taskgraphs/test_codec.py
```diff
@@ -99,9 +99,74 @@ def test_unescaping(self):
...
+
+class BinaryResultTest(unittest.TestCase):
+    def test_binary_result_of(self):
+        cases = (
+            (b"possession", _codec.BinaryResult("bytes", b"possession")),
+            (
+                frozenset(),
+                _codec.BinaryResult(
+                    "python_pickle",
+                    b"\x80\x04\x95\x04\x00\x00\x00\x00\x00\x00\x00(\x91\x94.",
```
johnkerl, line 134: can you add a comment how this hex string was created, so a less-Python-expert-than-Paul person could reproduce this in the future, e.g. if they ever try to add similar test cases?

johnkerl: i am guessing:

```
>>> import cloudpickle
>>> cloudpickle.dumps(b"possession")
b'\x80\x04\x95\x0e\x00\x00\x00\x00\x00\x00\x00C\npossession\x94.'
```

but that's not quite right as I'm not reproducing your line 134

also, please let a comment here regarding what Python version was used here

tiledb/cloud/taskgraphs/_codec.py
```diff
@@ -63,6 +84,44 @@ def _replace_sentinel(
         raise ValueError(f"Unknown sentinel type {kind!r}")
 
 
+class Escaper(visitor.ReplacingVisitor):
+    """Turns arbitrary Python values into TileDB JSON.
+
+    This escapes arbitrary native values so that they can be JSON-serialized.
+    It should only be used with ``NativeValue``s—that is, values that are
```
johnkerl, line 91: space around "—" please

tiledb/cloud/taskgraphs/_codec.py
```diff
@@ -71,3 +130,102 @@ def _replace_sentinel(
...
+        full_mime = resp.getheader("Content-type", "application/octet-stream")
+        mime, _, _ = full_mime.partition(";")
+        mime = mime.strip()
+        format = _MIME_TO_FORMAT.get(mime, "mime:" + mime)
+        data = resp.data
+        return cls(format, data)
+
+    def decode(self) -> types.NativeValue:
+        """Decodes this result into native Python data."""
+        # This is not lock-protected because we're ok with decoding twice.
```
johnkerl, line 161: is there a hidden performance problem here? if it someday became the case that we were decoding twice (or more), would we know?

tiledb/cloud/taskgraphs/types.py
```diff
@@ -0,0 +1,58 @@
...
+to decode them contained within their data.
+"""
+
+
+@attrs.define(frozen=True, slots=True)
+class Arguments:
+    """The arguments and keyword arguments that can be sent to a function.
+
+    You usually shouldn't call the constructor directly; instead use
+    :meth:``of``.
```
johnkerl, line 46: maybe show, or point/link to, an example invocation of `of`?

tiledb/cloud/taskgraphs/types.py
```diff
@@ -0,0 +1,58 @@
+"""User-facing types used in task graphs."""
```
johnkerl, line 1: used in _registered_ task graphs -- ?

tiledb/cloud/taskgraphs/_codec.py
```diff
@@ -71,3 +130,102 @@ def _replace_sentinel(
...
+class BinaryResult(TDBJSONEncodable):
+    """Container for a binary-encoded value, decoded on-demand.
+
+    This is used to store results obtained from the server, such that it is not
+    necessary to decode them between stages, and they only are decoded upon
+    request.
+    """
+
+    format: str
+    """The format of the data."""
```
johnkerl, line 145: How about:

The TileDB format of the data (see `_MIME_TO_FORMAT`)

tiledb/cloud/taskgraphs/_codec.py
```diff
@@ -1,33 +1,54 @@
...
+from tiledb.cloud.taskgraphs import types
 
 _T = TypeVar("_T")
 TOrDict = Union[_T, Dict[str, Any]]
-_SENTINEL_KEY = "__tdbudf__"
-_ESCAPE_CODE = "__escape__"
+
+_PICKLE_PROTOCOL = 4
+_ARROW_VERSION = pyarrow.MetadataVersion.V5
+SENTINEL_KEY = "__tdbudf__"
```
johnkerl, line 20: this would be a good place for a link to a spec doc / spec.md / whatever we have as a "start here" place to read about the format

johnkerl (changes requested):

tiledb/cloud/taskgraphs/_codec.py
thetorpedodog: Added detail.

johnkerl (approved):
LGTM, pending only
https://github.com/TileDB-Inc/TileDB-Cloud-Py/pull/254/files#r852228566

PR #258, opened 2022-04-08
Add `copy`, `roots`, and `leaves` methods to DepGraph.

thetorpedodog:
Making a copy of a DepGraph means the copy can be mutated, so we can
use it to track execution progress while actively running a graph.
As execution continues, we can remove completed nodes from the graph
and use the `roots` to track nodes which are ready to run.


thetorpedodog:
also part of [sc-14031]

johnkerl (approved):

PR #262, opened 2022-04-14
Move `testonly` package from `tiledb.cloud` into `tests`.

thetorpedodog:
I'm not sure if I just missed doing this when I was first creating it
or doing this didn't work right with pytest at the time, but this does
work now, and it means we no longer pollute `tiledb.cloud` with
code that should not be used by end users.


Shelnutt2 (approved):

thetorpedodog:
Works locally on my machine but not in the test environment :(

PR #263, opened 2022-04-18
Add utils.PYTHON_VERSION and test UUID generator.

thetorpedodog:
A small bit of prework for upcoming changes:

- Centralize construction of `X.Y.Z` Python version string.
- Create `testonly.sequential_uuids` generator function for mocks
  so that tests can have consistent UUIDs.


johnkerl (approved):
🚢

PR #264, opened 2022-04-18
Create the thing that builds task graphs.

thetorpedodog:
At last! We're getting to the meat! Things are finally happening.

This creates the initial version of the task graph builder. It consists
of only the part which builds a graph. Execution will come later.

The API is based on the existing `tiledb.cloud.dag` API, but with
better separation of concerns, and significantly streamlined. Some
features will be added in a follow-up, but the existing API is enough
to construct all but the most specialized task graphs.

Each function on the `TaskGraphBuilder` returns a specific type of
`Node` which, like the existing API, represents a step to be executed.
These `Node`s are inert and only represent the actions that will be
later executed by an Executor and have no state of their own.

`Node`s are capable of accepting other nodes as input not only in the
traditional UDF argument slots (i.e. those passed into user code),
but also in selected parameters of other functions where users may wish
to pass in parameters: array queries, SQL parameters, and a few others.

---

A few things that are worth pointing out:

- Naming of the functions that you call on a graph to add a new node. I went with things that look like `grf.udf(...)` when called here, but I am wondering whether it might be better to include an `add_` or maybe a `_node` in the name.

- Using nodes as non-argument parameters. I've included it because it's something we currently support (if accidentally), and I can see wanting parameterizable configuration of e.g. query ranges. But is this something we explicitly want? My impression is "yes" but I want to make sure…

- The way we accept ranges. The current implementation only accepts ranges in the form that is directly used by the backend. The difficulty with making it work the same way as the current `parse_ranges` implementation is that:

  - `parse_ranges` distinguishes between lists and tuples, which we can't do after JSON serialization.
  - If users want to pass in node outputs into ranges, we can't necessarily parse them that way.
  - Whatever interchange format we use should be implementation-independent, so that we can make an executor in another language (e.g. server-side) in the future.

  These are the reasons that I'm currently only handling the `[[start, end, start2, end2], [startB, endB, ...], ...]` format of range specification, because designing an API for that is an even larger task.

[sc-14031]

### DIFF
--- a/tests/taskgraphs/test_builder.py
+++ b/tests/taskgraphs/test_builder.py
@@ -3,7 +3,10 @@ import uuid
 from unittest import mock
 
 from tiledb.cloud import testonly
+from tiledb.cloud import utils
 from tiledb.cloud.taskgraphs import builder
+from tiledb.cloud.taskgraphs import depgraph
+from tiledb.cloud.taskgraphs import types
 
 
 class TestEscaper(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


thetorpedodog:
OK, we’re back. This should be much easier to deal with, since it’s step-by-step now. It still makes sense to submit as one unit, because the entire thing needs to be there to get the full functionality of task graphs, and many of the API decisions should be taken together. However, it should be easier to review since each step can be taken in one at a time.

antalakas (approved):
Naming of the functions: I am good with not using `add_` or `_node`
> Using nodes as non-argument parameters.
Agree on keep using this

PR #266, opened 2022-04-22
Allow nodes from old DAGs to be used in new DAGs.

thetorpedodog:
Previously, if you ran a DAG either via DAG or Delayed and passed in a
Node from a previous graph, it would either not work at all or give you
a bizarre error message (like "graph does not have any root nodes").
With this change, you can pass Nodes from a completed DAG to a new DAG
and they will be used as a caller would expect as parameters.

---

This is an issue that Chad ran into earlier that caused him problems, and also fixes a longstanding issue [sc-11457] where if you executed a `Delayed` and attempted to pass it into a later `Delayed`, that would fail as well.

### DIFF
--- a/tests/test_dag.py
+++ b/tests/test_dag.py
@@ -18,6 +18,7 @@ import tiledb.cloud
 from tiledb.cloud import client
 from tiledb.cloud import dag
 from tiledb.cloud import tasks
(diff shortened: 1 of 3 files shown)


johnkerl (approved):
LGTM

PR #268, opened 2022-04-26
Use ordered sets in depgraph for consistent tests and RPCs.

thetorpedodog:
This updates the dependency graph to use ordered sets for storing
parent–child relationships. This allows us to have consistent,
predictable (insertion-order–based) ordering for both RPCs that we send
and tests that we perform.

The set classes themselves live in the `ordered` package and use the
standard Python abstract base classes, which provide most of the
mechanics of being a Set given only a few method implementations.

### DIFF
--- a/tests/common/test_ordered.py
+++ b/tests/common/test_ordered.py
@@ -1,0 +1,77 @@
+import itertools
+import unittest
+
+from tiledb.cloud._common import ordered
+
+
+class TestSets(unittest.TestCase):
+    def test_frozen(self):
+        s = ordered.FrozenSet(
+            ("one", "two", "three", "two", "three", "four", "a billion")
+        )
+        self.assertEqual(5, len(s))
+        self.assertEqual(("one", "two", "three", "four", "a billion"), tuple(s))
+        self.assertIn("one", s)
+        self.assertNotIn("threeve", s)
+
+        other = ordered.FrozenSet(("one", "three", "two", "four", "a billion"))
+        # For compatibility, ordering is not considered when comparing OFSs.
+        self.assertEqual(other, s)
+        self.assertEqual(hash(other), hash(s))
+        # You can also compare an OrderedFrozenSet with a regular frozenset.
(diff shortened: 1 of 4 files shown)


tiledb/cloud/_common/ordered.py
```diff
@@ -0,0 +1,82 @@
...
+_T_co = TypeVar("_T_co", covariant=True)
+_Self = TypeVar("_Self", bound="FrozenSet")
+"""Annotation for returning an object of your own type.
+
+See https://peps.python.org/pep-0673/.
+"""
+
+
+class FrozenSet(AbstractSet[_T_co]):
+    """A frozenset that remembers the insertion order of elements.
```
johnkerl, line 15: ❤️

tiledb/cloud/_common/ordered.py
```diff
@@ -0,0 +1,82 @@
...
+
+    Unlike dicts, Python sets don't remember insertion order. This class
```
johnkerl, line 17: boo! insertion-order-disrespecting sets/dicts are (IMNSHO) a recipe for disaster -- i say that over many years with many languages. one of the things javascript gets right .... :D 

what ends up happening is good solid people like yourself end up inventing what should have been the default all along

tests/common/test_ordered.py
```diff
@@ -0,0 +1,77 @@
...
+
+class TestSets(unittest.TestCase):
+    def test_frozen(self):
+        s = ordered.FrozenSet(
+            ("one", "two", "three", "two", "three", "four", "a billion")
+        )
+        self.assertEqual(5, len(s))
+        self.assertEqual(("one", "two", "three", "four", "a billion"), tuple(s))
+        self.assertIn("one", s)
+        self.assertNotIn("threeve", s)
```
johnkerl, line 15: :D

johnkerl (approved):
❤️ 
🚀 
✌️

PR #270, opened 2022-04-27
Add resource classes to udf exec

anastasop:
https://github.com/TileDB-Inc/TileDB-Cloud-API-Spec/pull/322/

### DIFF
--- a/tiledb/cloud/udf.py
+++ b/tiledb/cloud/udf.py
@@ -33,6 +33,7 @@ def exec_base(
     store_results: bool = False,
     stored_param_uuids: Iterable[uuid.UUID] = (),
     timeout: int = None,
+    resource_class: Optional[str] = None,
     _download_results: bool = True,
     _server_graph_uuid: Optional[uuid.UUID] = None,
     _client_node_uuid: Optional[uuid.UUID] = None,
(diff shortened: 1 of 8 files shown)


tiledb/cloud/udf.py
```diff
@@ -113,6 +116,9 @@ def exec_base(
     if timeout is not None:
         udf_model.timeout = timeout
 
+    if resource_class is not None:
+        udf_model.resource_class = resource_class
```
thetorpedodog, line 120: Because the default here is `None` in both the function and the generated API code, I think you can put this just as

```python
udf_model = models.GenericUDF(
    # ...
    resource_class=resource_class,
)
```

instead of including it here.

thetorpedodog (commented):
One small thing: You're also going to have to do essentially the same change to `exec_base` in `array.py`. Apart from that (and the small note below), this looks great!

thetorpedodog:
Actually, one other thing: Is there a way for the UDF to know what size of environment it's running in? If there's a deterministic way to do this, it would be good to add a test where you call a function with regular vs. large size (or whatever we're calling them) and verify that you get the right output.

tiledb/cloud/udf.py
anastasop: Correct. Yes the constructor does the same check.

anastasop:
> One small thing: You're also going to have to do essentially the same change to `exec_base` in `array.py`. Apart from that (and the small note below), this looks great!

Good catch. I overlooked this

anastasop:
> Actually, one other thing: Is there a way for the UDF to know what size of environment it's running in? If there's a deterministic way to do this, it would be good to add a test where you call a function with regular vs. large size (or whatever we're calling them) and verify that you get the right output.

No, a UDF does not know the environment even on the REST server. If we decide to pass it as an environment variable so that UDF can access it then we add add such a test here

thetorpedodog (approved):

PR #274, opened 2022-05-06
Only cancel descendants of failed DAG nodes

thetorpedodog:
When a DAG fails, it currently immediately aborts every single node in its structure, even if they have no dependency upon the failed node. This is not needed; we only need to cancel those nodes which depend upon the failed node. This change does that.

In order to support doing so, it also restructures the way that Node works to handle its entire lifecycle with a Future, using that as the only source of truth for its status and results.

### DIFF
--- a/tests/test_dag.py
+++ b/tests/test_dag.py
@@ -4,7 +4,6 @@ import collections.abc as cabc
 import itertools
 import pickle
 import threading
-import time
 import unittest
 import uuid
 from concurrent import futures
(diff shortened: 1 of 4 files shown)


antalakas (approved):
Looks good!

Shelnutt2:
I thought this was going to become a behavior option for the user to set? The old behavior is still very much desired for other, different use cases. I.e for read operations where to keep going on failure doesn't bring benefits and just "wastes" compute time. We need to support both behaviors and let the user toggle what to do on failure.

This also then needs to be fully documented with examples in the docs

PR #275, opened 2022-05-09
Put the Node's name into the task name so it's recorded.

thetorpedodog:
While our Nodes have `name` fields, that was not correctly connected
with the `task_name` argument to the `whatever_base` functions, so they
were not recorded correctly in the logs. This plumbs that all the way
through.


aaronwolen:
Awesome. Thanks @thetorpedodog!

aaronwolen (approved):

PR #279, opened 2022-05-13
Create the Array Node and supporting code.

thetorpedodog:
Arguably the next most important node after the UDF node (which does all the work) is the Array Node (which makes this actually a TileDB task graph, rather than just a generic Python task graph). These two changes:

1. Create the result-processing infrastructure necessary to run array nodes.
2. Create the array node itself, integrate it, and add tests.

This and subsequent changes should be a lot smaller and friendlier than the last couple, now that all the hardest work is done.


antalakas (approved):
lgtm

PR #282, opened 2022-05-17
Implement Python registration API.

thetorpedodog:
The basics of the registration parts of the task graph API:

- Register a task graph
- Load a task graph
- Modify a task graph
- Delete a task graph

Includes all sorts of lovely tests.

---

This isn’t going to fully work until the necessary changes are deployed server-side, but I have tested this against staging.

[sc-14031]


antalakas (approved):
Tested on staging, lgtm

PR #285, opened 2022-05-19
Upgrade to cloudpickle 2.0.

thetorpedodog:
CloudPickle 2.0 isn't a breaking change, and the pickles it produces
remain compatible (since the unpickler is unchanged).


JohnMoutafis:
Not relevant anymore.
Already using cloudpickle v.2.2.1

PR #286, opened 2022-05-19
Create unified taskgraphs namespace.

thetorpedodog:
Creates a unified namespace where users can import everything they'll need for task graphs in one place, and adds some initial tests to make sure it all works together. Further integration testing will come in follow-up changes.


antalakas (approved):
I like the idea

thetorpedodog:
Merging this despite the current state of the autotests because they all seem to be server-side flakiness. At least one of the Pythons has passed, and on an earlier rev of this change they all passed. There’s nothing here that should vary between versions.

PR #291, opened 2022-05-26
Add DelayedMultiArrayUDF function

Shelnutt2:
Add DelayedMultiArrayUDF function

### DIFF
--- a/tiledb/cloud/compute/delayed.py
+++ b/tiledb/cloud/compute/delayed.py
@@ -1,6 +1,9 @@
 import numbers
+from typing import Callable, Union
 
+from tiledb.cloud.array import ArrayList
 from tiledb.cloud.array import apply as array_apply
+from tiledb.cloud.array import exec_multi_array_udf
 from tiledb.cloud.dag.dag import DAG
 from tiledb.cloud.dag.dag import Node
 from tiledb.cloud.sql import exec as sql_exec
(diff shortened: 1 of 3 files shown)


tiledb/cloud/compute/delayed.py
```diff
@@ -170,6 +172,39 @@ def __call__(self, *args, **kwargs):
         return self
 
 
+class DelayedMultiArrayUDF(DelayedBase):
+    def __init__(
+        self,
+        func: Union[str, Callable, None] = None,
```
thetorpedodog, line 178: Neither of these should have a default value, since they both need to be set.

thetorpedodog (approved):
Make sure you run `black` and `isort` to fix formatting issues and this one other thing:

tiledb/cloud/compute/delayed.py
Shelnutt2: @thetorpedodog I took this from the underlying function, should we change it there too?

thetorpedodog: No, that implements support for the legacy `name` argument, when you could pass the function as `name='some_registered_udf'`, but this is not supported in the Delayed system.

Shelnutt2: I meant from `exec_multi_array_udf_base`

thetorpedodog: That’s what I’m referring to.

Shelnutt2: Thanks for clarifying!

tiledb/cloud/compute/delayed.py
```diff
@@ -170,6 +173,39 @@ def __call__(self, *args, **kwargs):
         return self
 
 
+class DelayedMultiArrayUDF(DelayedBase):
+    def __init__(
+        self,
+        func: Union[str, Callable, None],
```
thetorpedodog, line 179: No need for None in this Union anymore. (Sorry for missing that earlier.)

thetorpedodog (approved):

tiledb/cloud/compute/delayed.py
thetorpedodog: and fyi if you push a change for this I'll merge it immediately assuming at least a couple of the tests finish cleanly because there's no real need to wait for all of them.

PR #306, opened 2022-06-29
Add retries into legacy task graph API.

thetorpedodog:
This change adds the ability to retry failed tasks to the legacy task
graphs/delayed API. Doing this required effectively rewriting most of
the way execution is handled. There's not much that can be done to
disentangle these changes.

Externally-visible changes are:

- Adds `retry()` method to Node and `retry_all()` to DAG.

- Nodes and DAGs can now "finish" multiple times and may be reset when
  they are retried.

- Adds a PARENT FAILED state, to ensure that we can differentiate
  between nodes that fail themselves or are cancelled and nodes that
  are unable to run because one of their parents failed.  For consumers
  using the Future-style API, this is treated like cancellation.

- There may be minor changes in the ordering of events when callbacks
  are called. Clients should not be adversely affected by these,
  since Nodes and DAGs now provide improved correctness.

DAG class changes:

- Add synchronization and lock things with a condition variable.
  This both preserves internal invariants and allows us to efficiently
  signal state changes (using wait/notify rather than polling).
  This was originally chosen to be less intrusive to the implementation
  than rewriting the entire thing with an event loop. Whether this
  ended up being true is debatable.

- Adds new "node executor" separate from the "task executor". This
  ensures that we can run the node lifecycle code (starting, parameter
  management, etc.) separately from the server-side calls. This resolves
  an issue where using processes instead of threads was no longer
  working (as it is impossible to serialize a lock to another process).

Node class changes:

- Handles lifecycle management in-house rather than delegating to an
  internal Future. We use the lifecycle condition as a lock to ensure
  access is properly synchronized and invariants are held.

- Implements the entire public API of Future in the Node itself,
  so that we don't need to proxy results to a proper futures.Future
  only for the purposes of returning to the user.  This is also
  necessary so that we can handle the "resettable future" semantics
  that retry functionality requires.

The trickiest parts of the control flow come from the way locking works.

- All callbacks always must be called when the object's condition lock
  is *not* held. Otherwise, the callback may try to re-enter the lock
  which is already held on the same thread. This is not allowed because
  we use traditional locks (not reentrant locks) so we're always
  guaranteed that at the time we take a lock, our invariants hold,
  and likewise that we have restored all applicable invariants when we
  release the lock.

- We can't `.cancel()` nodes while the DAG holds its own lock. This is
  an implementation-specific detail of the way both Python's stdlib
  Futures work and our reimplementation within Node does:

  When a lifecycle action like `.cancel()` is called, it immediately
  calls its own callback functions on the same thread. One of the
  actions is to notify the DAG that the node failed, which happens
  synchronously. This takes the DAG's lock, which is already held.

  This is also why setting PARENT FAILED happens in the `_dag_exec`
  function, which executes later on the node runner thread.

### DIFF
--- a/tiledb/cloud/dag/dag.py
+++ b/tiledb/cloud/dag/dag.py
@@ -7,7 +7,6 @@ import threading
 import time
 import uuid
 import warnings
-from concurrent import futures
 from typing import (
     Any,
     Callable,
@@ -32,6 +31,7 @@ from tiledb.cloud import sql
 from tiledb.cloud import tiledb_cloud_error as tce
 from tiledb.cloud import udf
 from tiledb.cloud import utils
+from tiledb.cloud._common import futures
 from tiledb.cloud._common import visitor
 from tiledb.cloud._results import results
 from tiledb.cloud._results import stored_params
--- a/tiledb/cloud/dag/dag.py
+++ b/tiledb/cloud/dag/dag.py
@@ -139,10 +139,6 @@ class Node(futures.FutureLike[_T]):
     def name(self, to: Optional[str]) -> None:
         self._name = to
 
-    @property
-    def error(self) -> Optional[Exception]:
-        return self._lifecycle_exception or self._exception
-
     def _find_deps(self):
         """Finds Nodes this depends on and adds them to our dependency list."""
         parents = _find_parent_nodes((self.args, self.kwargs))
@@ -207,7 +203,7 @@ class Node(futures.FutureLike[_T]):
     def result(self, timeout: Optional[float] = None) -> _T:
         with self._lifecycle_condition:
             self._wait(timeout)
-            to_raise = self._lifecycle_exception or self._exception
+            to_raise = self._error()
             if to_raise:
                 raise to_raise
             result = self._result
@@ -226,21 +222,14 @@ class Node(futures.FutureLike[_T]):
             self._cb_list.append(fn)
             if not self._done():
                 return
-        fn(self)
+        try:
+            fn(self)
+        except Exception:
+            pass
 
     # Bonus public methods.
 
     finished = done  # Alias.
-
-    def ready_to_compute(self) -> bool:
-        """
-        Is the node ready to execute? Are all dependencies completed?
-        :return: True if node is able to be run
-        """
-        if self.status is not Status.NOT_STARTED:
-            return False
-
-        return all(node.finished() for node in self.parents.values())
 
     @property
     def future(self) -> futures.FutureLike[_T]:
@@ -251,6 +240,10 @@ class Node(futures.FutureLike[_T]):
     def status(self) -> st.Status:
         with self._lifecycle_condition:
             return self._status
+
+    @property
+    def error(self) -> Optional[Exception]:
+        return self._error()
 
     def retry(self) -> bool:
         if not self.dag:
@@ -287,6 +280,9 @@ class Node(futures.FutureLike[_T]):
             self._wait(timeout)
 
     # Things for internal use and use by the DAG.
+
+    def _error(self) -> Optional[Exception]:
+        return self._lifecycle_exception or self._exception
 
     def _reset_internal(self) -> bool:
         """Prepares this Node to be retried if it failed."""
@@ -662,9 +658,9 @@ class DAG:
             st.Status.COMPLETED,
         )
 
-    def done(self):
+    def done(self) -> bool:
         with self._lifecycle_condition:
-            return self._done
+            return self._done()
 
     def add_node_obj(self, node):
         """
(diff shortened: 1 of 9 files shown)


thetorpedodog:
[sc-16904]

tiledb/cloud/dag/dag.py
```diff
@@ -777,28 +963,36 @@ def wait(self, timeout=None):
...
+            raise exc
 
     def cancel(self):
-        if self.status not in (st.Status.RUNNING, st.Status.NOT_STARTED):
-            return
-        for node in list(self.running_nodes.values()):
-            node.cancel()
-        for node in list(self.not_started_nodes.values()):
-            node.cancel()
+        with self._lifecycle_condition:
```
ktsitsi, line 976: Not sure if this is in the scope of this PR but some functions annotations like this one missing the return type. If no please ignore.

ktsitsi (commented):
LGTM

tiledb/cloud/dag/dag.py
```diff
@@ -169,49 +171,193 @@ def depends_on(self, node: "Node"):
...
+            if self._lifecycle_exception:
+                raise self._lifecycle_exception
+            return self._exception
+
+    def add_done_callback(self, fn: Callable[["Node[_T]"], None]) -> None:
+        with self._lifecycle_condition:
+            self._cb_list.append(fn)
+            if not self._done():
+                return
+        fn(self)
```
gsakkis, line 229: The respective `Dag` method guard this in try/except (+ also checks that `fn` is indeed callable); is there a reason they're not identical?

tiledb/cloud/dag/dag.py
```diff
@@ -689,44 +832,90 @@ def report_node_complete(self, node: Node):
...
+        with self._lifecycle_condition:
+            # A node may be either "running" or "not started" depending upon
+            # whether it failed or was cancelled.
+            self.running_nodes.pop(node.id, None)
+            self.not_started_nodes.pop(node.id, None)
+
+            if node.status is st.Status.COMPLETED:
+                self.completed_nodes[node.id] = node
+                for child in node.children.values():
+                    self._maybe_exec(child)
```
gsakkis, line 844: This also happens at line 859; is this deliberate?

thetorpedodog:
Step 1 before addressing further feedback: rebase changes onto tip; separate out the creation of `_common/futures.py`.

tiledb/cloud/dag/dag.py
thetorpedodog: Added a try-block here. I omitted the check found in `add_done_callback` because it's kind of superfluous, given that we have the type annotations to guide the user.

tiledb/cloud/dag/dag.py
```diff
@@ -169,49 +171,193 @@ def depends_on(self, node: "Node"):
...
-        Is the node finished
-        :return: True if the node's function is finished
+        Is the node ready to execute? Are all dependencies completed?
+        :return: True if node is able to be run
         """
-        return self._future.done()
+        if self.status is not Status.NOT_STARTED:
+            return False
+
+        return all(node.finished() for node in self.parents.values())
```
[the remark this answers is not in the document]
thetorpedodog: Actually, ready_to_compute was really only for internal usage inside `DAG`. `DAG` no longer uses it, so there's no need for it to exist anymore.

tiledb/cloud/dag/dag.py
```diff
@@ -514,36 +655,36 @@ def _report_server_completion(self) -> None:
...
+    def _done(self):
+        return self._status in (
             st.Status.FAILED,
             st.Status.CANCELLED,
             st.Status.COMPLETED,
         )
 
+    def done(self):
+        with self._lifecycle_condition:
+            return self._done
```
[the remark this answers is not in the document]
thetorpedodog: it should!

tiledb/cloud/dag/dag.py
thetorpedodog: It is not. It didn't *hurt* anything (`_maybe_exec` is safe to call repeatedly) but definitely not the intended result.

thetorpedodog (commented):
Thank you both for the reviews! I know this is (to put it generously) not a very pleasant change.

> Added a few minor comments and questions but overall I'm afraid I cannot vouch for it in terms of overall logical and concurrency correctness; there's too much going on to wrap my head around in the ~2 hours I spent on it.

Yeah. There's a lot going on here and (apart from what I just pulled out) not much more I think I can do to break it down.

gsakkis (approved):
:ship: it!

tiledb/cloud/dag/dag.py
thetorpedodog: I've been adding annotations to functions in an ad-hoc manner over the course of working on this module. It admittedly would probably be good to add them to everything but it's not a priority.

PR #310, opened 2022-07-07
Update documentation and workflows to use `main`.

thetorpedodog:
The default branch is now `main`, so things should be updated to reflect that.


johnkerl (approved):
🐑

PR #311, opened 2022-07-12
Add manual-layer methods for retrieving group metadata, and recursively cleaning up

johnkerl:
We needn't wrap the entire groups-API layer -- it is fine.

```
#!/usr/bin/env python

import tiledb.cloud
import tiledb.cloud.group
import tiledb.cloud.client

api_instance = tiledb.cloud.client.client.groups_api

page_index = 1
while True:
    page = api_instance.list_owned_groups(page=page_index)
    if page.groups == None:
        break
    for group in page.groups:
        print("%-30s %s" % (group.name, group.tiledb_uri))
    page_index += 1
```

with output

```
...
pbmc-small                     tiledb://johnkerl-tiledb/d2f37ba0-4772-4fdd-9bef-c298a97a50f5
pbmc3k_processed               tiledb://johnkerl-tiledb/84fca634-97a9-407f-b444-73bb92378281
raw                            tiledb://johnkerl-tiledb/12ea94b0-1a0e-4485-b046-54fa237ba4ae
soco123                        tiledb://johnkerl-tiledb/0fb98290-8e30-4b43-bd05-1a9b16d67ea6
soco3                          tiledb://johnkerl-tiledb/02e2484e-156b-4b2a-a20a-6e13c4a9bb07
try010                         tiledb://johnkerl-tiledb/c8aa3903-982f-4378-bbf1-aab36f24d132
try011                         tiledb://johnkerl-tiledb/80e31f45-f819-4b1f-b7aa-e848b0e38f56
try012                         tiledb://johnkerl-tiledb/062fbcf8-95b8-4aaa-801f-1007aac456a4
try013                         tiledb://johnkerl-tiledb/a541a85e-4082-40b5-a5ca-2b9fc5aceeb2
try030                         tiledb://johnkerl-tiledb/98e80d54-85ee-4ad9-aed4-2bd2c1341043
uns                            tiledb://johnkerl-tiledb/24e2b0dd-536c-4a11-867d-a62a97089a0f
uns                            tiledb://johnkerl-tiledb/747a2c4b-c514-4cf2-85cd-2759e33cc621
uns                            tiledb://johnkerl-tiledb/b2b27605-54c1-4c37-82cd-a13d93bdd090
varm                           tiledb://johnkerl-tiledb/89c3fcd7-6bd4-48eb-bd97-76c90910c924
varp                           tiledb://johnkerl-tiledb/62e0aa64-895a-4b47-b1d1-3cb8c2d4bbf7
X                              tiledb://johnkerl-tiledb/c037dc39-3b3d-4ad0-8484-2b2e006a40df
...
```

However, we can wrap a few convenient things:

* API-level methods (from OpenAPI autogen) tend to take `namespace` and `group_name` as separate args. `get_group` on this PR is an example of wrapping a simple API-level method with a manual-layer method whose chief value-add is letting users specify a URI like `tiledb://namespace/group_name`, which we will split for them into `('namespace', 'group_name')`
* Also on this PR we have a recursive cleaner-upper
* We can do more such methods, on this PR or (preferably) subsequents! :)


tiledb/cloud/group.py
```diff
@@ -0,0 +1,127 @@
...
+    """
+    Returns the cloud metadata
+
+    :return: metadata object
+    """
+    (namespace, group_name) = split_uri(uri)
+    api_instance = client.client.groups_api
+
+    try:
+        return api_instance.get_group(namespace, group_name, async_req=async_req)
```
thetorpedodog, line 23: `async_req` is never defined here; it's false by default so you can just remove it entirely:
```suggestion
        return api_instance.get_group(namespace, group_name)
```

tiledb/cloud/group.py
```diff
@@ -0,0 +1,127 @@
...
+    (namespace, group_name) = split_uri(uri)
+    api_instance = client.client.groups_api
+
+    try:
+        array_tiledb_uris = []
+        group_tiledb_uris = []
+        page_index = 1
+        while True:
+            page = api_instance.get_group_contents(
+                namespace, group_name, page=page_index
```
thetorpedodog, line 99: It's probably best to include a page size as well.

tiledb/cloud/group.py
```diff
@@ -0,0 +1,127 @@
...
+    """
+    (namespace, group_name) = split_uri(uri)
+    api_instance = client.client.groups_api
+    try:
+        return api_instance.delete_group(namespace, group_name)
+    except GenApiException as exc:
+        raise tiledb_cloud_error.check_exc(exc) from None
+
+
+def _get_group_element_tiledb_uris(uri: str) -> Dict[str, List[str]]:
```
thetorpedodog, line 83: Instead of a dict, I would return either a pair of lists here. While in most cases I am wary of using undifferentiated pairs, in this case since it's an internal function, I think it's probably fine given appropriate naming, something like:

`def _subgroups_and_arrays(uri: str) -> Tuple[List[str], List[str]]:`

johnkerl:
> Is there an easy way for you to add a test for this? Create a group and a bunch of arrays that live in it, then remove them, maybe?

@thetorpedodog perhaps best to imitate the existing `tiledb.cloud.array` -- which AFAICT has none? Do you have an example to recommend before I cut my own attempt, unskillfully, from whole cloth?

thetorpedodog:
> I'm not seeing a mechanism for writeful tests against cloud (prod, staging, etc.) -- looks like I'll be doing that for the first time within this repo?

We have a few tests that do writes, though basically all of them create UDFs or task graphs rather than regular arrays.

A test that creates an array would look something like:

```python
test_array = create_test_array(...)
try:
  # ... do stuff with test_array
finally:
  remove_if_present(test_array)
```

> I'm still trying to reverse-engineer test-creds -- not sure yet if this token is writeful (I suspect not):
> 
> ```
> $ rg --hidden TILEDB_CLOUD_HELPER_VAR
> .github/workflows/tiledb-cloud-py.yaml
> 98:          TILEDB_REST_TOKEN: ${{ secrets.TILEDB_CLOUD_HELPER_VAR }}
> ```

`TILEDB_CLOUD_HELPER_VAR` is a repository secret; it’s a login token for the `unittest` user. For local work, so long as you’ve logged into TileDB Cloud at some point (e.g. with `tiledb.cloud.login(...)`), you should have a stored credential that you can use in tests. They will run against your own account—all they do is perform some queries and occasionally create temporary arrays. (`test_array_activity` and `test_list_shared_arrays` will probably both fail, which is fine, since it expects the state of the `unittest` account, not your account). This is how I universally run my own local tests.

You can run the tests with:

```console
(venv) path/to/TileDB-Cloud-Py$ pytest -vs -k group
```

`-v` is verbose, `-s` prevents capturing output, `-k group` filters to only tests that contain `group` in the name.

thetorpedodog:
To save some time, I did a commit which should fix the import problems you were running into, and made a few other minor fixes along the way. Details are in the message and of course you are free to work from this and make further changes as you see fit.

PR #312, opened 2022-07-14
Skip tests that depend upon the unittest user to pass.

thetorpedodog:


johnkerl:
@thetorpedodog I just saw #312, will update wording :)

johnkerl (approved):
🚢

PR #315, opened 2022-07-30
Add and test support for query layouts.

thetorpedodog:
A feature I left as a TODO initially, turned out to be even less work
than I anticipated. Added with some nice enum-based type safety.

### DIFF
--- a/tiledb/cloud/taskgraphs/types.py
+++ b/tiledb/cloud/taskgraphs/types.py
@@ -1,7 +1,8 @@
 """User-facing types used in task graphs."""
 
+import enum
 import itertools
-from typing import Any, Dict, List, Tuple, TypeVar, Union
+from typing import Any, Dict, List, Optional, Tuple, TypeVar, Union
 
 import attrs
 import numpy as np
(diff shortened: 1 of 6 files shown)


tiledb/cloud/taskgraphs/types.py
```diff
@@ -43,6 +44,38 @@
 """Type returned from an array query."""
 
 
+class Layout(enum.Enum):
+    """The layout of a TileDB query."""
+
+    ROW_MAJOR = "R"
```
Shelnutt2, line 50: @ihnorton @nguyenv 

I know I originally used these layout types but should we sync with python enums?

thetorpedodog: If there’s already enums for this in the base Python library I would be thrilled to use them.

thetorpedodog: Or, since there aren’t at the moment (from what I can tell), put them here for now, and then later upstream them, replacing this version with an alias.

Shelnutt2 (approved):

PR #316, opened 2022-08-01
The first part of the Delayed API 2.0: graph management.

thetorpedodog:
Task graphs in the new Delayed API are managed by the DelayedGraph
class, which acts as an abstraction layer over the Builder and Executor
classes. The first core task of the DelayedGraph is to manage the
structure of its own graphs before it can build or execute them.

### DIFF
--- a/tests/taskgraphs/delayed/test_build.py
+++ b/tests/taskgraphs/delayed/test_build.py
@@ -1,0 +1,44 @@
+import unittest
+
+from tiledb.cloud.taskgraphs.delayed import _graph
+
+
+class DummyNode(_graph.Node):
+    @classmethod
+    def create(cls, *parents: _graph.Node):
+        mgr = _graph.Merger()
+        mgr.visit(parents)
+        new_owner = mgr.merge_visited()
+        new = cls(new_owner)
+        new_owner._add(new, parents=mgr.unexecuted_nodes)
+        return new
+
+
+class TestBuild(unittest.TestCase):
+    def test_merge(self):
+        node_0 = DummyNode.create()
+        owner_0 = node_0._owner
+        self.assertIsNotNone(owner_0)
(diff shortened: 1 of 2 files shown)


thetorpedodog:
part of [sc-19584]

thetorpedodog:
The best place for examples on what the API will look like is in the [test cases of the everything-rolled-into-one-commit version of this](https://github.com/TileDB-Inc/TileDB-Cloud-Py/blob/b60a56317cd31a61bb4fe710db8812cc09221614/tests/taskgraphs/delayed/test_delayed_complex.py). But for a brief sampler:

```python
some_val = Delayed(my_func, ...)(foo, bar, name=baz)
#                  +-----------  +-----------------
#                  |our args     | the user's args

downstream = Delayed(another_func)(some_val)

output_from_array = Delayed("registered/udf")(
    DelayedArray(uri, ...),
    other,
    inputs,
    ...,
)

result = output_from_array.compute(
    timeout=10,
    name="name-of-execution",
)

# output_with_array also has the various Future methods and other bonus
# lifecycle stuff that you get, like `retry`, `retry_all`, info about
# the DAG that is running it, etc.

# Registration will be accomplished roughly like this
# (exact details may change but this is generally it):
output_from_array.register("registered_graph_name")
```

The biggest changes are:
- the *strict* separation of our arguments from the user's arguments. We don’t reserve any names that the user can’t pass to their functions; everything we take is specified in the `Delayed` half of the call.
- array inputs can be specified with a `DelayedArray` call. This makes it possible (and easy!) to pass multiple arrays to one Delayed function, rather than having to build an ArrayList etc. (We can also add a `DelayedArrayUDF`-style shortcut, but would just be a wrapper that builds you a `Delayed` and a `DelayedArray`.)

antalakas (approved):
Thanks @thetorpedodog

PR #318, opened 2022-08-02
Allow nodes of modern task graphs to be executed locally.

thetorpedodog:
This is a first implementation of local execution for task graphs.
I've assumed that local nodes may be compute-intensive, so they always
run under a new process rather than a thread on the current one.
We assume UDF components are more or less independent and as such should
not rely on communicating within the same thread, so thread-based
local execution is a niche use case.

This depends upon this change being submitted and deployed server-side;
registration of UDFs to be executed on the client side will not work and
the registration-related tests will not pass until then.

https://github.com/TileDB-Inc/TileDB-Cloud-API-Spec/pull/355

Open questions:

- Is always running these in separate processes a good idea?
  It does lead to a simpler API.

- Do we want to automatically fall back to running server-side
  if we can't run client-side, or should we require that be explicitly
  requested somehow?

Improvements to come to this changeset:

- Better testing of local execution.
- Other cleanup.

Improvements planned for follow-ons:

- Use a process pool, stored in the Executor, rather than creating
  a new process for each UDF call.
- Separate the count of UDFs that are allowed to be run in parallel
  on the server side from those on the client side (i.e. even if you're
  already running your maximum of 5 UDFs on the server, client-side
  execution has a separate quota pool).

### DIFF
--- a/tests/taskgraphs/test_client_executor.py
+++ b/tests/taskgraphs/test_client_executor.py
@@ -1,11 +1,13 @@
 import datetime
 import operator
+import tempfile
 import time
 import unittest
 
 import numpy
 import pyarrow
 
+from tiledb.cloud._common import futures
 from tiledb.cloud.taskgraphs import builder
 from tiledb.cloud.taskgraphs import client_executor
 from tiledb.cloud.taskgraphs import executor
@@ -84,6 +86,62 @@ class ClientExecutorTestUDFs(unittest.TestCase):
         self.assertEqual("the value is really 0.0", exec.node(succeedchild).result(10))
         exec.wait(10)
         self.assertIs(executor.Status.FAILED, exec.status)
+
+
+class ClientExecutorTestLocal(unittest.TestCase):
+    def test_basic(self):
+        grf = builder.TaskGraphBuilder(name="test_local")
+        double = grf.udf(lambda x: 2 * x, types.args(2), local=True)
+
+        exec = client_executor.LocalExecutor(grf, name="test_local exec")
+        exec.execute()
+        double_exec = exec.node(double)
+        self.assertEqual(4, double_exec.result(30))
+        self.assertIsNone(double_exec.task_id())
+        exec.wait(10)
+
+    def test_concurrent(self):
+        import os
+        import os.path
+        import time
+
+        def touch_and_wait(dir, file):
+            full_path = os.path.join(dir, file)
+            open(full_path, "w").close()
+            while os.path.exists(full_path):
+                time.sleep(0.01)
+            return file
+
+        with tempfile.TemporaryDirectory() as tmpdir:
+            grf = builder.TaskGraphBuilder()
+            one = grf.udf(touch_and_wait, types.args(tmpdir, "one"), local=True)
+            two = grf.udf(touch_and_wait, types.args(tmpdir, "two"), local=True)
+            out = grf.udf(" ".join, types.args((one, two)), local=True)
+            exec = client_executor.LocalExecutor(grf)
+            exec.execute()
+            with self.assertRaises(futures.TimeoutError):
+                exec.node(one).result(1)
+            with self.assertRaises(futures.TimeoutError):
+                exec.node(two).result(1)
+
+            p_one = os.path.join(tmpdir, "one")
+            p_two = os.path.join(tmpdir, "two")
+
+            while not os.path.exists(p_one):
+                time.sleep(0.01)
+            while not os.path.exists(p_two):
+                time.sleep(0.01)
+            os.remove(p_one)
+            os.remove(p_two)
+            self.assertEqual("one two", exec.node(out).result(5))
+
+    def test_timeout(self):
+        grf = builder.TaskGraphBuilder()
+        timer = grf.udf(time.sleep, types.args(10), timeout=2, local=True)
+        exec = client_executor.LocalExecutor(grf)
+        exec.execute()
+        with self.assertRaises(TimeoutError):
+            exec.node(timer).result(5)
 
 
 class ClientExecutorTestArrays(unittest.TestCase):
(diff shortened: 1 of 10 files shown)


thetorpedodog:
This is now ready, since these questions have been answered:

> * Is always running these in separate processes a good idea?
>   It does lead to a simpler API.

**Yes.** Because we always pickle functions when building the task graph, there is no reason *not* to run things in separate processes.

> * Do we want to automatically fall back to running server-side
>   if we can't run client-side, or should we require that be explicitly
>   requested somehow?

**No.** After some discussion with normanb, we agreed that since the primary documented use case of local UDFs is interacting with local resources (file system, etc.), automatically retrying server side is *not* a good idea. However, in the future, something like `.retry(allow_server_side=True)` (or a similar-looking API) is probably a good idea.

Shelnutt2:
> Is always running these in separate processes a good idea?

I do not think so. This means pickling all data back and forth between processes. I think this is the opposite of the expected behavior and the common use xases. Local items are never compute intensive. If they are compute intensive they are remote.

Please schedule time to discuss this. I find it concerning that the common use cases and patterns are not considered. We need to align all work with our use cases.

thetorpedodog:
> > Is always running these in separate processes a good idea?
> 
> I do not think so. This means pickling all data back and forth between processes. I think this is the opposite of the expected behavior and the common use xases. Local items are never compute intensive. If they are compute intensive they are remote.
> 
> Please schedule time to discuss this. I find it concerning that the common use cases and patterns are not considered. We need to align all work with our use cases.

When the graph is at this point in execution, everything is already in serialized form; there is effectively no penalty to serialize and deserialize them across this boundary.

As far as behavior from the user perspective is concerned, there won’t be any observable difference between tasks running in the same process vs. tasks running in different processes (except for the part about the GIL). There’s nothing *wrong* with running things in the same process, but that’s the only thing that will change.

Given that:

- Using a process pool is an option in the old DAG API
- The benefits of executing in a separate process were available with minimal cost in terms of complexity
- The end result would work the same in other respects

that’s what led me to conclude that running in separate processes makes sense.

tests/taskgraphs/test_client_executor.py
```diff
@@ -86,6 +88,62 @@ def test_failure(self):
...
+            with self.assertRaises(futures.TimeoutError):
+                exec.node(two).result(1)
+
+            p_one = os.path.join(tmpdir, "one")
+            p_two = os.path.join(tmpdir, "two")
+
+            while not os.path.exists(p_one):
+                time.sleep(0.01)
+            while not os.path.exists(p_two):
+                time.sleep(0.01)
```
gsakkis, line 133: Are these `sleep` calls necessary? IIRC the files must be created after line 125

gsakkis (approved):
Some cosmetic comments but LGTM overall :rocket: (at least as far as I could follow the several abstractions involved)

tests/taskgraphs/test_client_executor.py
thetorpedodog: Not quite; as of lines 122–125 the `touch_and_wait` tasks are stuck on their own sleep-loop (lines 111–112) waiting for the files to be removed. I’ve added some more comments explaining the sequencing of the test.

tiledb/cloud/taskgraphs/client_executor/_replacers.py
```diff
@@ -81,3 +84,71 @@ def maybe_replace(self, arg) -> Optional[visitor.Replacement]:
...
+    ``kwargs`` values, without performing any parsing of the actual values
+    inside each of the arguments (that is, it turns the JSON list into a
+    list of args and a dict of kwargs, but does not perform further processing).
+    """
+    # Usually we don't include internal type-checks, but this function is also
+    # used on the server-side UDF executor, where we do want to ensure that
+    # we don't get the wrong type of data.
+    arg_type = type(json_args)
+    if arg_type is not list:
+        raise ValueError(
```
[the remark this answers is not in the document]
thetorpedodog: Good question! This is data that comes out of `json.Parse`, so any (non-`str`) sequences that we get will always be `list`s rather than anything else. This is technically not *strictly* necessary but is a useful check to ensure that everything is parsing as we expect.

tiledb/cloud/taskgraphs/client_executor/udf_node.py
```diff
@@ -46,20 +67,85 @@ def task_id(self, timeout: Optional[float] = None) -> Optional[uuid.UUID]:
...
         self, parents: Dict[uuid.UUID, _base.Node], input_value: Any
     ) -> None:
         assert input_value is _base.NOTHING
 
+        if self._is_local():
+            self._exec_local(parents)
+            # TODO: Add a way that users can specify that they want to retry
+            # a local UDF remotely.
+            return
+        self._exec_remote(parents)
```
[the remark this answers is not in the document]
thetorpedodog: That does make sense for now; I think the original way I was doing it was from a time when there was more logic and it would have made sense to fall through to `_exec_remote` based if `_exec_local` failed. Since we no longer do that, a regular if-else is indeed better.

tiledb/cloud/taskgraphs/builder.py
```diff
@@ -497,14 +501,22 @@ def __init__(
...
         name: Optional[str],
+        local: bool,
     ):
         """Initializes this UDF node.
 
         See :meth:`TaskGraphBuilder.udf` for details.
         """
         utils.check_funcable(func=func)
+        if isinstance(func, str):
+            if local:
```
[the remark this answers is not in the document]
thetorpedodog: Whoops! I think this was another leftover from when there was more/different logic here.

thetorpedodog (commented):
Thanks for the review!

tests/taskgraphs/test_client_executor.py
gsakkis: >  the touch_and_wait tasks are stuck on their own sleep-loop (lines 111–112) waiting for the files to be removed

Right, but before entering their sleep loop they create the files. This implies `os.path.exists(p_one) and os.path.exists(p_two)` after the timeout errors, in which case the `sleep()` at lines 131 and 133 are never called, no? (sorry if I'm still missing it).

thetorpedodog: While unlikely, the tasks may take a little while to spin up, in which case even after 1 second the creation may not have yet occurred. Waiting here to ensure that the files *do* exist guarantees the correctness of the test.

PR #321, opened 2022-08-03
Wire up DelayedGraph executor and create initial delayed UDFs.

thetorpedodog:
These two changes together create the most complex and most important parts of Delayed execution. The two steps are technically somewhat independent, but without an executor, there’s not really much for a delayed UDF to do, and without a delayed UDF, there’s not much for an executor to execute. Hence, while they’re in two separate changes, I have introduced them here as one review.

This change is long, but a lot of it is docstrings, mostly targeting end-user developers who are trying to build things with delayed task graphs.

### DIFF
--- a/tests/taskgraphs/delayed/test_delayed_funcs.py
+++ b/tests/taskgraphs/delayed/test_delayed_funcs.py
@@ -1,0 +1,132 @@
+import operator
+import unittest
+
+import numpy as np
+
+from tiledb.cloud import client
+from tiledb.cloud import testonly
+from tiledb.cloud._common import futures
+from tiledb.cloud.taskgraphs import delayed
+from tiledb.cloud.taskgraphs import executor
+
+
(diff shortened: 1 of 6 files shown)


tests/taskgraphs/delayed/test_delayed_funcs.py
```diff
@@ -0,0 +1,132 @@
...
+
+        self.assertEqual(2, node_1.result())
+        self.assertEqual(4, node_2.result())
+        self.assertEqual(8, node_3.result())
+
+    def test_kwargs(self):
+        def string_multi(multiplier, st=None):
+            return int(multiplier) * st
+
+        node_1 = delayed.udf(np.median, name="node_1")([1, 2, 3])
```
antalakas, line 49: I really like the separation

antalakas (approved):

PR #323, opened 2022-08-04
Add the Array Node, which enables Array UDFs.

thetorpedodog:
This adds and tests the Array node, so that users can actually start
doing interesting TileDB-specific things with their UDFs by passing in
their TileDB arrays as values.

This also streamlines the process of default-handling with `NOTHING`,
by making a central place to strip `NOTHING`s out of kwarg dictionaries.

### DIFF
--- a/tests/taskgraphs/delayed/test_delayed_complex.py
+++ b/tests/taskgraphs/delayed/test_delayed_complex.py
@@ -7,6 +7,7 @@ import tiledb.cloud
 from tiledb.cloud import testonly
 from tiledb.cloud.taskgraphs.delayed import Delayed
 from tiledb.cloud.taskgraphs.delayed import DelayedArray
+from tiledb.cloud.taskgraphs.delayed import DelayedArrayUDF
 
 SPARSE = "tiledb://TileDB-Inc/quickstart_sparse"
 DENSE = "tiledb://TileDB-Inc/quickstart_dense"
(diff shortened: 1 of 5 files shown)


antalakas (approved):

antalakas (changes requested):
Probably we need to address this https://github.com/TileDB-Inc/TileDB-Cloud-Py/pull/324#discussion_r939183623 comment

thetorpedodog:
Re @Shelnutt2’s comment on the other change:

> This should be `DelayedArrayUDF` to match the existing API. `DelayedArray` also has special meaning in the ecosystem which is why we used the specific name of `DelayedArrayUDF`.

I think the use of `DelayedArray` here is correct; the object you get back from `DelayedArray` is something you can *pass into* a UDF to pass the array into your UDF.

Separately, I’ve added a `DelayedArrayUDF` function which combines `Delayed` and `DelayedArray` together into one, and matches up with the existing `DelayedArrayUDF` API:

```python
arr_udf = DelayedArrayUDF(
    some_func,
    "tiledb://array/path",
    **array_kwargs,
    **delayed_kwargs,
)(other, inputs)
# Will eventually call some_func(array_contents, other, inputs)

# equivalent to

arr = DelayedArray("tiledb://array/path", **array_kwargs)
arr_udf = Delayed(some_func, **delayed_kwargs)(arr, other, inputs)
# Also will eventually call some_func(array_contents, other, inputs)
```

antalakas (commented):
I am good with this, but I'd like an approve from @Shelnutt2 also

JohnMoutafis:
Closing this as it is not relevant anymore.

PR #324, opened 2022-08-05
Delayed 2.0

thetorpedodog:


tiledb/cloud/taskgraphs/delayed/__init__.py
```diff
@@ -13,9 +13,18 @@
     d_sql = DelayedSQL(...)
 """
 
+from tiledb.cloud.taskgraphs.delayed import _nodes
 from tiledb.cloud.taskgraphs.delayed import _udf
 
 udf = _udf.DelayedFunction.create
 Delayed = udf
+array = _nodes.Array.create
+DelayedArray = array
```
Shelnutt2, line 22: This should be `DelayedArrayUDF` to match the existing API. `DelayedArray` also has special meaning in the ecosystem which is why we used the specific name of `DelayedArrayUDF`.

thetorpedodog: This was a draft that I set up to test commit 84491ff8b1282b1e9dcd7881dab2dd5510b9d1c3; to discuss the array part go to change #323.

PR #338, opened 2022-09-09
Reimplement the Delayed API atop the new task graph API.

thetorpedodog:
The existing Delayed API is an extremely thin layer on top of the old
DAG API, which is itself a relatively thin layer atop the various
`exec_something` functions. The new task graph API is neither of those
things. The implementation of the old one leaked a whole lot of
abstraction, and this new implementation represents a best-effort
attempt to match the behavior of the existing Delayed API atop the new
task graph API.

To the best of my knowledge and ability, existing code that uses
“standard” methods of the existing Delayed API should continue to work
unchanged. However, this breaks compatibility with essentially anything
that reaches into or otherwise depends upon the internals of the
existing Delayed objects, including any of the `Node` specifics or the
`delayed_obj.dag` DAG object. Common interactions where that takes place
have been replaced with methods on the common node type.

Unfortunately, this has to be one commit because, since the existing API
is one integrated package, it can only be replaced all at once.

---

The tests do not yet pass because this change depends upon functionality provided in #315 and #318.

### DIFF
--- a/tests/test_delayed.py
+++ b/tests/test_delayed.py
@@ -1,5 +1,7 @@
 import operator
-import threading
+import os
+import shutil
+import tempfile
 import unittest
 from concurrent import futures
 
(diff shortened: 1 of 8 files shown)


Shelnutt2 (approved):
:tada:

PR #339, opened 2022-09-13
Overwrite logic in notebook arrays

ktsitsi:
This PR will make easier the progress of [sc-20869], plus will enhance the `cloud` codebase with some semantics considered in the past as well. It introduces an `Enum` class `UploadOption` for selecting upon `uploading` the desirable logic. A prerequisite for overwriting logic is the existence of the notebook at the first place. The `auto-increment` logic will raise a `NotImplementedError` and the default functionality is the `FAIL` as had already been implemented and applied.

### DIFF
--- a/tiledb/cloud/notebook.py
+++ b/tiledb/cloud/notebook.py
@@ -243,6 +243,7 @@ def _create_notebook_array(
             "Please select a proper namespace (username or organization name).",
         )
 
+
     # The array will be be 1-dimensional with domain of 0 to max uint64. We
     # use a tile extent of 1024 bytes.
     dom = tiledb.Domain(
@@ -280,8 +281,9 @@ def _create_notebook_array(
                     raise tiledb_cloud_error.TileDBCloudError(
                         f"Error creating file: {array_name!r} already exists in namespace {namespace!r}."
                     )
-            # Retry other TileDB erors
-            tries -= 1
+            # Retry other TileDB errors
+            if upload_options is upload_options.FAIL:
+                tries -= 1
             if tries <= 0:
                 raise tiledb_cloud_error.check_exc(e) from None
 
--- a/tiledb/cloud/notebook.py
+++ b/tiledb/cloud/notebook.py
@@ -204,13 +204,17 @@ def upload_notebook_contents(
         {"rest.creation_access_credentials_name": storage_credential_name}
     )
 
-    tiledb_uri, array_name = _create_notebook_array(
-        storage_path,
-        array_name,
-        namespace,
-        ctx,
-        upload_options,
-    )
+    if upload_options is upload_options.FAIL:
+        tiledb_uri, array_name = _create_notebook_array(
+            storage_path,
+            array_name,
+            namespace,
+            ctx,
+            upload_options,
+        )
+    else:
+        # The array should already exist
+        tiledb_uri = "tiledb://" + posixpath.join(namespace, array_name)
 
     _write_notebook_to_array(tiledb_uri, ipynb_file_contents, ctx)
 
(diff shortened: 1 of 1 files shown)


tiledb/cloud/notebook.py
```diff
@@ -22,6 +22,14 @@
 CHARACTER_ENCODING = "utf-8"
 
 
+class UploadOption(enum.Enum):
```
thetorpedodog, line 25: `UploadOption` seems a bit too generic; how about something like `IfExists` (`OnExists`?) Then the docstring can be something like `"""Action to take if the array already exists."""`.

tiledb/cloud/notebook.py
```diff
@@ -22,6 +22,14 @@
...
+    """Kind of upload option."""
+
+    FAIL = enum.auto()
+    OVERWRITE = enum.auto()
+    AUTO_INCREMENT = enum.auto()
```
thetorpedodog, line 30: Since this isn’t implemented yet, I would leave it out of the enum and just make it a TODO to add in the future.

thetorpedodog (commented):
The implementation here is great! I really like the use of an enum here rather than a boolean parameter—it makes things more extensible and less ambiguous. I have a couple small suggestions but they’re only around naming and a TODO; the logic and structure itself are good.

antalakas (commented):
I am a little bit confused on this: Does it really overwrite the original array? I would expect it to update and create a new version of the notebook

thetorpedodog:
> I am a little bit confused on this: Does it really overwrite the original array? I would expect it to update and create a new version of the notebook

These seems synonymous to me. Am I missing something?

ktsitsi:
I have manually tested the functionality by `pip` installing the current branch and running the following on an existing notebook. 
```
tiledb.cloud.notebook.upload_notebook_from_file(..., upload_options=tiledb.cloud.notebook.UploadOption.OVERWRITE)
```
I checked also in UI and indeed on the same notebook I have two versions with the latest being the overwritten data, which makes sense since it is an extra `write` on an existing array right?

tiledb/cloud/notebook.py
ktsitsi: Yeap np!

thetorpedodog: I don’t see any new revisions—did you forget to upload? (Certainly something I have done before.)

tiledb/cloud/notebook.py
```diff
@@ -189,12 +202,16 @@ def upload_notebook_contents(
...
+    if on_exists is OnExists.FAIL:
+        tiledb_uri, array_name = _create_notebook_array(
+            storage_path,
+            array_name,
+            namespace,
+            ctx,
+        )
+    else:
+        # The array should already exist
+        tiledb_uri = "tiledb://" + posixpath.join(namespace, array_name)
```
georgeSkoumas, line 214: Minor. `f"tiledb://{posixpath.join(namespace, array_name)}"`

tiledb/cloud/notebook.py
```diff
@@ -133,6 +142,7 @@ def upload_notebook_from_file(
       user's account settings.
     :param storage_credential_name: such as "janedoe-creds", typically from the
       user's account settings.
+    :param if_exists: such as IfExists.FAIL (default), OVERWRITE or AUTO-INCREMENT
```
thetorpedodog, line 145:
```suggestion
    :param on_exists: such as OnExists.FAIL (default), OVERWRITE or AUTO-INCREMENT
```

thetorpedodog (approved):
One minor thing to make the docstring match up—this looks great!

antalakas:
> 

👍

antalakas (approved):

PR #341, opened 2022-10-18
Record the owner of a task graph builder node.

thetorpedodog:
This ensures that (a) a user creating multiple task graphs at the same time doesn't cross contaminate and that (b) a user can get from a builder node back to its owner.


thetorpedodog:
[Who Owns Who?](https://www.youtube.com/watch?v=Qt_l2xFcXt0)

awenocur (approved):

PR #343, opened 2022-10-31
Prepare for generating v2 of the API and regenerate v1

thetorpedodog:
When we generate v2 of the TileDB Cloud API, we will need to do a few things to ensure that it will continue to work with common configuration etc. with the v1 version:

- we have to prepend the root path onto each operation
  - we have to update generated API names to ensure this doesn't mess anything up
- we need to stop adding `/v1` to the configuration

This set of changes covers those changes, so that actually generating the v2 API can happen separately.


thetorpedodog:
one step in [sc-22834]

Shelnutt2 (approved):

PR #344, opened 2022-10-31
Add "don't download results" functionality to new task graphs API

thetorpedodog:
This adds the "don't download results" functionality that was omitted from the initial release of the task graphs API. It works just the same: By default only terminal nodes (or nodes which have local dependents) are downloaded, but the user can override it.

### DIFF
--- a/tests/taskgraphs/test_builder.py
+++ b/tests/taskgraphs/test_builder.py
@@ -231,6 +231,7 @@ class TestBuilder(unittest.TestCase):
             format_it = grf.udf(
                 "sum of {name!r} is {sum!r}".format,
                 types.args(name=array_uri, sum=intify),
+                download_results=False,
             )
 
             # Execute an SQL query with parameterized input.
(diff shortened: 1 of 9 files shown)


Shelnutt2 (approved):

thetorpedodog:
so [this change](https://github.com/TileDB-Inc/TileDB-Cloud-Py/compare/3ed4eb9eb160d580e272f97af000731794da1482..4e80984806756c93f17833505a2e7fcc5590e93b) is from a mistake I made in node starting: because the lambda does not capture the *current value* of the `node` argument but instead a reference to the value itself, we were trying to start the same node multiple times.

This explains [the failures we were having on previous test runs](https://github.com/TileDB-Inc/TileDB-Cloud-Py/actions/runs/3364159251/jobs/5578392940). The new version should solve this since it pulls the lambda out into its own function where it is never re-bound.

thetorpedodog:
[sc-23079]

PR #345, opened 2022-10-31
Don't use multiprocessing when running nodes locally.

thetorpedodog:
Despite the potential advantages of running local nodes in multiple threads from 008cf67248b28aa15a810827e6b610e4b2959ef2 when it was added, there are difficult-to-diagnose issues with multiprocessing where it seems like certain processes are never starting. For this reason, we're dropping back to just executing everything on the same thread.

### DIFF
--- a/tests/taskgraphs/test_client_executor.py
+++ b/tests/taskgraphs/test_client_executor.py
@@ -142,14 +142,6 @@ class ClientExecutorTestLocal(unittest.TestCase):
             os.remove(p_two)
             self.assertEqual("one two", exec.node(out).result(5))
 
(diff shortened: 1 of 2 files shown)


thetorpedodog:
Adding a note for myself for later: It looks like a (the?) reason we’re sometimes running into mysterious trouble is the fact that we’re trying to serialize items of >2GB over the pipeline. Multiprocessing appears to only support things up to 2 GB and so when serializing either way it dies, possibly in an un-diagnosable state before anything even starts running?

antalakas (approved):

PR #348, opened 2022-11-02
Implement groups API creation/deletion functions.

thetorpedodog:
Adds:
- Create new group.
- Register existing group.
- Deregister group.

### DIFF
--- a/tiledb/cloud/groups.py
+++ b/tiledb/cloud/groups.py
@@ -1,0 +1,180 @@
+"""Functions for managing TileDB Cloud groups."""
+
+import posixpath
+import urllib.parse
+from typing import Iterable, Optional, Tuple, Union
+
+import tiledb.cloud.tiledb_cloud_error as tce
+from tiledb.cloud import client
+from tiledb.cloud import rest_api
+from tiledb.cloud import utils
+from tiledb.cloud._common import api_v2
+
+
+def create(
+    name: str,
+    *,
+    namespace: Optional[str] = None,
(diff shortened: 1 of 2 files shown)


tiledb/cloud/groups.py
```diff
@@ -0,0 +1,181 @@
...
+        recursively) before deregistering the group itself.
+    """
+    namespace, name = utils.split_uri(uri)
+
+    array_api = client.build(rest_api.ArrayApi)
+    groups_api = client.build(api_v2.GroupsApi)
+    groups_api_v1 = client.build(rest_api.GroupsApi)
+    if recursive:
+        while True:
+            contents: rest_api.GroupContents = groups_api_v1.get_group_contents(
```
Shelnutt2, line 108: It might be easier to use the tiledb group function:
```
with tiledb.Group(uri, ctx=tiledb.cloud.Ctx() as G:
   for member in G:
      memory.uri
```

Shelnutt2 (changes requested):
Looks great! Only one comment/suggestion/option, but lgtm.

Shelnutt2 (approved):
Meant to hit approve

tiledb/cloud/groups.py
thetorpedodog: At this point I would prefer to use *only* direct use of the HTTP API rather than mixing the HTTP API with using the local TileDB Core APIs, but it might make sense to do this in the future.

thetorpedodog:
[sc-22835]

PR #349, opened 2022-11-03
Add global group import to __init__.py

Shelnutt2:
This should make `import tiledb.cloud` have the `tiledb.cloud.groups` available.


JohnMoutafis:
I believe that this is not relevant anymore.
`groups` is already a part of the `tiledb.cloud` imports.

PR #350, opened 2022-11-03
Don't try to derive the default groups credential client-side.

thetorpedodog:
If the server is not given a storage credential name, it will select the correct credential. We do not have to do this on the client side.


thetorpedodog:
[sc-23185]

snagles (approved):

antalakas (approved):

PR #351, opened 2022-11-07
Use the default version scheme from setuptools_scm.

thetorpedodog:
The (new?) default setuptools_scm version scheme includes the short revision hash in the version string. This is helpful for when you want to install a development version, because it will install `x.y.z.devN+somerev` rather than just `x.y.z.devN`, which would be the same if you have two separate branches that are both N revs away from the latest tag.


gsakkis (approved):
:ship: :it:

PR #353, opened 2022-11-08
Move away from deprecated GitHub stuff.

thetorpedodog:
- set-output is no longer a thing
- a bunch of actions need to be updated to run on newer Node


gsakkis (approved):
:shipit:

Issue #354, opened 2022-11-25
Check for write access


vladkasianenko:
https://github.com/TileDB-Inc/TileDB-Cloud-Py/blob/fb3004833e49bc856e8df99d5af8aa7ec43366e8/tiledb/cloud/config.py#L24

It tries to create a file, but the app might run in the container with read-only filesystem.

thetorpedodog:
Thanks for reporting this! The fix is now available on PyPI as of `tiledb-cloud` v0.8.1.

PR #355, opened 2022-11-30
Don't fail when we can't save the config file on login.

thetorpedodog:
Users running in a read-only filesystem may not be able to write the TileDB Cloud config file but this should not prevent login. Record a warning so that the user can be alerted of the situation if it's not intended, but allow execution to continue.

Fixes https://github.com/TileDB-Inc/TileDB-Cloud-Py/issues/354

### DIFF
--- a/tiledb/cloud/client.py
+++ b/tiledb/cloud/client.py
@@ -1,6 +1,7 @@
 import enum
 import os
 import threading
+import warnings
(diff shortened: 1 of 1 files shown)


thetorpedodog:
for internal tracking: [sc-23837]

Shelnutt2 (approved):

PR #362, opened 2023-02-16
adding connector for the Python DB 2.0 API

DimitrisStaratzis:
This PR introduces a connector in accordance with the Python DB API 2.0 standards. 

In my initial implementation I created a separate repository https://github.com/TileDB-Inc/TileDB-Cloud-PythonDB, but @ihnorton suggested that we should include the connector here instead. 

All the code is included in this PR but I am not sure how to package it correctly. Any suggestions are welcome @thetorpedodog

### DIFF
--- a/src/tiledb/cloud/sql/README.md
+++ b/src/tiledb/cloud/sql/README.md
@@ -1,0 +1,22 @@
+# DB API 2.0 Connector
+
+This package features the TileDB-Cloud-PythonDB connector, which aligns with the Python DB API 2.0 specification. It offers a convenient way for Python developers to connect to TileDB-Cloud and perform all necessary operations. 
+This can also be seen as an alternative to a JDBC or ODBC driver.
+
+## Usage
+
+```python
+from tiledb.cloud.sql.tiledb_connection import TileDBConnection
+
+connection = TileDBConnection()
+cursor = connection.cursor()
+cursor.execute("SELECT * from `tiledb://TileDB-Inc/quickstart_dense`")
+
+row = cursor.fetchone()
+while row:
+    print(row)
+    row = cursor.fetchone()
+
+print(cursor.description())
+
+```
--- a/src/tiledb/cloud/sql/db_api_exceptions.py
+++ b/src/tiledb/cloud/sql/db_api_exceptions.py
@@ -1,0 +1,38 @@
+class Error(Exception):
+    """Base class for exceptions in the DB API 2.0 implementation."""
+
+
+class InterfaceError(Error):
+    """Raised when the database encountered a programming error."""
+
+
+class DatabaseError(Error):
+    """Raised when the database encountered a programming error."""
+
+
+class Warning(Exception):
+    """Exception for important warnings."""
+
+
+class ProgrammingError(Error):
+    """Raised when the database encountered a programming error."""
+
+
+class DataError(Error):
+    """Raised when the database encountered data-related errors."""
+
+
+class OperationalError(Error):
+    """Raised when the database encountered operational errors."""
+
+
+class IntegrityError(Error):
+    """Raised when the database encountered an integrity constraint error."""
+
+
+class InternalError(Error):
+    """Raised when the database encountered an internal error."""
+
+
+class NotSupportedError(Error):
+    """Raised when the database encountered a feature not supported error."""
--- a/src/tiledb/cloud/sql/tiledb_connection.py
+++ b/src/tiledb/cloud/sql/tiledb_connection.py
@@ -1,0 +1,21 @@
+from tiledb.cloud.sql.db_api_exceptions import NotSupportedError
+from tiledb.cloud.sql.tiledb_cursor import Cursor
+
+
+class TileDBConnection:
+    def cursor(self):
+        return Cursor()
+
+    def commit(self):
+        # Commit must work, even if it doesn't do anything
+        # No rollback method due to no transaction support
+        pass
+
+    def close(self):
+        pass
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, type, value, traceback):
+        pass
--- a/src/tiledb/cloud/sql/tiledb_cursor.py
+++ b/src/tiledb/cloud/sql/tiledb_cursor.py
@@ -1,0 +1,118 @@
+import pyarrow as pa
+
+from tiledb.cloud.sql._execution import exec
+from tiledb.cloud.sql.db_api_exceptions import DataError
+from tiledb.cloud.sql.db_api_exceptions import ProgrammingError
+
+
+def _get_db_type(dtype):
+    if pa.types.is_timestamp(dtype):
+        return "DATETIME"
+    elif pa.types.is_boolean(dtype):
+        return "BOOLEAN"
+    elif pa.types.is_string(dtype):
+        return "STRING"
+    elif pa.types.is_integer(dtype) or pa.types.is_floating(dtype):
+        return "NUMBER"
+    return "UNKNOWN"
+
+
+class Cursor:
+    def __init__(self):
+        self._results = None
+        self._row_index = 0
+        self.arraysize = 1
+        self._description_cache = None
+
+    def executemany(self, query, seq_of_parameters):
+        for params in seq_of_parameters:
+            self.execute(query, params)
+
+    def execute(self, query, params=()):
+        try:
+            self._description_cache = None
+            self._results = exec(query=query, parameters=params, raw_results=True)
+        except Exception as e:
+            self._results = None
+            raise DataError(f"Error executing query: {e}") from e
+
+    def fetchmany(self, size=-1):
+        if not self._results:
+            raise DataError("Failed to fetch results")
+
+        if size == -1:
+            size = self.arraysize
+
+        if size + self._row_index > len(self._results):
+            # give all that is remaining
+            size = len(self._results) - self._row_index
+
+        if size == 0:
+            # There are no more records
+            return []
+
+        rows = self._results[self._row_index : self._row_index + size]
+        self._row_index += size
+        return rows.to_pylist()
+
+    def setinputsizes(self, sizes):
+        return None
+
+    def setoutputsize(self, size, column):
+        return None
+
+    @property
+    def rowcount(self):
+        return -1 if self._results is None else len(self._results)
+
+    def fetchone(self):
+        result = self.fetchmany(1)
+        if len(result) == 0:
+            return None
+        return result[0]
+
+    @property
+    def description(self):
+        if self._description_cache is not None:
+            return self._description_cache
+
+        if not self._results:
+            return None
+
+        description = [
+            (field.name, _get_db_type(field.type), None, None, None, None, None)
+            for field in self._results.schema
+        ]
+        self._description_cache = description
+        return description
+
+    def fetchall(self):
+        if self._row_index > 0:
+            return self.fetchmany(len(self._results) - self._row_index)
+        if self._results is None:
+            raise DataError("The query results are null")
+        return self._results.to_pylist()
+
+    def close(self):
+        pass
+
+    def reset(self):
+        self._row_index = 0
+        self.arraysize = 1
+
+    def scroll(self, value, mode="relative"):
+        if mode == "relative":
+            new_index = self._row_index + value
+        elif mode == "absolute":
+            new_index = value
+        else:
+            raise ProgrammingError(
+                f"Invalid mode '{mode}'. Please choose between 'relative' and 'absolute'."
+            )
+
+        if not 0 <= new_index < len(self._results) - 1:
+            raise IndexError(
+                f"Row index {new_index} is out of bounds (0 to {len(self._results) - 1})."
+            )
+
+        self._row_index = new_index
--- a/tests/db_connector/test_connector.py
+++ b/tests/db_connector/test_connector.py
@@ -1,0 +1,76 @@
+import os
+import unittest
+
+from tiledb.cloud.sql.tiledb_connection import TileDBConnection
+
+
+class TileDBConnectionTest(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        # Using setUpClass to run query only once for this class
+        con = TileDBConnection()
+        cls.cur = con.cursor()
+        cls.cur.execute(
+            "SELECT * from `tiledb://TileDB-Inc/quickstart_dense` WHERE `rows` > ?",
+            (0,),
+        )
+        cls.correct_data = [
+            {"rows": 1, "cols": 1, "a": 1},
+            {"rows": 1, "cols": 2, "a": 2},
+            {"rows": 1, "cols": 3, "a": 3},
+            {"rows": 1, "cols": 4, "a": 4},
+            {"rows": 2, "cols": 1, "a": 5},
+            {"rows": 2, "cols": 2, "a": 6},
+            {"rows": 2, "cols": 3, "a": 7},
+            {"rows": 2, "cols": 4, "a": 8},
+            {"rows": 3, "cols": 1, "a": 9},
+            {"rows": 3, "cols": 2, "a": 10},
+            {"rows": 3, "cols": 3, "a": 11},
+            {"rows": 3, "cols": 4, "a": 12},
+            {"rows": 4, "cols": 1, "a": 13},
+            {"rows": 4, "cols": 2, "a": 14},
+            {"rows": 4, "cols": 3, "a": 15},
+            {"rows": 4, "cols": 4, "a": 16},
+        ]
+
+    def test_row_count(self):
+        # reset the cursor
+        self.cur.scroll(0, "absolute")
+        self.assertEqual(16, self.cur.rowcount)
+
+    def test_fetchall(self):
+        # reset the cursor
+        self.cur.scroll(0, "absolute")
+        results = self.cur.fetchall()
+        self.assertEqual(results, self.correct_data)
+
+    def test_fetchmany(self):
+        self.cur.scroll(0, "absolute")
+        results = self.cur.fetchmany(5)
+        self.assertEqual(results, self.correct_data[:5])
+        results = self.cur.fetchmany(2)
+        self.assertEqual(results, self.correct_data[5:7])
+
+    def test_fetchone(self):
+        self.cur.scroll(1, "absolute")
+        results = self.cur.fetchone()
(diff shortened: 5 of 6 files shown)


tiledb/db_connector/__init__.py
```diff
@@ -0,0 +1,10 @@
+from tiledb.db_connector.cursor import Cursor
+from tiledb.db_connector.connection import TileDBConnection
+
+apilevel = "2.0"
+threadsafety = 2
+paramstyle = "pyformat"
```
thetorpedodog, line 6: I believe this should be `qmark`, since that is what MariaDB supports.

tiledb/cloud/sql/tiledb_connection.py
```diff
@@ -0,0 +1,25 @@
+import tiledb.cloud
+from tiledb.cloud.sql.tiledb_cursor import Cursor
+from tiledb.cloud.sql.tiledb_error import NotSupportedError
+
+
+class TileDBConnection:
+    def __init__(self, token):
+        self.token = token
+        tiledb.cloud.login(token=self.token)
```
Shelnutt2, line 9: Lets support all the parameters of `tiledb.cloud.login`.

Also I think we should login only for the python process and not write to the home directory. @thetorpedodog what do you think? I'm thinking the user might not expect passing parameters to the database contructor to effect their global login. I think we'll need a new option for the login function.

tiledb/cloud/sql/tiledb_cursor.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def setoutputsize(self, sizes):
+        raise NotSupportedError("Operation not supported")
+
+    def fetchone(self):
+        return self.fetchmany(1)
+
+    def description(self):
+        if self.description_cache is not None:
+            return self.description_cache
```
thetorpedodog, line 65: `description_cache` should probably be reset on each `execute` call, otherwise it will never be recalculated.

tiledb/cloud/sql/tiledb_cursor.py
```diff
@@ -0,0 +1,108 @@
...
+            self.row_index = value
+        else:
+            raise ProgrammingError(
+                f"Invalid mode '{mode}'. Please choose between 'relative' and 'absolute'."
+            )
+
+        if self.row_index < 0 or self.row_index > len(self.results) - 1:
+            raise IndexError(
+                f"Row index {self.row_index} is out of bounds (0 to {len(self.results) - 1})."
+            )
```
thetorpedodog, line 108: This should be put first in the function, otherwise you end up leaving the index at an impossible value.

Also, you can shorten this with chained comparisons:

```python
if not 0 <= self.row_index < len(self.results) - 1:
```

tiledb/cloud/sql/util.py
```diff
@@ -0,0 +1,15 @@
+# This file includes the datatype mapping from pandas to the types required by the Python DB API 2.0
+import pyarrow as pa
+
+
+def getDBType(dtype):
```
thetorpedodog, line 5: This can also probably be put in the same file as the two classes, and is probably better named `_get_db_type`.

thetorpedodog (commented):
Since the submission of #365, the root directory is now `src`, so you’ll want to rebase and update the location of your files to match.

tiledb/cloud/sql/tiledb_connection.py
thetorpedodog: I didn’t see this comment before—my thought was that we could [assume the user was already logged in](https://github.com/TileDB-Inc/TileDB-Cloud-Py/pull/362#discussion_r1116095551). If we want to do this separately, that’s going to get somewhat hairy since we would need to keep track of a separate client object, API functions, etc. This isn’t necessarily a bad idea, but I think would be a good idea for a follow-up to keep the scope of this change smaller.

Shelnutt2: I'm flexible here, I just don't love the login call. If for v1 we want to force the user to call login outside I'm good with that.

DimitrisStaratzis: The login call was a leftover from when this was a separate package. I think the best way is to have the user call login

tests/db_connector/test_connector.py
```diff
@@ -0,0 +1,71 @@
...
+import unittest
+
+from tiledb.cloud.sql.tiledb_connection import TileDBConnection
+
+
+class TileDBConnectionTest(unittest.TestCase):
+    cur = None
+
+    @classmethod
+    def setUpClass(cls):
```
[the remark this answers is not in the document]
DimitrisStaratzis: My aim is to run the query only once. If I use setUp I believe it will be called before the execution of each test

tests/db_connector/test_connector.py
```diff
@@ -0,0 +1,71 @@
...
+        cls.tkn = os.environ.get("TILEDB_REST_TOKEN")
```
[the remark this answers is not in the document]
DimitrisStaratzis: The user will have to call login() outside of this connector.

tiledb/cloud/sql/tiledb_cursor.py
```diff
@@ -0,0 +1,108 @@
+import tiledb.cloud
+from tiledb.cloud.sql.tiledb_error import DataError
+from tiledb.cloud.sql.tiledb_error import NotSupportedError
+from tiledb.cloud.sql.tiledb_error import ProgrammingError
+from tiledb.cloud.sql.util import getDBType
+
+
+class Cursor:
```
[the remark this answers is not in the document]
DimitrisStaratzis: This is the only change I haven't implemented, Personally, I prefer having the classes in separate files and have a cleaner structure

tiledb/cloud/sql/tiledb_cursor.py
```diff
@@ -0,0 +1,108 @@
...
+    def __init__(self):
+        self.results = None
+        self.row_index = 0
+        self.inputarraysize = 1
+        self.description_cache = None
```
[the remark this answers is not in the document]
DimitrisStaratzis: Good catch thank you! I mis-read the DB API document

DimitrisStaratzis:
I have applied/responded to all comments above and rebased against main.

thetorpedodog (commented):
This is getting pretty close; I think there are some additional things that will need to be *added* (the various datetime constructors, a few other things) after it’s done. However, in order to keep the size of this changeset manageable I would like to just make sure that what’s being added here is correct, submit that, and then go on and add the rest of the missing parts.

src/tiledb/cloud/sql/tiledb_cursor.py
```diff
@@ -0,0 +1,117 @@
...
+        return rows.to_pylist()
+
+    def nextset(self):
+        raise NotSupportedError("Operation not supported")
+
+    def setinputsizes(self, sizes):
+        raise NotSupportedError("Operation not supported")
+
+    def setoutputsize(self, sizes):
+        raise NotSupportedError("Operation not supported")
```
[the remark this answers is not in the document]
DimitrisStaratzis: I thought the user should now that the operation is not supported so I left it as it is. Let me know what you think.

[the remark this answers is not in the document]
thetorpedodog: Since these are both essentially hints to the engine rather than operations that affect the semantics of querying, I think ignoring them is the right thing to do:

> Implementations are free to have this method do nothing and users are free to not use it.

tests/db_connector/test_connector.py
```diff
@@ -0,0 +1,68 @@
...
+        self.cur.scroll(0, "absolute")
+        results = self.cur.fetchmany(5)
+        self.assertEqual(results, self.correctData[:5])
+        results = self.cur.fetchone()
+        self.assertEqual(results, self.correctData[5])
+        results = self.cur.fetchall()
+        self.assertEqual(results, self.correctData[6:])
+
+    def test_description(self):
+        correctDescription = [
```
thetorpedodog, line 61: nit: `correct_description`

src/tiledb/cloud/sql/db_api_exceptions.py
```diff
@@ -0,0 +1,38 @@
...
+class InterfaceError(Error):
+    """Raised when the database encountered a programming error."""
+
+
+class DatabaseError(Error):
+    """Raised when the database encountered a programming error."""
+
+
+class _Warning(Exception):
+    """Exception for important warnings."""
```
thetorpedodog, line 14: While this will shadow the built-in `Warning` type, the DB API specifies that it should be named that way, so just `class Warning` here.

src/tiledb/cloud/sql/tiledb_cursor.py
```diff
@@ -0,0 +1,117 @@
+import pyarrow as pa
+
+from tiledb.cloud.sql._execution import exec
+from tiledb.cloud.sql.db_api_exceptions import DataError
+from tiledb.cloud.sql.db_api_exceptions import NotSupportedError
+from tiledb.cloud.sql.db_api_exceptions import ProgrammingError
+
+
+def _getDBType(dtype):
```
thetorpedodog, line 9: `_get_db_type`

src/tiledb/cloud/sql/tiledb_cursor.py
```diff
@@ -0,0 +1,117 @@
...
+    def __init__(self):
+        self._results = None
+        self._row_index = 0
+        self.arraysize = 1
+        self._description_cache = None
+
+    def executemany(self, query):
+        self.execute(query)
+
+    def execute(self, query):
```
thetorpedodog, line 31: This needs to accept a `parameters` optional parameter. The default value should probably be `()` (i.e., the empty tuple).

Likewise, `executemany` takes a sequence of `parameters`es, and it might look something like:

```python
def executemany(self, query, seq_of_parameters):
  for params in seq_of_parameters:
    self.execute(query, params)

File: src/tiledb/cloud/sql/tiledb_cursor.py [hunk 15]
```diff
@@ -0,0 +1,117 @@
...
+        try:
+            self._results = exec(query=query, raw_results=True)
+            self._description_cache = None
```
thetorpedodog, line 34: This probably needs to go above the `self._results` call.

src/tiledb/cloud/sql/tiledb_cursor.py
```diff
@@ -0,0 +1,117 @@
...
+        except Exception as e:
+            self._results = None
+            raise DataError(f"Error executing query: {e}") from e
+
+    def fetchmany(self, size=-1):
+        if self._results is None or len(self._results) == 0:
```
thetorpedodog, line 40: These can be simplified to `if not self._results`; sequences of length 0 are considered falsey (here specifically, and also mostly in general).

src/tiledb/cloud/sql/tiledb_cursor.py
```diff
@@ -0,0 +1,117 @@
...
+            return None
+
+        description = [
+            (field.name, _getDBType(field.type)) for field in self._results.schema
+        ]
+        self._description_cache = description
+        return description
+
+    def rowcount(self):
```
thetorpedodog, line 83: This should be an `@property`.

src/tiledb/cloud/sql/tiledb_cursor.py
```diff
@@ -0,0 +1,117 @@
...
+        if self._results is None or len(self._results) == 0:
```
thetorpedodog, line 84: I think this should be 0 when there are specifically 0 results (as opposed to the `_results is None` case, which should still be −1).

thetorpedodog (commented):
oh my goodness, this completely dropped off my radar. sorry for the delay.

src/tiledb/cloud/sql/tiledb_cursor.py
```diff
@@ -0,0 +1,116 @@
...
+    def setoutputsize(self, size, column):
+        raise NotSupportedError("Operation not supported")
+
+    def fetchone(self):
+        result = self.fetchmany(1)
+        if len(result) == 0:
+            return None
+        return result[0]
+
+    def description(self):
```
thetorpedodog, line 73: This needs to be an `@property`.

src/tiledb/cloud/sql/tiledb_cursor.py
```diff
@@ -0,0 +1,116 @@
...
+
+        rows = self._results[self._row_index : self._row_index + size]
+        self._row_index += size
+        return rows.to_pylist()
+
+    def setinputsizes(self, sizes):
+        raise NotSupportedError("Operation not supported")
+
+    def setoutputsize(self, size, column):
+        raise NotSupportedError("Operation not supported")
```
thetorpedodog, line 65: Bumping making these do nothing rather than return, for the reasons covered in [this comment, which might have gotten lost](https://github.com/TileDB-Inc/TileDB-Cloud-Py/pull/362#discussion_r1132904022)

src/tiledb/cloud/sql/README.md
```diff
@@ -0,0 +1,22 @@
+# DB API 2.0 Connector
+
+This package features the TileDB-Cloud-PythonDB connector, which aligns with the Python DB API 2.0 specification. It offers a convenient way for Python developers to connect to TileDB-Cloud and perform all necessary operations. 
+This can also be seen as an alternative to a JDBC or ODBC driver.
+
+## Usage
+
+```python
+from src.tiledb.cloud.sql.tiledb_connection import TileDBConnection
```
thetorpedodog, line 9:
```suggestion
from tiledb.cloud.sql.tiledb_connection import TileDBConnection
```

src/tiledb/cloud/sql/tiledb_connection.py
```diff
@@ -0,0 +1,23 @@
+from tiledb.cloud.sql.db_api_exceptions import NotSupportedError
+from tiledb.cloud.sql.tiledb_cursor import Cursor
+
+
+class TileDBConnection:
+    def cursor(self):
+        return Cursor()
+
+    def rollback(self):
+        raise NotSupportedError("Operation not supported")
```
thetorpedodog, line 10: per the PEP](https://peps.python.org/pep-0249/#id51) it is recommended to [have this not be present](https://peps.python.org/pep-0249/#id51). (It might be worth adding a comment under `commit` to the effect of “no `rollback` method due to no transaction support”

src/tiledb/cloud/sql/tiledb_cursor.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def description(self):
+        if self._description_cache is not None:
+            return self._description_cache
+
+        if not self._results:
+            return None
+
+        description = [
+            (field.name, _get_db_type(field.type)) for field in self._results.schema
```
thetorpedodog, line 81: This needs to be a 7-tuple; the rest of the entries will all be None:

```python
(field.name, _get_db_type(field.type), None, None, None, None, None)
```

(yeah this looks a bit ridiculous)

src/tiledb/cloud/sql/tiledb_cursor.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def executemany(self, query, seq_of_parameters):
+        for params in seq_of_parameters:
+            self.execute(query, params)
+
+    def execute(self, query, params=()):
+        try:
+            self._description_cache = None
+            self._results = exec(query=query, raw_results=True)
+            self.rowcount = len(self._results)
```
thetorpedodog, line 37: To make sure `rowcount` and `_results` never get out of sync, it might be best to make `rowcount` a `@property`:

```python
@property
def rowcount(self):
    return -1 if self._results is None else len(self._results)
```

src/tiledb/cloud/sql/tiledb_cursor.py
```diff
@@ -0,0 +1,116 @@
...
+        self._description_cache = None
+
+    def executemany(self, query, seq_of_parameters):
+        for params in seq_of_parameters:
+            self.execute(query, params)
+
+    def execute(self, query, params=()):
+        try:
+            self._description_cache = None
+            self._results = exec(query=query, raw_results=True)
```
thetorpedodog, line 36: the parameters should be passed into this `exec`

src/tiledb/cloud/sql/tiledb_cursor.py
```diff
@@ -0,0 +1,116 @@
...
+            self.rowcount = len(self._results)
+        except Exception as e:
+            self._results = None
+            raise DataError(f"Error executing query: {e}") from e
+
+    def fetchmany(self, size=-1):
+        if not self._results:
+            return None
```
thetorpedodog, line 44: Looking at the documentation for `fetchmany`, we should handle the case of 0 results (`self._results` is empty) vs. unset results (`self._results` is None) differently; in the None case this should raise.

tests/db_connector/test_connector.py
```diff
@@ -0,0 +1,73 @@
...
+from tiledb.cloud.sql.tiledb_connection import TileDBConnection
+
+
+class TileDBConnectionTest(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        # Using setUpClass to run query only once for this class
+        con = TileDBConnection()
+        cls.cur = con.cursor()
+        cls.cur.execute("SELECT * from `tiledb://TileDB-Inc/quickstart_dense`")
```
thetorpedodog, line 13: One thing before submission: Could you add a parameter substitution to this test? That way we can verify that that part works. Maybe like `where rows > ?` and provide 2, or any other formulation you prefer.

thetorpedodog (approved):
Looking good!

tests/db_connector/test_connector.py
DimitrisStaratzis: Could you explain this a bit more? You mean using it like this:
``` cls.cur.execute("SELECT * from `tiledb://TileDB-Inc/quickstart_dense` WHERE rows > ?", (2, 3, 4))```

and essentially run the tests as many times as the values in the list for the placeholder? And in each test adapt the `results` variable according to this value?

thetorpedodog: Sorry I didn’t get to this before I was off for a week! I would say you really only need to include 1 parameter here (e.g. `execute("select * from uri where rows > ?", (0,))`). That way we can be sure that substituting in parameters works correctly.

DimitrisStaratzis: tests passed!

thetorpedodog:
This all looks good. Squash and merge at your leisure!

PR #363, opened 2023-02-17
Warn users when not logged in

Shelnutt2:
Also return proper boolean for logged in status on startup.

I'm not sure if the warning is best or if we should introduce a logger here.

### DIFF
--- a/tiledb/cloud/config.py
+++ b/tiledb/cloud/config.py
@@ -1,5 +1,6 @@
 import json
 import os.path
+import warnings
 from pathlib import Path
 from typing import Optional
 
@@ -46,6 +47,7 @@ def save_configuration(config_file):
 
 
 def load_configuration(config_path):
+    logged_in = True
     # Look for env variables
     token = os.getenv("TILEDB_REST_TOKEN", None)
     if token is not None and token != "":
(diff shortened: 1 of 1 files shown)


tiledb/cloud/config.py
```diff
@@ -91,7 +93,10 @@ def load_configuration(config_path):
                 verify_ssl = config_obj["verify_ssl"]
 
     if (token is None or token == "") and (username is None or username == ""):
-        return "You must first login before you can run commands"
+        warnings.warn(
+            "You must first login before you can run commands. Please run tiledb.cloud.login."
+        )
+        logged_in = False
```
thetorpedodog, line 99: Any reason not to `return False` here? (in which case we can also lose the `logged_in` variable and `return True` at the end of the function)

thetorpedodog (commented):
A good change! One question but otherwise good.

tiledb/cloud/config.py
Shelnutt2: Well some public assets and functions work without being logged in, but we've got to configure the host for it to work. There is probably a cleaner way to do this, or a better message to have but I didn't want the warning to get too long.

`tiledb.cloud.info("tiledb://TileDB-Inc/quickstart_sparse")`

thetorpedodog (approved):

PR #364, opened 2023-02-17
Move sql implementation into a directory.

thetorpedodog:
This prepares for the inclusion of the TileDB Cloud Python Database API connector.  All names are exported and available from the same place, so there is no impact on user code.


Shelnutt2 (approved):
Nice change!

DimitrisStaratzis (approved):

PR #365, opened 2023-02-18
Switch to `pyproject.toml` and use src-layout.

thetorpedodog:
By using the src-layout convention and `exclude` functionality*, we can eliminate our custom `setup.py` logic entirely and switch to using a purely-declarative `pyproject.toml` file.

*: https://setuptools.pypa.io/en/latest/userguide/package_discovery.html


nguyenv (approved):

Shelnutt2:
Does this handle if we want other namespaces besides the top level `tiledb`?

thetorpedodog:
> Does this handle if we want other namespaces besides the top level `tiledb`?

It does, yes.

pyproject.toml
```diff
@@ -1,5 +1,41 @@
...
+
 [build-system]
 requires = ["setuptools>=42", "wheel", "setuptools_scm>=6"]
 
+[tool.setuptools]
+zip-safe = false
+
+[tool.setuptools.packages.find]
+where = ["src"]
+exclude = ["tiledb"]
```
ihnorton, line 38: Why the exclude here?

thetorpedodog: I could have sworn this was necessary to make it not clobber `tiledb/__init__.py`, but apparently not? No complaints about removing it from me.

ihnorton (approved):

PR #366, opened 2023-02-20
Fix bug for passing kwargs to Batch UDF

NikolaosPapailiou:


tiledb/cloud/dag/dag.py
```diff
@@ -1404,7 +1404,9 @@ def _build_batch_taskgraph(self):
                     args.append(models.TGUDFArgument(value=esc.visit(arg)))
 
             for name, arg in node.kwargs.items():
-                if isinstance(arg, Node):
+                if name in ["image_name", "timeout", "result_format"]:
```
Shelnutt2, line 1407: Should this be a global defined list?

NikolaosPapailiou: This list is only used here but we could add a global definition. We could also pop the kwargs in the Node constructor but I didn't know if this is going to result in any side effects. What should I prefer?

Shelnutt2 (approved):

PR #367, opened 2023-02-21
Initial implementation of dynamic batch taskgraphs

NikolaosPapailiou:


tiledb/cloud/dag/dag.py
```diff
@@ -895,6 +899,33 @@ def submit_udf(self, func, *args, **kwargs):
...
     submit = submit_udf
 
+    def submit_udf_stage(
+        self, func, *args, expand_node_output: Optional[Node] = None, **kwargs
+    ):
+        """
+        Submit a function that will be executed in the cloud serverlessly
+        :param func: function to execute
+        :param args: arguments for function execution
+        :param expand_node_output: the Node that we want to expand the output of. The output of the node should be a
```
Shelnutt2, line 909: For `expand_node_output`, can we support multiple parameters being passed or complex structures? Does it have to be JSON encoded? Can we use arrow or native serialization?

NikolaosPapailiou: I think that Argo can only expand one node output. In this version, we don't have support for passing more complex structures (i.e. pickled values). However, I think that this is achievable by adding a custom serialization to the UDF images in order to have an output of a JSON list of TGArgs

tests/test_dag.py
```diff
@@ -623,6 +623,52 @@ def test_batch_dag_failure(self):
...
+        split = d.submit(
+            generate_split,
+            name="split",
+            resources={"cpu": "1", "memory": "500Mi"},
+            result_format=models.ResultFormat.JSON,
+        )
+        mult_1 = d.submit_udf_stage(
+            lambda x: x * 2,
+            split,
+            expand_node_output=split,
```
Shelnutt2, line 645: Is there a case where we'd not want the `expand_node_output` and the `split` input args to be the same? I.e could we infer the `expand_node_output` or do we always need to pass the node parameter twice?

snagles (approved):

PR #371, opened 2023-02-24
Distributed taskgraph ingestion bioimg files

ktsitsi:
This PR:

Exposes a function `ingest` under `tiledb.cloud.bioimg.ingestion` for the creation of a taskgraph to ingest microscopy images.

Usage:
```
from tiledb.cloud.bioimg.ingestion import ingest
conf = tiledb.Config()
conf['vfs.s3.access_key_id'] = ...
conf['vfs.s3.secret_access_key'] = ...
ingest('s3://tiledb-isaiah2/jjdemo/test3/orig/C3L-00445-21.svs', 's3://tiledb-bioimg/datasets/TCIA-23/cloud_test/', conf)
```


src/tiledb/cloud/bioimg/ingestion.py
```diff
@@ -0,0 +1,119 @@
...
+from functools import partial
+from typing import Any, Mapping, Optional, Sequence, Tuple, Union
+
+import tiledb
+from tiledb.cloud.utils import logger
+
+
+def ingest(source: Union[Sequence[str], str],
+           output: str, 
+           access_key_id: str,
```
Shelnutt2, line 12: Lets support roles here too, perhaps we should generically take a dictionary of configs to pass to tiledb's vcf?

src/tiledb/cloud/bioimg/ingestion.py
```diff
@@ -0,0 +1,119 @@
...
+           **kwargs):
+    
+    """The function ingests microscopy images into TileDB arrays
+
+    :param source: uri / iterable of uris of input files e.g 
+    :param output: output dir for the ingested tiledb arrays
+    :param access_key_id: AWS_ACCESS_KEY_ID
+    :param secret_access_key: AWS_SECRET_ACCESS_KEY
+    :param taskgraph_name: Optional name for taskgraph, defaults to None
+    :param num_batches: Number of graph nodes to spawn performs it sequentially if default, defaults to 1
```
Shelnutt2, line 28: I think we want this to be higher than one? With one doesn't this mean a single node will sequentially process all images? We should use some internal heuristics, likely building one task per image and then limit the concurrency?

src/tiledb/cloud/bioimg/ingestion.py
```diff
@@ -0,0 +1,119 @@
+import math
+import os
+from functools import partial
+from typing import Any, Mapping, Optional, Sequence, Tuple, Union
+
+import tiledb
+from tiledb.cloud.utils import logger
+
+
+def ingest(source: Union[Sequence[str], str],
```
Shelnutt2, line 10: Can we support `source` being an s3 path being also a s3 path? For the user to give a path and say "ingest everything in it"?

ktsitsi: > Can we support source being an s3 path being also a s3 path

You mean an `s3` bucket?

src/tiledb/cloud/bioimg/ingestion.py
ktsitsi: Yeap a `config` dict would be better.

src/tiledb/cloud/bioimg/ingestion.py
ktsitsi: Sure I had it in mind too, but I followed the simple and straightforward path. The change is easy. However I do not know if we also would like to spawn a taskgraph of 1000 nodes for a dataset of 1000 images.

src/tiledb/cloud/bioimg/ingestion.py
Shelnutt2: No I mean for the user to give a "folder", a "single file" or a "list of files". To support all 3 permutations.

src/tiledb/cloud/bioimg/ingestion.py
Shelnutt2: There is support for limiting the concurrency of a task graph, so you could launch 1000 independent nodes but say "only run 20 at a time". @NikolaosPapailiou can provide the configuration for that.

NikolaosPapailiou: Yes you can set `max_workers` in the Dag object to limit parallelism.

src/tiledb/cloud/bioimg/ingestion.py
```diff
@@ -0,0 +1,133 @@
...
+    vfs = tiledb.VFS(config=config)
+
+    def create_output_path(input_file, output_dir):
+        return os.path.join(output_dir, os.path.basename(input_file) + ".tdb")
+
+    def iter_paths(sequence):
+        result = []
+        output_uris = []
+        for uri in sequence:
+            result.append(uri)
```
ihnorton, line 124: You could rewrite this as `result.append(uri, create_output_path...)` and avoid the `zip` below.

Shelnutt2 (approved):

PR #376, opened 2023-03-10
Monkey-patch old Pandas to read newer Pandas pickles.

thetorpedodog:
This extracts the necessary parts of the pickle-reading code from Pandas v1.5.3 and monkey-patches older Pandas versions to be able to read pickles produced by the newer Pandas versions.  This enables forwards compatibility of deserialization and will eventually eliminate the need to hold Pandas back to v1.2.5.

[sc-8724]


thetorpedodog:
Forgot to add: the thing that makes this work is the fact that *new Pandas can already read old Pandas pickles*, so v1.5.3 understands v1.2.5 pickles even without this change.

Shelnutt2 (approved):
Amazing you got this working so quicklu!

PR #377, opened 2023-03-14
Sa/sc 26280/unpin pandas

anastasop:
This is stage 2 of https://github.com/TileDB-Inc/TileDB-Cloud-REST/pull/3471

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,7 @@ dependencies = [
   "pyarrow>=3.0.0",
   "python-dateutil",
   "six>=1.10",
-  "tiledb>=0.5.0",
+  "tiledb<0.21",
   "urllib3>=1.26",
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,8 +9,7 @@ dependencies = [
   "certifi",
   "cloudpickle==1.4.1",
   # as of 2021-07-06, the runtime (pinned to 1.1) is incompatible with 1.3
-  "pandas<1.3; python_version < '3.10'",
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -9,12 +9,11 @@ dependencies = [
...
   "cloudpickle==1.4.1",
   # as of 2021-07-06, the runtime (pinned to 1.1) is incompatible with 1.3
-  "pandas<1.3; python_version < '3.10'",
-  "pandas>=1.3; python_version >= '3.10'",
+  "pandas",
   "pyarrow>=3.0.0",
   "python-dateutil",
   "six>=1.10",
-  "tiledb>=0.5.0",
+  "tiledb==0.20.0",
```
thetorpedodog, line 16: Why this pin?

thetorpedodog: Makes sense. This looks good overall; just waiting for the server release to go out so we can re-run the tests and verify everything.

anastasop: I will change this to `<0.21` just in case someone uses the release while we updating notebooks and udfs

thetorpedodog (approved):
Everything looks good! Squash and merge at will.

pyproject.toml
Shelnutt2: I don't think we want to pin like this. It's going to force new version of TileDB cloud Py just to update the pin for every TileDB core release. We've been fine with an upperbound pin before this PR. I don't think we should introduce it now.

thetorpedodog: This is a temporary measure; we’re going to remove it shortly.

PR #380, opened 2023-03-22
Re-plumb `resource_class` parameter through old task graph API.

thetorpedodog:
This was popped from the `kwargs` in the initial implementation of batch task graphs, but never restored into the kwargs sent to the relevant function.


Shelnutt2 (approved):

NikolaosPapailiou (approved):
Sorry for the test failures, they are not related to this change. Will send a PR to fix them.

PR #382, opened 2023-03-27
VCF batch ingestion

gspowley:
Add `tiledb.cloud.vcf.ingest` to support ingestion of VCF files.

Example:
```python
from tiledb.cloud.vcf import ingest

dataset_uri = "s3://..."
config = {}
conf['vfs.s3.access_key_id'] = ...
conf['vfs.s3.secret_access_key'] = ...
search_uri = "s3://1000genomes-dragen-v3.7.6/data/individuals/hg38-graph-based"
pattern = "*.hard-filtered.gvcf.gz"

# limit ingestion size for testing
max_files=20

ingest(
    dataset_uri,
    config=config,
    search_uri=search_uri,
    pattern=pattern,
    max_files=max_files,
)

```


src/tiledb/cloud/vcf/ingestion.py
```diff
@@ -0,0 +1,1107 @@
...
+
+    cp = configparser.ConfigParser()
+    cp.read(os.path.expanduser(path))
+    try:
+        config["vfs.s3.aws_access_key_id"] = cp["default"]["aws_access_key_id"]
+        config["vfs.s3.aws_secret_access_key"] = cp["default"]["aws_secret_access_key"]
+    except KeyError:
+        pass
+
+    return config
```
ktsitsi, line 66: These `utils` can be used in other ingestion functions in the future, but also in some currently implemented e.g. `bioimg`. Could we move it under a `utils` folder for common use?

gspowley: Yes, it's a good idea to share the utility functions and the event logger. I can move them to `tiledb.cloud.utils` or somewhere else, if there's a better place.

src/tiledb/cloud/vcf/ingestion.py
```diff
@@ -0,0 +1,987 @@
...
+    :param max_files: maximum number of URIs to ingest, defaults to None
+    :param batch_size: manifest batch size, defaults to MANIFEST_BATCH_SIZE
+    :param workers: maximum number of parallel workers, defaults to MANIFEST_WORKERS
+    :param extra_attrs: INFO/FORMAT fields to materialize, defaults to None
+    :param vcf_attrs: VCF with all INFO/FORMAT fields to materialize, defaults to None
+    """
+
+    logger = get_logger()
+
+    graph = tiledb.cloud.dag.DAG(
```
Shelnutt2, line 657: This is where I think we should be using a dynamic dag instead of running three separate dags. @NikolaosPapailiou Can you work with @gspowley and show this could be switched to a dynamic data for determining the files to ingest.

NikolaosPapailiou: Sure I will be happy to help with this. The main idea would be to have only one DAG that does all the processing (generate_uris, update_manifest, ingest). Dynamic taskgraphs can help with dynamically creating tasks based on the output of other task. So in this example generate_uris could return a list of sample uris that update_manifest and ingest tasks expand upon creating one task per sample uri. This can be done by just returning a list object and adding `expand_output` on the dynamic DAG node (you can check an example in https://github.com/TileDB-Inc/TileDB-Cloud-Py/blob/main/tests/test_dag.py#L643).

Please keep in mind that the `expand_output` functionality is only available for bath taskgraphs. So if you want the same logic to run with realtime taskgraphs the code needs to be duplicated and follow another pattern for local or realtime execution.

src/tiledb/cloud/vcf/ingestion.py
gspowley: Reusable functions are now in `tiledb.cloud.utilities` since `utils.py` already existed.

gspowley:
VCF ingestion is ready for review and release as an experimental feature while we continue to make improvements.

src/tiledb/cloud/utilities/common.py
```diff
@@ -0,0 +1,196 @@
...
+import logging
+import os
+import sys
+from typing import Any, Mapping
+
+import tiledb
+import tiledb.cloud
+
+
+def read_config(
```
Shelnutt2, line 11: Can we rename this to something like `read_aws_config`?

src/tiledb/cloud/utilities/common.py
```diff
@@ -0,0 +1,196 @@
...
+    for key in keys:
+        try:
+            config["vfs.s3." + key] = cp[section][key]
+        except KeyError:
+            pass
+
+    return config
+
+
+def set_context(config: Mapping[str, Any] = {}):
```
Shelnutt2, line 45: Lets rename this to `set_aws_context`, or support more cloud vendors. Is there a reason this is setting os envs?

Shelnutt2 (approved):
Some minor nits, we need to get some follows tracking better handling for the configs. This can be all done in follow ups.

src/tiledb/cloud/utilities/common.py
gspowley: We can rename this now and support more cloud vendors in the future. It's setting os envs for commands that access S3 directly, like AWS SDK CLI and bcftools.

src/tiledb/cloud/vcf/ingestion.py
thetorpedodog: > Reusable functions are now in `tiledb.cloud.utilities` since `utils.py` already existed.

`utils` is really for internal consumption only; it could be moved into the `_common` directory.

src/tiledb/cloud/utilities/common.py
```diff
@@ -0,0 +1,197 @@
...
+import logging
+import os
+import sys
+from typing import Any, Mapping
+
+import tiledb
+import tiledb.cloud
+
+
+def read_aws_config(
```
thetorpedodog, line 11: Does this need to be exported for general use? It seems like it is mostly useful as an internal tool for other functions.

It's fine if we want to; I just want to make sure we keep a sensible API surface.

src/tiledb/cloud/utilities/common.py
```diff
@@ -0,0 +1,197 @@
...
+    for key in keys:
+        try:
+            config["vfs.s3." + key] = cp[section][key]
+        except KeyError:
+            pass
+
+    return config
+
+
+def set_aws_context(config: Mapping[str, Any] = {}):
```
thetorpedodog, line 45: While it looks like `config` is never mutated, I'm still uneasy with mutable objects as defaults. Maybe just `None`?

Also, add `-> None` to these functions which return nothing.

src/tiledb/cloud/vcf/ingestion.py
```diff
@@ -0,0 +1,977 @@
...
+
+import tiledb.cloud
+from tiledb.cloud.utilities import Profiler
+from tiledb.cloud.utilities import create_log_array
+from tiledb.cloud.utilities import get_logger
+from tiledb.cloud.utilities import run_dag
+from tiledb.cloud.utilities import set_aws_context
+from tiledb.cloud.utilities import write_log_event
+
+LOCAL_INGEST = False
```
thetorpedodog, line 20: Since these aren't really constants they probably shouldn't be uppercase.

src/tiledb/cloud/vcf/ingestion.py
```diff
@@ -0,0 +1,977 @@
...
+VERBOSE = False
+TRACE = False
+
+
+def local_ingest():
```
thetorpedodog, line 25: I don't entirely understand the intent of these functions. Could a user not simply set the module-level variable themselves?

If you want to keep them, it would be helpful to give them a name that reflects that they are taking an action, maybe `set_local_ingest`.

src/tiledb/cloud/vcf/ingestion.py
```diff
@@ -0,0 +1,977 @@
...
+
+
+class Contigs(enum.Enum):
+    ALL = enum.auto()
+    CHROMOSOMES = enum.auto()
+    OTHER = enum.auto()
+
+
+def setup(
+    config: Union[Mapping[str, Any], None],
```
thetorpedodog, line 67: Optional is exactly equivalent to Union[..., None].

src/tiledb/cloud/vcf/ingestion.py
```diff
@@ -0,0 +1,977 @@
...
+    dataset_uri: Optional[str] = None,
+    id: Optional[str] = None,
+) -> Union[logging.Logger, Tuple[logging.Logger, Profiler]]:
```
thetorpedodog, line 70: It's not obvious from the call under what circumstances the function will return a logger versus a logger AND a profiler.

To avoid a potential pitfall for users when using this function, I would instead make the return type one shape:

`-> Tuple[logging.Logger, Optional[Profiler]:`.

src/tiledb/cloud/vcf/ingestion.py
```diff
@@ -0,0 +1,977 @@
...
+    global TRACE
+    TRACE = True
+
+
+# Array names
+LOG_ARRAY = "log"
+MANIFEST_ARRAY = "manifest"
+
+# Default attributes to materialize
+DEFAULT_ATTRIBUTES = ["fmt_GT", "fmt_DP", "fmt_GQ"]
```
thetorpedodog, line 45: This should probably be a tuple.

src/tiledb/cloud/vcf/ingestion.py
```diff
@@ -0,0 +1,977 @@
...
+    d0 = tiledb.Dim(name="sample_name", dtype="ascii", filters=ascii_fl)
+    dom = tiledb.Domain(d0)
+
+    attrs = []
+    attrs += [tiledb.Attr(name="status", dtype="ascii", filters=ascii_fl)]
+    attrs += [tiledb.Attr(name="vcf_uri", dtype="ascii", filters=ascii_fl)]
+    attrs += [tiledb.Attr(name="vcf_bytes", dtype=np.uint64, filters=int_fl)]
+    attrs += [tiledb.Attr(name="index_uri", dtype="ascii", filters=ascii_fl)]
+    attrs += [tiledb.Attr(name="index_bytes", dtype=np.uint64, filters=int_fl)]
+    attrs += [tiledb.Attr(name="records", dtype=np.uint64, filters=int_fl)]
```
thetorpedodog, line 128:
```python
attrs = [
    tiledb.Attr(...),
    ...,
]
```
?

src/tiledb/cloud/utilities/profiler.py
```diff
@@ -0,0 +1,194 @@
...
+        """
+
+        self.uri = uri
+        self.id = id
+        self.trace = trace
+        self.array = tiledb.open(uri, "w")
+
+        # Log node start time and system info
+        node_id = open("/proc/sys/kernel/random/boot_id").read().strip()
+        cmd = "uptime && free -h && uname -a && lscpu"
```
thetorpedodog, line 104: Would it be better to run these individually so that (a) we're not using `shell=True` and (b) we get the data back in a more structured manner?

src/tiledb/cloud/utilities/profiler.py
```diff
@@ -0,0 +1,194 @@
...
+        """
+        Run a command and return the output.
+
+        :param cmd: command to run
+        :return: stdout from the command
+        """
+
+        res = subprocess.run(cmd, shell=True, capture_output=True, text=True)
+        if res.stderr:
+            print(res.stderr)
```
thetorpedodog, line 126: `stdout=subprocess.PIPE, stderr=subprocess.STDERR`?

src/tiledb/cloud/utilities/profiler.py
```diff
@@ -0,0 +1,194 @@
...
+
+        # Start profiling timer
+        self.done = False
+        self.period_sec = 5
+        self.t_start = time.time()
+        self.t_next = self.t_start
+        self.stats = ["time,cpu,mem"]
+        self._timeout()
+
+    def run_cmd(self, cmd: str) -> str:
```
thetorpedodog, line 116: This doesn't use any self-state. Should it be a module-level function?

src/tiledb/cloud/utilities/profiler.py
```diff
@@ -0,0 +1,194 @@
...
+    with tiledb.open(uri, "w") as A:
+        A[t_now_ms] = {
+            "id": [id],
+            "op": [op],
+            "data": [data],
+            "extra": [extra],
+        }
+
+
+class Profiler(object):
```
thetorpedodog, line 78: It seems like it might make more sense to have this as a context manager so you can do:

```python
with my_profiler.profile(whatever_param):
  ...
```

This might also help with the potential dangling array below (`self.array = tiledb.open(uri, "w")`)

src/tiledb/cloud/utilities/profiler.py
```diff
@@ -0,0 +1,194 @@
...
+
+        self.write("stats", mem, "\n".join(self.stats))
+
+    def _timeout(self):
+        """
+        Capture system stats and schedule the next timeout.
+        """
+
+        try:
+            cpu = open("/sys/fs/cgroup/cpu/cpuacct.usage").readline().strip()
```
thetorpedodog, line 180: consider `pathlib.Path(...).read_text().strip()` to ensure everything is closed (and maybe just pull open-read-strip into a function)

src/tiledb/cloud/utilities/common.py
```diff
@@ -0,0 +1,197 @@
...
+            graph.wait()
+        except tiledb.cloud.tiledb_cloud_error.TileDBCloudError as e:
+            print(f"Fatal graph error:\n{e}")
+            print_logs(graph, debug)
+            raise e
+
+    print_logs(graph, debug)
+
+
+def print_logs(graph: tiledb.cloud.dag.DAG, debug: bool = False):
```
thetorpedodog, line 153: `debug` should probably be kw-only.

src/tiledb/cloud/utilities/profiler.py
```diff
@@ -0,0 +1,194 @@
...
+class Profiler(object):
+    """
+    A simple profiler to log start/finish events with CPU and memory usage.
+    """
+
+    def __init__(
+        self,
+        uri: str,
+        id: str,
+        trace: bool = False,
```
thetorpedodog, line 87: kw-only here too

src/tiledb/cloud/vcf/ingestion.py
```diff
@@ -0,0 +1,977 @@
...
+import logging
+import subprocess
+from collections import defaultdict
+from math import ceil
+from typing import Any, Mapping, Optional, Sequence, Tuple, Union
+
+import numpy as np
+import tiledbvcf
+
+import tiledb.cloud
```
thetorpedodog, line 12: To avoid circular import weirdness, it would be best to import the specific modules that contain whatever you're using.

src/tiledb/cloud/vcf/ingestion.py
```diff
@@ -0,0 +1,977 @@
...
+    :param id: profiler event id, defaults to None
+    :return: logger instance, optionally a profiler object
+    """
+
+    set_aws_context(config)
+
+    level = logging.DEBUG if VERBOSE else logging.INFO
+    logger = get_logger(__name__, level)
+
+    logger.debug(
```
thetorpedodog, line 85: let the logger do the substitution:

```python
logger.debug(
    "tiledb=%s, libtiledb=%s, tiledbvcf=%s", tiledb.version(), ...
)

File: src/tiledb/cloud/vcf/ingestion.py [hunk 22]
```diff
@@ -0,0 +1,977 @@
...
+    :param extra_attrs: INFO/FORMAT fields to materialize, defaults to None
+    :param vcf_attrs: VCF with all INFO/FORMAT fields to materialize, defaults to None
+    :return: dataset URI
+    """
+
+    logger = setup(config)
+
+    # Check if the dataset already exists
+    if tiledb.object_type(dataset_uri) != "group":
+        logger.info(f"Creating dataset: '{dataset_uri}'")
```
thetorpedodog, line 173: e.g. here `logger.info("Creating dataset: %r", dataset_uri)`

src/tiledb/cloud/vcf/ingestion.py
```diff
@@ -0,0 +1,977 @@
...
+    :param max_files: maximum number of URIs returned, defaults to None
+    :return: list of URIs
+    """
+
+    logger, profiler = setup(config, dataset_uri)
+
+    use_s3 = search_uri.startswith("s3://")
+    # Run command to find URIs matching the pattern
+    if use_s3:
+        cmd = f"exec aws s3 sync --dryrun --exclude '*' --include '{pattern}' {search_uri} ."
```
thetorpedodog, line 269: Using `shell` here is vulnerable to accidental shell injection; instead the command should be built as a sequence:

```python
cmd = ("aws", "s3", "sync" "--dryrun", "--exclude", "*", "--include", pattern, search_uri, ".")
```

and then executed without `shell=True`.

src/tiledb/cloud/vcf/ingestion.py
```diff
@@ -0,0 +1,977 @@
...
+
+    logger.debug(cmd)
+    p2 = subprocess.Popen(
+        cmd,
+        shell=True,
+        text=True,
+        stdin=p1.stdout,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+    )
```
thetorpedodog, line 296: Could we do this in Python, iteratively reading in lines and handling them as needed rather than shelling out to grep?

src/tiledb/cloud/vcf/ingestion.py
```diff
@@ -0,0 +1,977 @@
...
+    p1.kill()
+    if res_stderr:
+        print(res_stderr)
+
+    # Build list of URIs from command output.
+    # Example line from s3:
+    # (dryrun) download: s3://1000genomes-dragen-v3.7.6/foo to foo
+    result = []
+    if res_stdout:
+        for line in res_stdout.strip().split("\n"):
```
thetorpedodog, line 309: .splitlines?

src/tiledb/cloud/utilities/common.py
gspowley: This is useful in a notebook environment, where we have to read the credentials from somewhere. We have been passing around functions in notebooks that do the same thing, so I thought we could formalize it.

It's fine if we take some time to think about it. Let me know and I can remove `read_aws_config`.

src/tiledb/cloud/vcf/ingestion.py
gspowley: These functions can go away. They were added with some other code to test the code on Cloud without putting everything in one file, and they probable weren't necessary then either.

gspowley:
Thanks for the detailed review @thetorpedodog! I addressed all of the comments except the ones related to `find_uris_udf`, which we plan to rewrite and replace in a future PR.

src/tiledb/cloud/vcf/ingestion.py
```diff
@@ -0,0 +1,970 @@
...
+            sample_uris[i * batch_size : (i + 1) * batch_size],
+            config=config,
+            id=f"manifest-ingest-{i}",
+            name=f"Ingest VCF Manifest {i+1}/{num_partitions} ",
+        )
+        if prev_consolidate:
+            ingest.depends_on(prev_consolidate)
+
+        if consolidate:
+            consolidate.depends_on(ingest)
```
thetorpedodog, line 739: This is a little hard for me to follow. Is this designed to have N ingest steps, then a consolidate step, then N more ingest steps, then a consolidate step, etc.? It might be worth adding a comment explaining this.

thetorpedodog (commented):
Overall looks pretty good. A few things I didn't see last time but mostly minor changes now.

src/tiledb/cloud/vcf/ingestion.py
gspowley: Thanks again @thetorpedodog. I addressed the latest comments and got rid of all the `shell=True` subprocesses.

src/tiledb/cloud/vcf/ingestion.py
```diff
@@ -0,0 +1,997 @@
...
+
+        def get_records(vcf: str) -> int:
+            # TODO: remove when bcftools index -n works with S3
+            return 0
+
+            cmd = ["bcftools", "index", "-n", vcf]
+            res = subprocess.run(cmd, capture_output=True, text=True)
+            if res.stderr:
+                print(res.stderr)
+                return 0
```
thetorpedodog, line 467: This code won’t execute, so this is not a critical thing, but:

Just to make sure—if there is any stderr output, you want to return 0? If that’s what you want to do, that’s fine, but it might be worth adding a comment explaining why.

thetorpedodog (approved):
One suggestion but otherwise this looks good!

src/tiledb/cloud/vcf/ingestion.py
gspowley: Yes, the 0 in the manifest will indicate there's a problem with the index that needs to be addressed. Added a comment there to explain.

PR #391, opened 2023-04-22
Add tool to force function pickling by value.

thetorpedodog:
When registering certain UDFs, like `trampoline.run_python_function`, we want to ensure that they're registered by reference, not by value. Removing the `__module__` from the function is the simplest way to make that happen with the current version of cloudpickle.

This may be worth exporting in the future, or alternately even using it in the default UDF execution/registration process, but until we have a better idea of how to work with it, I'd like it to stay internal-only.

---

This was originally based on [code in the TileDB-Cloud-UDFs package](https://github.com/TileDB-Inc/TileDB-Cloud-UDFs/blob/35d53852387941cb78f4703cb3a8ad1b065450d0/register_udfs.py#L22-L28), but with some more digging into the exact reason that it worked (notably, that clearing `__globals__` was a red herring).

### DIFF
--- a/src/tiledb/cloud/_common/functions.py
+++ b/src/tiledb/cloud/_common/functions.py
@@ -1,6 +1,7 @@
 """Utilities for dealing with functions for serialization and registration."""
 
 import inspect
+import types
 from typing import Callable, Optional, TypeVar, Union
 
 _T = TypeVar("_T")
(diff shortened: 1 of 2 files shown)


ihnorton (approved):
👍 pending CI. Nice tests! I suspect we'll have to adjust this to take version differences into account, especially 3.11 when we get to that (may be the cause of the 3.7 CI failures too, can't quite tell).

thetorpedodog:
> 👍 pending CI. Nice tests! I suspect we'll have to adjust this to take version differences into account, especially 3.11 when we get to that (may be the cause of the 3.7 CI failures too, can't quite tell).

That was the issue—now it just computes the pickle (in-process) rather than having one static value.

thetorpedodog:
> Nice tests!

I should also add that I wrote the tests as much to confirm my own understanding of how cloudpickle does things as anything else (and it has been tremendously useful for that).

PR #392, opened 2023-04-24
Initial version of SOMA ingestor.

thetorpedodog:
This builds and executes a trivial server-side task graph (just the one node) that ingests SOMA data into a TileDB array. `soma.io` currently only supports ingesting the whole thing at once.

For maximum flexibility, credentials and other configuration can be provided directly as TileDB config settings.

With some trickery, we can pass the TileDB VFS IO object directly into `anndata.read_h5ad` and avoid having to make a local copy of the input h5ad file.

---

It took me way too much digging through anndata and h5py's source code and banging my head against a wall to come up with a solution to read h5ad files that *didn't* involve literally just copying them to a local temporary folder. This came to me only just now.

The main thing I'm concerned about is: are we providing the right set of options here? For context, we plan on providing a UI for this, and the UI will take the steps of turning e.g. storage keys into TileDB config data, so we are not expecting users to directly enter `{"tiledb.vfs.whatever": "some-value"}`–type JSON data. However, doing it this way will allow us to support any backend with only changes on the UI side and not needing to change the backend UDF itself.

---

Context: epic sc-30665 and story sc-30669

### DIFF
--- a/src/tiledb/cloud/soma.py
+++ b/src/tiledb/cloud/soma.py
@@ -1,0 +1,141 @@
+import logging
+from typing import Dict, Optional
+
+import tiledb
+from tiledb.cloud import dag
+from tiledb.cloud._common import functions
+
+_DEFAULT_RESOURCES = {"cpu": "8", "memory": "8Gi"}
+"""Default resource size; equivalent to a "large" UDF container."""
+
+
+def ingest_h5ad(
+    *,
+    output_uri: str,
+    input_uri: str,
+    measurement_name: str,
+    extra_tiledb_config: Optional[Dict[str, object]],
+    platform_config: Optional[Dict[str, object]],
+    ingest_mode: str,
+) -> None:
+    """Performs the actual work of ingesting H5AD data into TileDB.
+
+    :param output_uri: The output URI to write to. This will probably look like
+        ``tiledb://namespace/some://storage/uri``.
+    :param input_uri: The URI of the H5AD file to read from. This file is read
+        using TileDB VFS, so any path supported (and accessible) will work.
+    :param measurement_name: The name of the Measurement within the Experiment
+        to store the data.
+    :param extra_tiledb_config: Extra configuration for TileDB.
+    :param platform_config: The SOMA ``platform_config`` value to pass in,
+        if any.
+    :param ingest_mode: One of the ingest modes supported by
+        ``tiledbsoma.io.read_h5ad``.
+    """
+
+    import anndata
+    import tiledbsoma
+    import tiledbsoma.logging as somalog
+    from tiledbsoma import io
+
+    logging.basicConfig(level=logging.INFO)
+    somalog.info()
+
+    # While h5ad supports any file-like object, annndata specifically
+    # wants only an `os.PathLike` object. The only thing it does with
+    # the PathLike is to use it to get the filename.
+    class _FSPathWrapper:
+        """Tricks anndata into thinking a file-like object is an os.PathLike.
+
+        While h5ad supports any file-like object, anndata specifically wants
+        an os.PathLike object, which it uses *exclusively* to get the "filename"
+        of the opened file.
+
+        We need to provide ``__fspath__`` as a real class method, so simply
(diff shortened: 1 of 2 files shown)


johnkerl (commented):
I'm confused -- why do this in tiledb-cloud-py? We have no VCF or other apps in tiledb-cloud-py.

I thought we were going to use https://github.com/TileDB-Inc/TileDB-Cloud-Life-Sciences -- ?

thetorpedodog:
> I'm confused -- why do this in tiledb-cloud-py? We have no VCF or other apps in tiledb-cloud-py.
>
> I thought we were going to use https://github.com/TileDB-Inc/TileDB-Cloud-Life-Sciences -- ?

This isn’t correct; there’s [an entire VCF module in this package](https://github.com/TileDB-Inc/TileDB-Cloud-Py/tree/main/src/tiledb/cloud/vcf) which contains similar functionality. This code is intended to be available as a library function to be executed from a UDF.

johnkerl:
@thetorpedodog thank you!

My main concern about this is:

* For UDFs per se we have 2GB nodes and 8GB nodes (`resource_class=="large"`) for the latter
* My previous attempts to experiment with distributed ingest (admittedly, half a year or more ago) resulted on node-level OOMs even with `resource_class=="large"`
* It had been my hope (perhaps not articulated by me in as many places as I had hoped) that we could use batch-mode runners which have fewer memory constraints

In any case before going farther we need to test this with (at least) H5AD with over a million cells.

To get such, IOO've gone to https://cellxgene.cziscience.com/datasets then moved the slider for cell count, then save as `.h5ad`. For example:


* https://cellxgene.cziscience.com/collections/c114c20f-1ef4-49a5-9c2e-d965787fb90c
* https://cellxgene.cziscience.com/collections/0a839c4b-10d0-4d64-9272-684c49a2c8ba
* https://cellxgene.cziscience.com/collections/edb893ee-4066-4128-9aec-5eb2b03f8287

src/tiledb/cloud/soma.py
```diff
@@ -0,0 +1,143 @@
...
+    measurement_name: str,
+    extra_tiledb_config: Optional[Dict[str, object]],
+    platform_config: Optional[Dict[str, object]],
+    ingest_mode: str,
+) -> None:
+    """Performs the actual work of ingesting H5AD data into TileDB.
+
+    :param output_uri: The output URI to write to. This will probably look like
+        ``tiledb://namespace/some://storage/uri``.
+    :param input_uri: The URI of the H5AD file to read from. This file is read
```
johnkerl, line 26: Empirically I've found that:

* If we use `anndata.read_h5ad(foo)` where `foo` is a `tiledb.VFS`, this does work
* If we use `anndata.read_h5ad(foo, "r")` then this does _not_ work
  * Note that `"r"` is important for memory reduction in memory-constrained environments (e.g. UDF nodes)
  * Issue is that `anndata.read_h5ad` docstrings say the first argument is a path, but in reality, a `tiledb.VFS` object plumbs almost all the way through ...
  * ... except with `"r"` mode where at one point in the anndata code it does assume the first arg is a path per se

src/tiledb/cloud/soma.py
```diff
@@ -0,0 +1,143 @@
...
+    logging.basicConfig(level=logging.INFO)
+    somalog.info()
+
+    # While h5ad supports any file-like object, annndata specifically
+    # wants only an `os.PathLike` object. The only thing it does with
+    # the PathLike is to use it to get the filename.
+    class _FSPathWrapper:
+        """Tricks anndata into thinking a file-like object is an os.PathLike.
+
+        While h5ad supports any file-like object, anndata specifically wants
```
johnkerl, line 52: Here you may be having re-discovered what I pointed out above

johnkerl (approved):
LGTM as an initial version as long as the CI fail is addressed

johnkerl:
@thetorpedodog what are the remaining steps to get this PR merged?

johnkerl:
@thetorpedodog it looks like the sole CI fail might be resolved by a retry but, oddly, I don't appear to have permissions to restart the failed CI job.

PR #394, opened 2023-04-25
Add distributed VCF query

gspowley:
Add a one-line distributed VCF query, which takes a list of regions and samples, partitions the query across multiple tasks, runs a task graph, and returns the results in an Arrow table.

```python
from tiledb.cloud.vcf import read

dataset_uri = ...
config = ...
regions = ...
samples = ...

result = read(
    dataset_uri=dataset_uri,
    config=config,
    regions=regions,
    samples=samples,
)
```

### DIFF
--- a/src/tiledb/cloud/vcf/query.py
+++ b/src/tiledb/cloud/vcf/query.py
@@ -103,6 +103,9 @@ def vcf_query_udf(
     # Adjust inputs
     if config is None:
         config = {}
+    else:
+        config = config.copy()
+
     if isinstance(attrs, str):
         attrs = [attrs]
     if isinstance(regions, str):
@@ -186,9 +189,7 @@ def concat_tables_udf(
     logger = setup(config, verbose)
 
     with Profiler(array_uri=log_uri) as prof:
-        table = pa.concat_tables(
-            [x for x in tables if x is not None and x.num_rows > 0]
-        )
+        table = pa.concat_tables(x for x in tables if x is not None and x.num_rows > 0)
         prof.write("result", table.num_rows, table.nbytes)
 
     memory_usage_gb = max_memory_usage() / (1 << 30)
@@ -251,10 +252,10 @@ def build_read_dag(
 
     # Set number of sample partitions
     num_samples = len(samples)
-    sample_batch_size = ceil(num_samples // (max_workers // num_region_partitions))
+    sample_batch_size = ceil(num_samples * num_region_partitions / max_workers)
     sample_batch_size = min(sample_batch_size, MAX_SAMPLE_BATCH_SIZE)  # max batch size
     sample_batch_size = max(sample_batch_size, MIN_SAMPLE_BATCH_SIZE)  # min batch size
-    num_sample_partitions = max(1, num_samples // sample_batch_size)  # at least one
+    num_sample_partitions = ceil(num_samples / sample_batch_size)
 
     logger.debug("num_samples=%d", num_samples)
     logger.debug("sample_batch_size=%d", sample_batch_size)
--- a/src/tiledb/cloud/vcf/query.py
+++ b/src/tiledb/cloud/vcf/query.py
@@ -25,12 +25,11 @@ MAX_WORKERS = 40
 MAX_SAMPLE_BATCH_SIZE = 500
 MIN_SAMPLE_BATCH_SIZE = 20
 
-# Test and debug hooks
-verbose = False
-trace = False
 
-
-def setup(config: Optional[Mapping[str, Any]] = None) -> logging.Logger:
+def setup(
+    config: Optional[Mapping[str, Any]] = None,
+    verbose: bool = False,
+) -> logging.Logger:
     """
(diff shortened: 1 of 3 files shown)


src/tiledb/cloud/vcf/query.py
```diff
@@ -0,0 +1,308 @@
...
+    :param regions: genomics regions to read, defaults to None
+    :param bed_file: URI of a BED file containing genomics regions to read, defaults to None
+    :param num_region_partitions: number of region partitions, defaults to 1
+    :param samples: sample names to read, defaults to None
+    :param memory_budget_mb: VCF memory budget in MiB, defaults to 1024
+    :param af_filter: allele frequency filter, defaults to None
+    :param log_uri: log array URI for profiling, defaults to None
+    :param namespace: TileDB-Cloud namespace, defaults to None
+    :param resource_class: TileDB-Cloud resource class for UDFs, defaults to None
+    :param return_dag: return a DAG instead of the query results, defaults to False
```
thetorpedodog, line 236: I'm not a fan of "return one thing, or if this parameter is set, a completely different thing". Maybe make two functions, one which forwards its arguments to the other and returns the output?

```python
def build_read_dag(...) -> Tuple[...]:
  ...

def read(dataset_uri: str, **kwargs) -> ...:
  dag, table = build_read_dag(dataset_uri, **kwargs)
  # ...
  return ...
```

gspowley: Ok, I'll make that change.

src/tiledb/cloud/vcf/query.py
```diff
@@ -0,0 +1,308 @@
...
+    try:
+        tiledb.default_ctx(config)
+    except tiledb.TileDBError:
+        # Ignore error if the default context was already set
+        pass
+
+    set_aws_context(config)
+
+    level = logging.DEBUG if verbose else logging.INFO
+    logger = get_logger(level)
```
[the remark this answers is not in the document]
gspowley: Right, I had that in the back of my mind. It works in a UDF with my local test setup, but won't work when the code is released.

I'll add a verbose arg as needed.

src/tiledb/cloud/vcf/query.py
gspowley: I ended up keeping the full function signature for `read` to preserve the docstrings, since these are user facing functions.

src/tiledb/cloud/vcf/query.py
```diff
@@ -0,0 +1,364 @@
...
+        attrs = [attrs]
+    if isinstance(regions, str):
+        regions = [regions]
+    if isinstance(samples, str):
+        samples = [samples]
+
+    logger = setup(config, verbose)
+    logger.debug("tiledbvcf=%s", tiledbvcf.version)
+
+    config["rest.use_refactored_array_open"] = True
```
thetorpedodog, line 116: When running locally, this will mutate the dictionary that the user passed in (and which might be used by later steps). Is that OK?

src/tiledb/cloud/vcf/query.py
```diff
@@ -0,0 +1,364 @@
...
+    :param log_uri: log URI for profiling, defaults to None
+    :param verbose: verbose logging, defaults to False
+    :return: concatenated Arrow table
+    """
+
+    logger = setup(config, verbose)
+
+    with Profiler(array_uri=log_uri) as prof:
+        table = pa.concat_tables(
+            [x for x in tables if x is not None and x.num_rows > 0]
```
thetorpedodog, line 190: nit: no need for the `[]`s in here; giving it an iterable of tables will work fine

src/tiledb/cloud/vcf/query.py
```diff
@@ -0,0 +1,364 @@
...
+    # Return an empty table if no samples or regions are specified.
+    # This avoids reading the entire array by accident.
+    if samples is None or (regions is None and bed_file is None):
+        return pa.table({})
+
+    attrs = attrs or DEFAULT_ATTRS
+
+    # Set number of sample partitions
+    num_samples = len(samples)
+    sample_batch_size = ceil(num_samples // (max_workers // num_region_partitions))
```
thetorpedodog, line 254: `//` is `__floordiv__`, so `ceil`ing it won’t do anything. Do you want `__truediv__` (i.e., `/`)? (And maybe `num_samples * num_region_partitions / max_workers`?)

thetorpedodog (approved):
A couple questions/suggestions which may be fine as-is, otherwise looks good.

src/tiledb/cloud/vcf/query.py
gspowley: Good catch. It's probably OK, since local will be used for testing, but it's easy to fix that.

src/tiledb/cloud/vcf/query.py
gspowley: Right, that math needs some work after some recent changes. Fixing that.

PR #396, opened 2023-04-28
Improve notebook APIs: add timestamp, handle on-exists better.

thetorpedodog:
- Adds a timestamp parameter to the notebook downloaders to provide time travel directly from the notebook API.
- Sets default values for upload functions to avoid needing to pass every single parameter.
- Handles overwrite-on-exists case better; will not fail if the array doesn't yet exist.

### DIFF
--- a/src/tiledb/cloud/notebook.py
+++ b/src/tiledb/cloud/notebook.py
@@ -3,10 +3,11 @@ Python support for notebook I/O on Tiledb Cloud. All notebook JSON content
 is assumed to be encoded as UTF-8.
 
 """
+import datetime
 import enum
 import posixpath
 import time
-from typing import Optional, Tuple
+from typing import Optional, Tuple, Union
 
 import numpy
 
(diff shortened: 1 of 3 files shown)


src/tiledb/cloud/notebook.py
```diff
@@ -181,37 +163,47 @@ def upload_notebook_contents(
...
 
-    ctx = tiledb.cloud.Ctx(
-        {"rest.creation_access_credentials_name": storage_credential_name}
-    )
+    ctx = client.Ctx({"rest.creation_access_credentials_name": storage_credential_name})
 
-    if on_exists is OnExists.FAIL:
+    tiledb_uri = f"tiledb://{namespace}/{array_name}"
+    try:
+        arr = tiledb.open(tiledb_uri, ctx=ctx)
```
Shelnutt2, line 189: You can also do `tiledb.array_exists`, which is light weight check compared to a full array open.

Shelnutt2 (approved):
Nice improvements, one suggestion.

src/tiledb/cloud/notebook.py
thetorpedodog: It looks like internally it does the same thing, but it does not (currently) accept a `ctx`.

Shelnutt2: It does not quite do the same thing. Opening the array might also return the nonempty domain, metadata, and fragment metadata in addition to the array schema. The exist function today will only fetch the schema and then discard it.

thetorpedodog: https://github.com/TileDB-Inc/TileDB-Py/blob/0.21.1/tiledb/highlevel.py#L139-L153

thetorpedodog:
[sc-28337]

PR #397, opened 2023-05-02
Set the executing graph name from the builder's name.

thetorpedodog:
We forgot to take the builder's name into account when running a graph, so if you built a graph and ran it:

    grf = tg.Builder("my graph")
    # ...
    tg.execute(grf)

the name in the UI would still be "untitled". This propagates the graph name into the executor.


Shelnutt2 (approved):

aaronwolen (approved):
Thanks!!

PR #406, opened 2023-05-17
Add wrapper function to call bioimaging ingestion from UDFs.

thetorpedodog:
Returning certain types of objects with complex local state doesn't work (and in this case, is also much more information than we need). This change adds a wrapper function which returns only the information needed by the UI after an ingestor is run.


ktsitsi (approved):
👍

PR #411, opened 2023-05-25
Adding direct support as kwarg of access_credential_name

ktsitsi:
This PR:

- adds support of `access_credential_name` as an extra argument to the ingestion function. Since the functionality could be replicated by the config argument `rest.create_access_credential_name` as well we prioritise upon existence the user function argument over the config parameter, by editing the config value.

### DIFF
--- a/src/tiledb/cloud/bioimg/ingestion.py
+++ b/src/tiledb/cloud/bioimg/ingestion.py
@@ -48,6 +48,7 @@ def ingest(
         :param io_uris: Pairs of tiff input - output tdb uris
         :param config: dict configuration to pass on tiledb.VFS
         :param workers: Number of threads that will spawn for parallelizing ingestion
+        :param access_credentials_name: The registered access key to REST server.
         """
 
         from tiledb.bioimg.converters.ome_tiff import OMETiffConverter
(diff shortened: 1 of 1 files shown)


Shelnutt2 (changes requested):
We need to support passing in the access credential name to the nodes where the ingestion happens too. There is a parameter on `submit`, `access_credential_name`. This will add the role to the nodes for access to the bucket during ingestion.

I think we might always want to separate out credentials for the source bucket vs the arrays? This would allow the raw data to live in one location and the ingested TileDB data to be in another?

ktsitsi:
> We need to support passing in the access credential name to the nodes where the ingestion happens too. There is a parameter on `submit`, `access_credential_name`. This will add the role to the nodes for access to the bucket during ingestion.
> 
> I think we might always want to separate out credentials for the source bucket vs the arrays? This would allow the raw data to live in one location and the ingested TileDB data to be in another?

Discussing with @antalakas this morning we might want to approach this in a broad fashion. Quoting below @thetorpedodog in the other PR for SOMA ingestor and this might also apply in the VCF one 

> are we providing the right set of options here? For context, we plan on providing a UI for this, and the UI will take the steps of turning e.g. storage keys into TileDB config data, so we are not expecting users to directly enter {"tiledb.vfs.whatever": "some-value"}–type JSON data. However, doing it this way will allow us to support any backend with only changes on the UI side and not needing to change the backend UDF itself.

ktsitsi:
This will be covered by #412

PR #412, opened 2023-05-25
Run all i/o parts of the bioimaging ingestor in the graph.

thetorpedodog:
Because we cannot (currently) provide access credentials to a UDF executed directly from the client, we need to do all i/o related tasks in the server-side task graph that we build. This pulls the initial listing of input URIs into the graph itself and uses the parallelization support ("output expansion") of server-side graphs to run the ingestion in parallel.

---

I don’t have source data setup for doing this. Kostas, can you do a test run?

### DIFF
--- a/src/tiledb/cloud/bioimg/ingestion.py
+++ b/src/tiledb/cloud/bioimg/ingestion.py
@@ -1,9 +1,15 @@
+import base64
 import os
 import posixpath
 from typing import Any, Dict, List, Mapping, Optional, Sequence, Tuple, Union
 
+import cloudpickle
+
 import tiledb
+from tiledb.cloud import client
 from tiledb.cloud import dag
+from tiledb.cloud import rest_api
+from tiledb.cloud._common import functions
 from tiledb.cloud._common.utils import logger
 
 DEFAULT_RESOURCES = {"cpu": "8", "memory": "4Gi"}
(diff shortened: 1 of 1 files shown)


src/tiledb/cloud/bioimg/ingestion.py
```diff
@@ -34,9 +41,35 @@ def ingest(
...
+    source_list = [source] if isinstance(source, str) else list(source)
+    max_workers = None if num_batches else 20  # Default picked heuristically.
+
+    def build_io_uris() -> List[Tuple[str, str]]:
+        """Match input uri/s with output destinations
+
+        :param source: A sequence of paths or path to input
+        :param output_dir: A path to the output directory
+        """
+        vfs = tiledb.VFS(config=config)
```
antalakas, line 53: This has to be ` vfs = tiledb.VFS()` since it will use `/home/udf/.aws/config` to load credentials

antalakas (changes requested):

src/tiledb/cloud/bioimg/ingestion.py
```diff
@@ -34,9 +41,35 @@ def ingest(
...
+            for uri in my_source_list
+        ]
+
+    def build_input_batches():
+        """Groups input URIs into batches."""
+        uri_pairs = build_io_uris()
+        # If the user didn't specify a number of batches, run every import
+        # as its own task.
+        my_num_batches = num_batches or len(uri_pairs)
+        return [uri_pairs[n::num_batches] for n in range(my_num_batches)]
```
ktsitsi, line 69: If the `num_batches` given by the user is greater than the `uri_pairs` length. We get empty lists like:
```python
pairs = [1,2,3,4,5,6,7,8,9,10]
num_batches = 20
[pairs[n::num_batches] for n in range(num_batches)]
Out[6]: 
[[1], [2], [3], [4], [5], [6], [7], [8], [9], [10], [], [], [], [], [], [], [], [], [], []]
```
Will this create empty tasks for the lister?

ktsitsi (changes requested):

src/tiledb/cloud/bioimg/ingestion.py
thetorpedodog: It would. Fixed.

src/tiledb/cloud/bioimg/ingestion.py
thetorpedodog: I have a tentative change to do this but have it separated out as its own rev pending investigating other issues.

ktsitsi:
Closing as covered by #492

PR #418, opened 2023-06-08
Use node ID, not node name, to find nodes sent back from server.

thetorpedodog:
When associating nodes in a task graph log to those in the graph itself, we were using the node's name. This was not intended to be used for that purpose; instead we should be using its UUID, which is IDentifies it, Uniquely (Universally).


Shelnutt2 (approved):

Shelnutt2 (approved):

Shelnutt2:
@thetorpedodog shall we merge this?

NikolaosPapailiou (approved):

PR #419, opened 2023-06-09
VCF ingestion updates

gspowley:
Updates to VCF ingestion:
* Add `access_credentials_name` parameter to support AWS assume role.
* Add task graph retry strategy and enable `resume` by default.
* Support creating index files on the fly.
* Add `anchor_gap` parameter to set the anchor gap length, which is useful for structural variant VCFs.
* Add `max_sample` parameter to limit the number of samples ingested, which is different than using `max_files` to limit the number of samples added to the manifest.
* Add `process_stream` utility function to stream a file into a subprocess using VFS. This replaces use cases like `aws s3 cp - | bcftools` and enables all storage backends supported by VFS.

### DIFF
--- a/src/tiledb/cloud/utilities/_common.py
+++ b/src/tiledb/cloud/utilities/_common.py
@@ -233,7 +233,8 @@ def process_stream(
                 try:
                     process.stdin.write(data)
                 except BrokenPipeError:
-                    # The subprocess has exited, stop reading
+                    # The subprocess has exited, stop reading.
+                    # This is an expected situation.
                     break
                 data = fp.read(read_size)
             try:
--- a/src/tiledb/cloud/utilities/_common.py
+++ b/src/tiledb/cloud/utilities/_common.py
@@ -2,9 +2,11 @@ import configparser
 import logging
 import os
 import pathlib
+import subprocess
 import sys
-from typing import Any, Mapping, Optional
+from typing import Any, Mapping, Optional, Tuple
 
+import tiledb
 from tiledb.cloud import dag
 from tiledb.cloud.tiledb_cloud_error import TileDBCloudError
 
@@ -138,7 +140,11 @@ def run_dag(
             _print_logs(graph, debug=debug)
             raise
 
-    _print_logs(graph, debug=debug)
+    try:
+        _print_logs(graph, debug=debug)
+    except Exception:
+        # Ignore errors in case logs are not available.
+        pass
 
 
 def _print_logs(
@@ -152,11 +158,7 @@ def _print_logs(
     :param graph: DAG object
     :param debug: print debug logs, defaults to False
     """
-    try:
-        server_logs = dag.server_logs(graph)
-    except Exception:
-        # TODO: get server logs from a batch UDF
-        return
+    server_logs = dag.server_logs(graph)
 
     for node in server_logs.nodes:
         if debug:
@@ -200,3 +202,44 @@ def max_memory_usage() -> int:
         result = 0
 
     return result
+
+
+def process_stream(
+    uri: str,
+    cmd: str,
+    read_size: int = 16 << 20,
+) -> Tuple[str, str]:
+    """
+    Process a stream of data from VFS with a subprocess.
+
+    If the file is large and the subprocess only reads a small amount of data,
+    then reduce `read_size` to improve performance.
+
+    :param uri: file URI
+    :param cmd: command to run in the subprocess
+    :param read_size: number of bytes to read per iteration, defaults to 16 MiB
+    :return: stdout and stderr from the subprocess
+    """
+
+    with tiledb.VFS().open(uri) as fp:
+        with subprocess.Popen(
+            cmd.split(),
+            stdin=subprocess.PIPE,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+        ) as process:
+            while data := fp.read(read_size):
+                try:
+                    process.stdin.write(data)
+                except BrokenPipeError:
+                    # The subprocess has exited, stop reading
+                    break
+            try:
+                process.stdin.close()
+            except BrokenPipeError:
+                # Ignore broken pipe when closing stdin
+                pass
+            stdout = process.stdout.read().decode().strip()
+            stderr = process.stderr.read().decode().strip()
+
+    return stdout, stderr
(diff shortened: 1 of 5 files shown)


src/tiledb/cloud/utilities/_common.py
```diff
@@ -138,7 +140,11 @@ def run_dag(
             _print_logs(graph, debug=debug)
             raise
 
-    _print_logs(graph, debug=debug)
+    try:
+        _print_logs(graph, debug=debug)
+    except Exception:
+        # Ignore errors in case logs are not available.
+        pass
```
thetorpedodog, line 147: Would it be worth it to do this in `_print_logs` itself?

src/tiledb/cloud/utilities/_common.py
```diff
@@ -200,3 +202,44 @@ def max_memory_usage() -> int:
...
+
+    :param uri: file URI
+    :param cmd: command to run in the subprocess
+    :param read_size: number of bytes to read per iteration, defaults to 16 MiB
+    :return: stdout and stderr from the subprocess
+    """
+
+    with tiledb.VFS().open(uri) as fp:
+        with subprocess.Popen(
+            cmd.split(),
```
thetorpedodog, line 226: To avoid problems caused by e.g. spaces in filenames or other unexpected behavior, I would recommend passing this in as the list of arguments to be passed to the process.

src/tiledb/cloud/utilities/_common.py
```diff
@@ -200,3 +202,44 @@ def max_memory_usage() -> int:
...
+            stdin=subprocess.PIPE,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+        ) as process:
+            while data := fp.read(read_size):
```
thetorpedodog, line 231: The walrus operator is py3.8+–only. We currently support py3.7 in TileDB Cloud, and I assume we probably don’t want to limit parts of this package to py3.8+. (Oh how I wish we could use it!)

src/tiledb/cloud/utilities/_common.py
```diff
@@ -200,3 +202,44 @@ def max_memory_usage() -> int:
...
+                try:
+                    process.stdin.write(data)
+                except BrokenPipeError:
+                    # The subprocess has exited, stop reading
+                    break
```
thetorpedodog, line 236: Is this a sign of a premature exit or error condition on whatever program you’re running?

src/tiledb/cloud/vcf/ingestion.py
```diff
@@ -584,7 +558,27 @@ def ingest_samples_udf(
...
+        # TODO: Handle un-bgzipped files
+        """
+        new_sample_uris = []
+        for uri in sample_uris:
+            if not is_bgzipped(uri):
+                logger.debug("bgzipping and indexing %r", uri)
+                uri = bgzip_and_index(uri)
+            new_sample_uris.append(uri)
+        sample_uris = new_sample_uris
+        """
```
thetorpedodog, line 581: I would either remove this or make it a comment rather than a multiline string.

src/tiledb/cloud/vcf/utils.py
```diff
@@ -0,0 +1,127 @@
...
+    """
+    Returns the sample name in a VCF file.
+
+    If there are multiple samples, return a comma-separated list of sample names.
+
+    :param vcf_uri: URI of the VCF file
+    :return: sample name
+    """
+
+    cmd = "bcftools query -l"
```
thetorpedodog, line 48: e.g. here `('bcftools', 'query', '-l')`

src/tiledb/cloud/utilities/_common.py
gspowley: Sure, we can put it there in case we decide to call `_print_logs` somewhere else.

src/tiledb/cloud/utilities/_common.py
gspowley: Good point, changing that.

src/tiledb/cloud/utilities/_common.py
gspowley: Too bad, the walrus is nice and concise.

src/tiledb/cloud/utilities/_common.py
gspowley: Also an expected early exit. For example, if we're streaming a 2G file to a program that only reads the first 10k of a file, it will exit early and we can stop streaming. `bcftools` does this when reading the sample name from a VCF file.

src/tiledb/cloud/vcf/ingestion.py
gspowley: Removing.

src/tiledb/cloud/vcf/utils.py
gspowley: Changing all of these commands to lists.

src/tiledb/cloud/vcf/utils.py
```diff
@@ -0,0 +1,127 @@
...
+    res = subprocess.run(cmd, capture_output=True, text=True)
+
+    # If there is an error, this means there was a problem reading the
+    # index file or the index file is an old format that does not
+    # contain the required metadata. In either case, return 0 to
+    # indicate a problem with the index that needs to be addressed
+    # before ingesting the sample.
+    if res.stderr:
+        print(res.stderr)
+        return 0
```
thetorpedodog, line 87: another case where `Optional` or an exception might be better, so that the failure value is of a different type than the success value

src/tiledb/cloud/vcf/utils.py
```diff
@@ -0,0 +1,127 @@
...
+    # for `bcftools index -n`, which only reads the index file.
+    vcf_file = os.path.basename(vcf_uri)
+    open(vcf_file, "w").close()
+
+    # Make a local copy of the index file, rename extension to avoid issue in bcftools.
+    local_file = os.path.basename(index_uri).replace(".csi", ".tbi")
+    cmd = ("cp", "/dev/stdin", local_file)
+    _, stderr = process_stream(index_uri, cmd)
+    if stderr:
+        raise RuntimeError(f"Failed to create index: {stderr}")
```
thetorpedodog, line 74: [`shutil.copyfileobj`](https://docs.python.org/3/library/shutil.html#shutil.copyfileobj) might be better for this situation.

thetorpedodog (commented):
A few more things but this is looking pretty good overall.

thetorpedodog:
Per request, I applied changes for this last round of minor comments. Give it a quick look to make sure everything looks good.

gspowley:
Looks good, thanks @thetorpedodog!

PR #420, opened 2023-06-09
Look at the *child* nodes when looking for local child nodes.

thetorpedodog:
To find whether to download results, we wanted to check whether either it was terminal or had any local children. But instead of looking at the children, we looked at the *parents*.


Shelnutt2 (approved):

ihnorton (approved):

thetorpedodog:
test failures are flakes on paths this untouched by this code.

PR #422, opened 2023-06-13
Filters quasi-serializing UDF passthrough

ktsitsi:
This PR:

- quasi-serializes the filter/compressor to be used during ingestion on the client side
- passes the serialized filter into the udf
- udf deserialises by instantiating filter object by name
- disables `chunked=True` that caused an `IndexError: tuple index out of range` in some images till further investigation of these failed images and the `tiifffile` package where it derives from.

This might be a hotfix for now (needed by Quest) till we standarize the way a user can choose to pass unserialisable tiledb objects into udfs.

### DIFF
--- a/src/tiledb/cloud/bioimg/ingestion.py
+++ b/src/tiledb/cloud/bioimg/ingestion.py
@@ -38,7 +38,6 @@ def ingest(
     def ingest_tiff_udf(
         io_uris: Sequence[Tuple],
         config: Mapping[str, Any],
-        workers: int,
         *args: Any,
         **kwargs,
     ):
@@ -47,10 +46,22 @@ def ingest(
 
         :param io_uris: Pairs of tiff input - output tdb uris
         :param config: dict configuration to pass on tiledb.VFS
-        :param workers: Number of threads that will spawn for parallelizing ingestion
         """
 
+        from tiledb import filter as compressors
         from tiledb.bioimg.converters.ome_tiff import OMETiffConverter
+
+        compressor = kwargs.get("compressor", None)
+        if compressor:
+            compressor_type = compressor.get("type", None)
+            if compressor_type:
+                kwargs["compressor"] = vars(compressors).get(compressor_type)(
+                    **compressor.get("attrs", {})
+                )
+            else:
+                raise ValueError
+        else:
+            kwargs["compressor"] = compressor
 
         conf = tiledb.Config(params=config)
         vfs = tiledb.VFS(config=conf)
@@ -58,9 +69,7 @@ def ingest(
         with tiledb.scope_ctx(ctx_or_config=conf):
             for input, output in io_uris:
                 with vfs.open(input) as src:
-                    OMETiffConverter.to_tiledb(
-                        src, output, *args, max_workers=workers, chunked=True, **kwargs
-                    )
+                    OMETiffConverter.to_tiledb(src, output, **kwargs)
 
     if isinstance(source, str):
         # Handle only lists
@@ -87,8 +96,8 @@ def ingest(
     samples = get_uris(source, output, config)
 
     # Build the task graph
-    dag_name = "batch-ingest-bioimg" if taskgraph_name is None else taskgraph_name
-    task_prefix = f"{dag_name}  - Batch Task"
+    dag_name = "bioimg-ingestion" if taskgraph_name is None else taskgraph_name
+    task_prefix = f"{dag_name} - Task"
 
     logger.info("Building graph")
     graph = tiledb.cloud.dag.DAG(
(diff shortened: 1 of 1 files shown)


ktsitsi:
> I think it'd be better to use an enum + values for data until.someone implements serialization in TileDB-Py. @ihnorton what do you think?
> 
> Whatever we release has to be supported for backwards compatibility.

I had the same first thought described in shortcut https://app.shortcut.com/tiledb-inc/story/30028/filters-not-serialisable-for-batch-ingestion, but this enum if designed should be able to include all the filters that probably have different args e.g `level vs quality` and would probably become verbose as a data structure, this was my only concern given the constrained time limit we have to deliver this.

src/tiledb/cloud/bioimg/ingestion.py
```diff
@@ -110,13 +125,19 @@ def batch(iterable, chunks):
...
             **kwargs,
         )
 
     graph.compute()
     return graph
 
 
+def serialize_filter(filter):
+    if isinstance(filter, tiledb.Filter):
+        return {"type": type(filter).__name__, "attrs": filter._attrs_()}
```
thetorpedodog, line 138: We could borrow [the filter specification syntax from TileDB-SOMA](https://github.com/single-cell-data/TileDB-SOMA/tree/1.2.5/apis/python#:~:text=interface%20TDBFilterSpec%20%7B%0A%20%20/**%20The%20name%20of%20the%20filter.%20*/%0A%20%20_name%3A%20string%3B%0A%20%20/**%20kwargs%20that%20are%20passed%20when%20constructing%20the%20filter.%20*/%0A%20%20%5Bkwarg%3A%20string%5D%3A%20any%3B%0A%7D), which looks like:

```json
{"_name": "SomeFilter", "param1": 1, "param2": 2}
```

which would call `SomeFilter(param1=1, param2=2)`

src/tiledb/cloud/bioimg/ingestion.py
```diff
@@ -47,20 +46,30 @@ def ingest_tiff_udf(
 
         :param io_uris: Pairs of tiff input - output tdb uris
         :param config: dict configuration to pass on tiledb.VFS
-        :param workers: Number of threads that will spawn for parallelizing ingestion
         """
 
+        from tiledb import filter as compressors
```
thetorpedodog, line 51: Is there a reason we call this `compressors` rather than using `filter`? (Maybe to avoid confusing it with some kind of image filter?) Would be good to leave as a comment.

ktsitsi: I was inclined to use it as compressor for the reason you mentioned. Filters are somehow keyword in image domain and since we use it mainly for compression and iirc we use these terms interchangeably I thought it would make sense. It is not hard requirement though.

ihnorton (commented):
> I think it'd be better to use an enum + values for data until.someone implements serialization in TileDB-Py. @ihnorton what do you think?

I believe this should be backward-compatible -- the JSON representation is only intermediate, not what the user is expected to pass AFAICT. We check here if the `compressor` arg is a filter:

https://github.com/TileDB-Inc/TileDB-Cloud-Py/pull/422/files#diff-199ecfa3c75505b8d0abc41cb7f2b37b77ec42da9889c38a234d21852740cab6R137

And then wrap that in JSON for internal usage. So the external use would look like:
```
ingest(...
    compressor=tiledb.WebPFilter(...)
```
And when serialization/pickling is available, we'll drop the intermediate JSON.

src/tiledb/cloud/bioimg/ingestion.py
ihnorton: Let's use the same argument name as we use in `from_bioimg`...

ihnorton (approved):
> So the external use would look like:
> ingest(...
>     compressor=tiledb.WebPFilter(...)

If possible, let's add a test that this actually works. As long as the backward compatibility story is solid, LGTM, modulo the minor nits inline.

src/tiledb/cloud/bioimg/ingestion.py
```diff
@@ -47,20 +46,34 @@ def ingest_tiff_udf(
...
         from tiledb.bioimg.converters.ome_tiff import OMETiffConverter
 
+        compressor = kwargs.get("compressor", None)
+        if compressor:
+            compressor_name = compressor.get("_name", None)
+            if compressor_name:
+                compressor_args = {
+                    k: None if not v else v
+                    for k, v in compressor.get("attrs", {}).items()
+                }
```
thetorpedodog, line 61: to match the below, this would be

```python
# Copy to not mutate the passed-in dict
compressor_args = dict(compressor)
compressor_name = compressor_args.pop("_name")
```
and use the existing code for the rest.

src/tiledb/cloud/bioimg/ingestion.py
```diff
@@ -47,20 +46,34 @@ def ingest_tiff_udf(
...
+                kwargs["compressor"] = vars(filter).get(compressor_name)(
+                    **compressor_args
+                )
+            else:
+                raise ValueError
+        else:
+            kwargs["compressor"] = compressor
```
thetorpedodog, line 68: Since there's not a compressor argument, does this need to be set?

thetorpedodog (approved):
Given we need this in, it’s good for now. There’s some workaround stuff that we can clean up later.

thetorpedodog:
The tests are flaking right now, and none of the code here even touches that, so I'm submitting.

PR #428, opened 2023-06-27
Add typing-extensions to dependencies

ihnorton:
Without this, I get the following in a [package](https://github.com/TileDB-Inc/TileDB-Vector-Search/blob/main/apis/python/pyproject.toml#L21) which depends on Cloud-Py:

```
>   from typing_extensions import Self
E   ModuleNotFoundError: No module named 'typing_extensions'
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,6 +17,7 @@ dependencies = [
   # tblib transitively enabled, so this is needed for unpickling.
   "tblib~=1.7",
   "tiledb>=0.15.2",
+  "typing-extensions",
   "urllib3>=1.26",
 ]
 


NikolaosPapailiou (approved):

thetorpedodog:
Could you remove change 4dd493ff49631d3371678622094e94b83db79c7e from this branch? I will do a separate change to fix that (and the current fix is breaking everything else: https://github.com/TileDB-Inc/TileDB-Cloud-Py/actions/runs/5390289327/workflow#L44-L54)

ihnorton:
> Could you remove change https://github.com/TileDB-Inc/TileDB-Cloud-Py/commit/4dd493ff49631d3371678622094e94b83db79c7e from this branch

Done, and thanks.

thetorpedodog (approved):

PR #430, opened 2023-07-03
Make 3 attempts to run Python tests before failing the build.

thetorpedodog:
On the best of days, automated tests that access the network can be flaky and prone to hiccups, and our tests here are no different. For this reason, we want to retry them to avoid the hassle of repeatedly manually retrying the tests waiting for them all to succeed. We use pytest to rerun only failed tests.

### DIFF
--- a/.github/scripts/retry
+++ b/.github/scripts/retry
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+# Retries the given command up to N times.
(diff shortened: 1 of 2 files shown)


Shelnutt2 (approved):

thetorpedodog:
[sc-31223]

PR #431, opened 2023-07-04
Hotfix of failing job

ktsitsi:
This PR:

- Fixes an always failing ingestion node. This node was being created by the `get_uri` function and was adding an extra pseudosample in the ingestion DAG. It was a result of `vfs.ls` of a `bucket` , which returns not only the bucket contents but the bucket root itself too.

### DIFF
--- a/src/tiledb/cloud/bioimg/ingestion.py
+++ b/src/tiledb/cloud/bioimg/ingestion.py
@@ -169,7 +169,7 @@ def get_uris(source: Sequence[str], output_dir: str, config: Mapping[str, Any]):
 
     if len(source) == 1 and vfs.is_dir(source[0]):
         # Folder like input
-        return tuple(iter_paths(vfs.ls(source[0])))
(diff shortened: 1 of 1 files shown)


src/tiledb/cloud/bioimg/ingestion.py
```diff
@@ -169,7 +169,7 @@ def iter_paths(sequence) -> Iterator[Tuple]:
 
     if len(source) == 1 and vfs.is_dir(source[0]):
         # Folder like input
-        return tuple(iter_paths(vfs.ls(source[0])))
+        return tuple(iter_paths(vfs.ls(source[0])[1:]))
```
Shelnutt2, line 172: Do you need a length check here? What if the bucket is empty?

ktsitsi: That is correct, thanks! We were assuming already that there were some images to be ingested. But let's catch this edge case too.

Shelnutt2 (approved):

PR #435, opened 2023-07-11
Set AWS_DEFAULT_REGION in set_aws_context

gspowley:
This PR resolves an issue related to running AWS CLI commands in a batch UDF, because the AWS CLI requires the AWS region to be specified.

Note: This PR does not resolve issues related to assuming a role with ARN or `access_credentials_name` in batch UDFs.


Shelnutt2 (approved):
Well follow-up with a default of the region the code is running in. LGTM

PR #437, opened 2023-07-12
Add require_action in cases where the client picks a namespace for the user

anastasop:
The intention it to validate that the namespace chosed has this action for example run_job for udfs. If not, it returns a descriptive error message.

Please read the discussion at https://app.shortcut.com/tiledb-inc/story/31330/confusing-error-when-user-has-not-set-default-namespace-to-charge before reviewing.

### DIFF
--- a/src/tiledb/cloud/client.py
+++ b/src/tiledb/cloud/client.py
@@ -148,9 +148,18 @@ def default_user() -> models.User:
     return config.user
 
 
-def default_charged_namespace() -> str:
-    """Returns the namespace :func:`default_user` charges to by default."""
-    return find_organization_or_user_for_default_charges(default_user())
+def default_charged_namespace(required_action: Optional[str] = None) -> str:
+    """Returns the namespace :func:`default_user` charges to by default.
+
+    If `required_action` is set then it checks amond the user
+    organizations to find the first one that support this action.
+
+    :param required_action: a namespace action, must be an enum
+        from rest_api.NamespaceActions
+    """
+    return find_organization_or_user_for_default_charges(
+        default_user(), required_action
+    )
 
 
 def list_public_arrays(
@@ -520,7 +529,10 @@ def organization(organization, async_req=False):
         raise tiledb_cloud_error.check_exc(exc) from None
 
 
-def find_organization_or_user_for_default_charges(user):
+def find_organization_or_user_for_default_charges(
+    user: models.User,
+    required_action: Optional[str] = None,
+) -> str:
     """
     Takes a user model and finds either the first non public organization
         or the user itself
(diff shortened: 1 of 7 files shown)


src/tiledb/cloud/client.py
```diff
@@ -148,9 +148,9 @@ def default_user() -> models.User:
     return config.user
 
 
-def default_charged_namespace() -> str:
+def default_charged_namespace(require_action = None) -> str:
```
thetorpedodog, line 151: Please add documentation and a type annotation for what `require_action` is. Also, would `required_action` (or just `action`) be a better name?

thetorpedodog (commented):
Also be sure to run `pre-commit` to ensure formatting and related stuff is correct.

anastasop:
@antalakas @thetorpedodog There is a first round of comments (addressed) but before we continue with the review, i would like an overall opinion if this PR is in the right direction or we don't actually need this

src/tiledb/cloud/client.py
```diff
@@ -148,9 +148,18 @@ def default_user() -> models.User:
     return config.user
 
 
-def default_charged_namespace() -> str:
-    """Returns the namespace :func:`default_user` charges to by default."""
-    return find_organization_or_user_for_default_charges(default_user())
+def default_charged_namespace(required_action=None) -> str:
```
thetorpedodog, line 151:
```suggestion
def default_charged_namespace(required_action: Optional[str] = None) -> str:
```

src/tiledb/cloud/client.py
```diff
@@ -148,9 +148,18 @@ def default_user() -> models.User:
...
+    """Returns the namespace :func:`default_user` charges to by default.
+
+    If `required_action` is set then it checks amond the user
+    organizations to find the first one that support this action.
+
+    :param str required_action: a namespace action, must be an enum
```
thetorpedodog, line 157:
```suggestion
    :param required_action: a namespace action, must be an enum
```

src/tiledb/cloud/client.py
```diff
@@ -520,7 +529,7 @@ def organization(organization, async_req=False):
         raise tiledb_cloud_error.check_exc(exc) from None
 
 
-def find_organization_or_user_for_default_charges(user):
+def find_organization_or_user_for_default_charges(user, required_action=None):
```
thetorpedodog, line 532:
```suggestion
def find_organization_or_user_for_default_charges(
    user: models.User,
    required_action: Optional[str] = None,
) -> str:
```

src/tiledb/cloud/client.py
```diff
@@ -529,20 +538,23 @@ def find_organization_or_user_for_default_charges(user):
...
+        if required_action is None or required_action in org.allowed_actions:
+            return org.organization_name
+
+    if required_action is None or required_action in user.allowed_actions:
+        return user.username
+
+    raise Exception(
+        f"user {user.username} does not belong to "
+        + f"an organization supporting {required_action}. "
+        + "Please check the default_namespace in the console"
```
thetorpedodog, line 556: Since adjacent strings are automatically joined together, the + signs are not needed here.

```suggestion
        f"an organization supporting {required_action}. "
        "Please check the default_namespace in the console"
```

thetorpedodog (approved):
I do actually very much like this change. There are a few small Python-specific things to pick up, which I left as suggestions here. (Don’t feel bad about any of these; it’s fine particularly for somebody who doesn’t work primarily in Python to have encyclopaedic knowledge of the language and its newer features.)

PR #440, opened 2023-07-13
Filtering with suffix and fixing root paths inclusion

ktsitsi:
This PR:

- Filters out any other file except for`.tiff`, `.tif` or (`.tdb` in case of exportation) from dir URIs given for ingestion or exportation.
- Fixes a previous mistake on unsorted-undefined order slicing of a bucket's `ls`.

Disclaimer: The next PR will concern a complete test suite for these part of codebase.

### DIFF
--- a/src/tiledb/cloud/bioimg/helpers.py
+++ b/src/tiledb/cloud/bioimg/helpers.py
@@ -58,7 +58,7 @@ def batch(iterable, chunks):
         yield iterable[ndx : min(ndx + chunks, length)]
 
 
-def scale_calc(source, num_batches):
+def scale_calc(samples, num_batches):
     """Calculate scaling settings for batch_size and max_workers
 
     :param source: The source iterable containing files to be ingested/exported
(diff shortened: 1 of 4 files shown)


src/tiledb/cloud/bioimg/helpers.py
```diff
@@ -16,20 +16,22 @@ def get_uris(
...
 
     def create_output_path(input_file, output_dir) -> str:
-        return os.path.join(output_dir, os.path.basename(input_file) + f".{output_ext}")
+        filename = os.path.splitext(os.path.basename(input_file))[0]
+        return os.path.join(output_dir, filename + f".{output_ext}")
 
     def iter_paths(sequence) -> Iterator[Tuple]:
         for uri in sequence:
-            yield uri, create_output_path(uri, output_dir)
+            if uri.endswith((".tiff", ".tif", ".tdb")):
```
Shelnutt2, line 24: Do we need a constant list? We support more than `tiff` right?

ktsitsi: We will for now the client calls only the UDF with the OMETIFF converter. We have already implemented an abstraction `from_bioimg` that includes all converters and will auto-recognize based on extension. For now the functionality is being constrained to the above.

Shelnutt2 (approved):

thetorpedodog (approved):

thetorpedodog:
Failures are transient server-side errors; merging to get this out to the people.

PR #441, opened 2023-07-13
Bioimg cloud tests for batch ingestion

ktsitsi:
This PR:

- Adds tests that cover all the untested modules of both the ingestion and exportation API for bioimg. They accompany and complete the tests for DAGs and the tiledb-bioimg library that serves as a backend in the UDFs.
- Changes miscs mainly syntax sugar and covers also a case where the source is a dir and scaling would not be calculated properly.

[sc-28612]

### DIFF
--- a/tests/test_bioimg.py
+++ b/tests/test_bioimg.py
@@ -1,0 +1,129 @@
+import math
+import unittest
+from unittest import mock
+
+from tiledb.cloud.bioimg.helpers import batch
+from tiledb.cloud.bioimg.helpers import get_uris
+from tiledb.cloud.bioimg.helpers import scale_calc
+from tiledb.vfs import VFS
+
+
+class BioimgTest(unittest.TestCase):
+    def setUp(self) -> None:
+        self.out_path = "out"
+        return super().setUp()
+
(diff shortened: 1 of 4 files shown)


tests/test_bioimg.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+class BioimgTest(unittest.TestCase):
+    def setUp(self) -> None:
+        self.out_path = "out"
+        return super().setUp()
+
+    def test_get_uris_ingestion(self):
+        ingest_uri_sample = {
+            "test1": ["test1.tiff", "test2.tiff", "test3.tiff"],
```
Shelnutt2, line 18: Where do these test files come from?

ktsitsi: These are not real test files. In these tests we assume that the `vfs.ls` returns a list of files that has this form. We want to test the functionality of the function we have introduced in our case the `get_uris` and we assume that `vfs.ls` will correctly return this list, or an empty one or a list that contains files except for `tiffs`.

thetorpedodog (approved):
Really nicely structured tests!

PR #444, opened 2023-07-18
Reinstate "Initial version of SOMA ingestor. (#392)"

thetorpedodog:
This reverts commit 3045ed9a5190a42a29b5c170078b147686369254, but removes the problematic change adding `tiledbsoma` as an optional dependency.


johnkerl (approved):
🚢

PR #454, opened 2023-08-17
Read VCF URIs from a metadata array

gspowley:
Provide an option to VCF ingestion that reads VCF URIs from a metadata array.

The VCF URIs should be stored in an attribute of the metadata array named `uri` or the name of the metadata attribute can be passed with the `metadata_attr` argument.

For example:
```python

# Read URIs from the `uri` attribute of the metadata array
vcf.ingest(dataset_uri=dataset_uri, metadata_uri=metadata_uri, ...)

# Read URIs from the `vcf_uri` attribute of the metadata array
vcf.ingest(dataset_uri=dataset_uri, metadata_uri=metadata_uri, metadata_attr="vcf_uri", ...)
```

### DIFF
--- a/src/tiledb/cloud/vcf/ingestion.py
+++ b/src/tiledb/cloud/vcf/ingestion.py
@@ -255,6 +255,44 @@ def read_uris_udf(
         return result
 
 
+def read_metadata_uris_udf(
+    dataset_uri: str,
+    *,
+    config: Optional[Mapping[str, Any]] = None,
+    metadata_uri: str,
+    metadata_attr: str = "uri",
+    max_files: Optional[int] = None,
+    verbose: bool = False,
+) -> Sequence[str]:
+    """
+    Read a list of URIs from a TileDB array. The URIs will be read from the
+    attribute specified in the `metadata_attr` argument.
+
+    :param dataset_uri: dataset URI
+    :param config: TileDB config, defaults to None
+    :param metadata_uri: metadata array URI
+    :param metadata_attr: name of metadata attribute containing URIs, defaults to "uri"
+    :param max_files: maximum number of URIs returned, defaults to None
+    :param verbose: verbose logging, defaults to False
+    :return: list of URIs
+    """
+    logger = get_logger_wrapper(verbose)
+
+    with tiledb.scope_ctx(config):
+        with Profiler(group_uri=dataset_uri, group_member=LOG_ARRAY) as prof:
+            with tiledb.open(metadata_uri) as A:
+                df = A.query(dims=[], attrs=[metadata_attr]).df[:]
+            results = df[metadata_attr].to_list()
+
+            if max_files:
+                results = results[:max_files]
+
+            logger.info("Read %d VCF URIs from the metadata array.", len(results))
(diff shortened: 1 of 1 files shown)


src/tiledb/cloud/vcf/ingestion.py
```diff
@@ -255,6 +255,44 @@ def read_uris_udf(
...
+    :param verbose: verbose logging, defaults to False
+    :return: list of URIs
+    """
+    logger = get_logger_wrapper(verbose)
+
+    with tiledb.scope_ctx(config):
+        with Profiler(group_uri=dataset_uri, group_member=LOG_ARRAY) as prof:
+            with tiledb.open(metadata_uri) as A:
+                df = A.query(dims=[], attrs=[metadata_attr]).df[:]
+            results = df[metadata_attr].to_list()
```
Shelnutt2, line 285: What about just doing `A.query(dims=[], atts=[metadata_attr])[:]` then you can avoid the `.to_list()` from pandas.

gspowley: It's a little more complicated. I pushed a change that avoids using pandas.

src/tiledb/cloud/vcf/ingestion.py
```diff
@@ -1004,6 +1046,18 @@ def ingest_manifest_dag(
...
+            read_metadata_uris_udf,
+            dataset_uri_result,
+            config=config,
+            metadata_uri=metadata_uri,
+            metadata_attr=metadata_attr,
+            verbose=verbose,
+            name="Read VCF URIs from metadata ",
+            **kwargs,
+        )
+
```
thetorpedodog, line 1060: seems like you want an `else` below this, otherwise you’ll be doing all this hard work just to immediately overwrite it

src/tiledb/cloud/vcf/ingestion.py
```diff
@@ -1398,14 +1456,20 @@ def ingest(
...
-        raise ValueError("Either `search_uri` or `sample_list_uri` must be provided.")
+    if not search_uri and not sample_list_uri and not metadata_uri:
+        raise ValueError(
+            "One of `search_uri`, `sample_list_uri`, or `metadata_uri`"
+            " must be provided."
+        )
 
-    if search_uri and sample_list_uri:
-        raise ValueError("Cannot specify both `search_uri` and `sample_list_uri`.")
+    if sum([bool(search_uri), bool(sample_list_uri), bool(metadata_uri)]) > 1:
```
thetorpedodog, line 1465: How about `== 1` and say “Exactly one of ...”?

src/tiledb/cloud/vcf/ingestion.py
gspowley: It works as expected. I guess reusing the `sample_uris` variable is confusing. I can improve that.

src/tiledb/cloud/vcf/ingestion.py
gspowley: Good idea, I can collapse 2 checks into one.

Shelnutt2 (approved):

PR #456, opened 2023-09-01
Support single file tiledb uris for exportation

ktsitsi:
This PR:

- allows a single bioimg group (multi-resolution image) to be exported by giving as an argument the tiledb uri.
- Since touching these files fixes also a bug in the taskgraph UI for when the number of batches is by default None

### DIFF
--- a/src/tiledb/cloud/bioimg/exportation.py
+++ b/src/tiledb/cloud/bioimg/exportation.py
@@ -85,7 +85,7 @@ def export(
             work,
             config,
             *args,
-            name=f"{task_prefix} - {i}/{num_batches}",
+            name=f"{task_prefix} - {i}",
             mode=tiledb.cloud.dag.Mode.BATCH,
             resources=DEFAULT_RESOURCES if resources is None else resources,
             image_name=DEFAULT_IMG_NAME,
--- a/src/tiledb/cloud/bioimg/helpers.py
+++ b/src/tiledb/cloud/bioimg/helpers.py
@@ -24,19 +24,25 @@ def get_uris(
             if uri.endswith((".tiff", ".tif", ".tdb")):
                 yield uri, create_output_path(uri, output_dir)
 
-    if len(source) == 1 and vfs.is_dir(source[0]):
-        # Check if the dir is actually a tiledb group for exportation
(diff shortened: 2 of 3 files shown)


src/tiledb/cloud/bioimg/helpers.py
```diff
@@ -24,19 +24,25 @@ def iter_paths(sequence) -> Iterator[Tuple]:
...
-            if tiledb.object_type(source[0]) != "group":
-                # Folder like input
-                contents = vfs.ls(source[0])
-                if len(contents) != 0:
-                    return tuple(iter_paths(contents))
+    if len(source) == 1:
+        if source[0].startswith("tiledb://"):
+            # Support tiledb uri single image
+            return ((source[0], create_output_path(source[0], output_dir)),)
+        elif vfs.is_dir(source[0]):
```
Shelnutt2, line 31: Instead of checking for `is_dir` why not check for the object type?

ktsitsi: I check for the object type 2 locs below that one, but I wanted to validate arguments. Normally I should have add an error. In case no tiledb uri neither a vfs dir is given as input. Imagine a random single s3 blob as input, this should fail. Now it is  a no-op but maybe should throw an error.

ktsitsi: Also we support the case that the input is a folder of bioimg groups. This is why we check if it is single dir and if so we iterate to find the contents.

src/tiledb/cloud/bioimg/exportation.py
```diff
@@ -85,7 +85,7 @@ def export_tiff_udf(
             work,
             config,
             *args,
-            name=f"{task_prefix} - {i}/{num_batches}",
+            name=f"{task_prefix} - {i}",
```
thetorpedodog, line 88: are these changes related?

thetorpedodog (approved):
one question but otherwise this looks reasonable.

src/tiledb/cloud/bioimg/exportation.py
ktsitsi: Nope as noted in PR description I just thought worth changing it since it is a misc bug.

PR #459, opened 2023-09-07
Avoid importing Pandas as long as possible.

thetorpedodog:
Because importing Pandas can take up to a half a second in some situations, we want to avoid importing it at all unless the user actually does something which requires it.


ihnorton (approved):
Is there a reason not to use the [`LazyLoader`](https://docs.python.org/3/library/importlib.html#importlib.util.LazyLoader) from `importlib`?

No objection to merging, more curious. LazyLoader seems slightly safer in the sense that `pandas` should always be a usable object, not requiring to remember `import pandas` in each individual function.

johnkerl:
@ihnorton my one worry about `LazyLoader` (or any `FancyThing` for that matter) is, does it work all the way back to Python 3.7 -- which we must still support. I saw to my relief though that `LazyLoader` is (a) built-in, and (b) goes back to 3.5 🎉

thetorpedodog:
> Is there a reason not to use the [`LazyLoader`](https://docs.python.org/3/library/importlib.html#importlib.util.LazyLoader) from `importlib`?
> 
> No objection to merging, more curious. LazyLoader seems slightly safer in the sense that `pandas` should always be a usable object, not requiring to remember `import pandas` in each individual function.

Looking at the [sample code they provide to implement this](https://docs.python.org/3/library/importlib.html#implementing-lazy-imports), it looks like more trouble than it’s worth, particularly since I’m not sure how it would interact with dev tools like autocompletion. Also, doing so would make it easier to accidentally *refer to* some member of the Pandas module and thus break the laziness in a harder-to-diagnose way. I would rather just have this low-magic solution, where it’s simply unavailable at the module scope (apart from type annotations), and attempting to say `pandas.anything` will immediately blow up at you.

johnkerl:
I'm all for low-magic solutions -- thanks @thetorpedodog ! :)

ihnorton:
+1 thanks!

PR #462, opened 2023-09-18
Dynamic distributed fragment consolidation

gspowley:
Add a distributed fragment consolidation function, which scales fragment consolidation to run on multiple TileDB Cloud nodes.

The fragment consolidation function is run on the VCF stats arrays (`variant_stats` and `allele_count`) at the end of VCF ingestion to improve read query performance of these arrays.

### DIFF
--- a/src/tiledb/cloud/utilities/consolidate.py
+++ b/src/tiledb/cloud/utilities/consolidate.py
@@ -56,13 +56,16 @@ def consolidate(
     fragments: Sequence[tiledb.FragmentInfo],
     *,
     config: Optional[Mapping[str, Any]] = None,
-):
+    max_fragment_size: int = MAX_FRAGMENT_SIZE,
+) -> None:
     """
     Consolidate fragments
 
     :param array_uri: array URI
     :param fragments: list of fragments
     :param config: config dictionary, defaults to None
+    :param max_fragment_size: max size of consolidated fragments,
+        defaults to MAX_FRAGMENT_SIZE
     """
 
     logger = get_logger()
@@ -70,7 +73,7 @@ def consolidate(
 
     config = tiledb.Config(config)
     config["sm.consolidation.mode"] = "fragments"
-    config["sm.consolidation.max_fragment_size"] = MAX_FRAGMENT_SIZE
+    config["sm.consolidation.max_fragment_size"] = max_fragment_size
 
     # Consolidate fragments.
     fragment_names = [basename(fi.uri) for fi in fragments]
@@ -85,7 +88,7 @@ def convac(
     *,
     config: Optional[Mapping[str, Any]] = None,
     vacuum_fragments: bool = False,
-):
+) -> None:
     """
     Consolidate and vacuum commits and fragment metadata, with an option to
     vacuum fragments as the first step.
@@ -143,6 +146,7 @@ def consolidate_fragments(
     dependencies: Optional[Sequence[dag.Node]] = None,
     consolidate_resources: Optional[Mapping[str, str]] = None,
     namespace: Optional[str] = None,
+    max_fragment_size: int = MAX_FRAGMENT_SIZE,
 ) -> None:
     """
     Consolidate fragments in an array.
@@ -165,6 +169,8 @@ def consolidate_fragments(
     :param dependencies: list of nodes in the graph to depend on, defaults to None
     :param consolidate_resources: resources for the consolidate node, defaults to None
     :param namespace: TileDB Cloud namespace, defaults to the user's default namespace
+    :param max_fragment_size: max size of consolidated fragments,
+        defaults to MAX_FRAGMENT_SIZE
     """
 
     if graph is None and dependencies is not None:
@@ -176,10 +182,10 @@ def consolidate_fragments(
     # function.
     if not graph_provided:
         graph = dag.DAG(
-            name="distributed-consolidation-plan",
+            name="distributed-consolidation",
             namespace=namespace,
             mode=dag.Mode.BATCH,
-            max_workers=100,
+            max_workers=40,
             retry_strategy=RetryStrategy(
                 limit=3,
                 retry_policy="Always",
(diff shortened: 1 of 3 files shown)


src/tiledb/cloud/utilities/consolidate.py
```diff
@@ -0,0 +1,248 @@
...
+        fis = tiledb.FragmentInfoList(array_uri)
+
+        for fi in fis:
+            if group_first_dim:
+                fragment_info_lists[fi.nonempty_domain[0]].append(fi)
+            else:
+                fragment_info_lists["all"].append(fi)
+
+        # Create a list of lists of fragment info objects.
+        results = [x for x in list(fragment_info_lists.values())]
```
thetorpedodog, line 46: This seems like it reifies the same thing into a list two times:

 1. `list(fragment_info_lists.values())`
 2. the no-op comprehension

It can be reduced to `results = list(fragment_info_lists.values())`.

(Did this maybe used to have some condition that got dropped?)

src/tiledb/cloud/utilities/consolidate.py
```diff
@@ -0,0 +1,248 @@
...
+
+        logger.info(f"{len(fis)} fragments grouped into {len(results)} groups")
+        logger.info(f"max memory usage: {max_memory_usage() / (1 << 30):.3f} GiB")
```
thetorpedodog, line 49: minor: rather than substituting things in on the logger string itself, instead use `%` placeholders (so that if the logging event does not happen, the string doesn't need to get constructed):

```python
logger.info("%d fragments grouped into %d groups", len(fis), len(results))
```

src/tiledb/cloud/utilities/consolidate.py
```diff
@@ -0,0 +1,248 @@
...
+    with tiledb.scope_ctx(config):
+        if vacuum_fragments:
+            try:
+                logger.info("Vacuuming fragments")
+                tiledb.vacuum(
+                    array_uri,
+                    config=tiledb.Config({"sm.vacuum.mode": "fragments"}),
+                )
+            except:
+                raise
```
thetorpedodog, line 112: this `try/except` is a do-nothing; was there intended to be more going on here?

src/tiledb/cloud/utilities/consolidate.py
```diff
@@ -0,0 +1,248 @@
...
+    """
+
+    if graph is None and dependencies is not None:
+        raise ValueError("Graph must be provided if dependencies are provided")
+
+    graph_provided = graph is not None
+
+    # If a graph is not provided, create a new graph and run it at the end of this
+    # function.
+    if not graph_provided:
```
thetorpedodog, line 183: maybe make it `graph_omitted` so that you can use it in `if graph_omitted and dependencies is not None` to avoid some minor repetition.

src/tiledb/cloud/utilities/consolidate.py
```diff
@@ -0,0 +1,248 @@
...
+
+    # Consolidate fragments.
+    fragment_names = [basename(fi.uri) for fi in fragments]
+    with tiledb.open(array_uri, "w", config=config) as array:
+        array.consolidate(fragment_uris=fragment_names)
+
+    logger.info(f"max memory usage: {max_memory_usage() / (1 << 30):.3f} GiB")
+
+
+def convac(
```
thetorpedodog, line 86: I’m going back and forth on whether I like this name. On the one hand, it is snappy; on the other, it is maybe a bit *too* short, and could be a bit opaque until you already know what it does. That said, it is described in the first three words of the docstring. Not a hard request, just something to think about. If you want to keep it, it’s absolutely fine; I leave it up to your judgement.

src/tiledb/cloud/utilities/consolidate.py
gspowley: Yes, this line went through a few changes. I'll simplify.

src/tiledb/cloud/utilities/consolidate.py
gspowley: Thanks, another case of reusing code that can be simplified.

src/tiledb/cloud/utilities/consolidate.py
gspowley: I feel the same, the name is a little too short and not descriptive. I'll pick a different name.

src/tiledb/cloud/utilities/consolidate.py
```diff
@@ -0,0 +1,239 @@
...
+    config: Optional[Mapping[str, Any]] = None,
+    group_first_dim: bool = True,
+) -> Sequence[Sequence[tiledb.FragmentInfo]]:
+    """
+    Get a list of fragment info objects, optionally grouping fragments that have the
+    same value for the first dimension.
+
+    :param array_uri: array URI
+    :param config: config dictionary, defaults to None
+    :param group_first_dim: group by first dimension, defaults to True
```
thetorpedodog, line 27: `group_by_first_dim`, maybe?

src/tiledb/cloud/utilities/consolidate.py
```diff
@@ -0,0 +1,239 @@
...
+from typing import Any, Mapping, Optional, Sequence
+
+import tiledb
+from tiledb.cloud import dag
+from tiledb.cloud.rest_api.models import RetryStrategy
+from tiledb.cloud.utilities import get_logger
+from tiledb.cloud.utilities import max_memory_usage
+from tiledb.cloud.utilities import run_dag
+
+MAX_FRAGMENT_SIZE = 1 << 30  # bytes
```
thetorpedodog, line 12: I would just add `_BYTES` to the end of this so that the unit is always right there in the name.

src/tiledb/cloud/utilities/consolidate.py
```diff
@@ -0,0 +1,239 @@
...
+    """
+
+    graph_omitted = graph is None
+
+    if graph_omitted and dependencies is not None:
+        raise ValueError("Graph must be provided if dependencies are provided")
+
+    # If a graph is not provided, create a new graph and run it at the end of this
+    # function.
+    if graph_omitted:
```
thetorpedodog, line 174: maybe fold these up?

```python
if graph_omitted:
  if dependencies is not None:
    raise ...
  graph = ...
```

thetorpedodog (approved):
A few minor suggestions but otherwise looks good.

PR #463, opened 2023-09-22
Reorganize UDF data codecs to prepare for better serialization.

thetorpedodog:
To better separate concerns and eliminate cross-package circular dependencies, this change moves the main handling of the `tiledb_json` format to the `_results` package, and restructures the way codecs are defined and handled.

This will allow us to introduce `tiledb_json` as a format for serializing UDF results, which will allow us to use the `tiledb_json` format to serialize Pandas dataframes, with custom encoding/decoding logic based on Arrow, instead of using native pickles.  This will allow us to offer client/server compatibility across Pandas versions, without the hacks we currently resort to.

---

First step in [sc-34367]. [skip-ch]


thetorpedodog:
NOTE TO SELF: This may also require modifying the internals of the UDF runtime due to moved imports.

antalakas (approved):

PR #465, opened 2023-09-26
Add and test codec for encoding Pandas via Arrow.

thetorpedodog:
We currently serialize Pandas data using ordinary Python Pickles, but doing so across Python versions and Pandas versions is extremely unsupported.  We also would like to fully decouple the version of Pandas between the client and the server, so that users who need Pandas 1.x can use it, but users who need Pandas 2.0 can use it in UDFs, with no added effort on either of their parts.

Fortunately, Apache Arrow already supports serializing Pandas data in a cross-compatible format. Adding this new codec type will allow us to store Pandas DataFrames in the "TileDB JSON" format, with encoding and decoding being transparent to the user.

This change only adds the decoder; we need to wait for a release of this to enable encoding it, so that we're sure that we can decode it when we actually start *sending* Pandas DataFrames via TileDB JSON.


antalakas (approved):

src/tiledb/cloud/_results/codecs.py
```diff
@@ -113,6 +116,25 @@ def decode(cls, data: bytes) -> object:
...
+
+    NAME = "arrow_dataframe"
+    # We don't actually serve this MIME type; it is subject to change.
+    MIME = "application/vnd.tiledb.arrow-dataframe"
+
+    @classmethod
+    def encode(cls, data: "pandas.DataFrame") -> bytes:
+        """Converts a Pandas array to the byte format of an Arrow table."""
+        tbl = pyarrow.Table.from_pandas(data)
+        return ArrowCodec.encode(tbl)
```
ihnorton, line 130: IMO it would be worth catching `ArrowInvalid` here and providing a contextual rethrow, because a user may have no idea why such an error could occur when dealing with a pandas object.

```
import tiledb, numpy as np, pandas as pd, tiledb.cloud
import pyarrow as pa

class W:
    def __init__(self, v):
        self.o = v

df = pd.DataFrame({"foo": pd.Series([W("bar")])})
df
pa.Table.from_pandas(df)
```
->
```
ArrowInvalid: ('Could not convert <__main__.W object at 0x1157cc160> with type W: did not recognize Python value type when inferring an Arrow data type', 'Conversion failed for column foo with type object')
```

ihnorton (changes requested):
Thanks! Minor change request inline.

Also, is the fallback we discussed for pandas < 2.0 already implemented? I don't think there are any specific changes related to that in this PR, unless it is implicit in the codec order.

ihnorton (approved):

PR #468, opened 2023-10-04
Enable using the Pandas-as-Arrow codec for DataFrames.

thetorpedodog:
Because we now have the *decoder* available at both ends, we can now safely enable encoding Pandas dataframes in the Arrow format.  If it is not possible to encode a DataFrame to Arrow, we fall back to Pickle.

---

Submitting this will need to wait until v0.10.23 has been deployed to the UDF environment.

### DIFF
--- a/src/tiledb/cloud/_results/codecs.py
+++ b/src/tiledb/cloud/_results/codecs.py
@@ -1,13 +1,14 @@
 import abc
 import base64
 import json
+import sys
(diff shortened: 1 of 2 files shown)


ihnorton (approved):
Thanks! 💯

thetorpedodog:
[sc-34367] [skip-ch]

PR #470, opened 2023-10-05
Relative group members in VCF ingestion

gspowley:
Add manifest and log array as relative members of the VCF dataset to enable portability.


thetorpedodog (approved):
Sorry for missing this! It looks good.

PR #472, opened 2023-10-10
Fix check for `resources` on client-side task graphs.

thetorpedodog:
Because of the way parameters were handled within the `dag.Node` class, on a client-side task graph (i.e., with `Mode.REALTIME`), we would never check to ensure that the unsupported `resources` parameter was not set. This change adds that check, as well as checks to locally-executing nodes (where resources cannot be specified at all).

[sc-35354]


Shelnutt2 (approved):
lgtm

PR #473, opened 2023-10-12
Local ingestion and use of run_dag

ktsitsi:
This PR:

- introduces a boolean flag `local= False` for ingesting a bioimg in local mode.
- substitutes the `graph.compute` function with it's wrapper `run_dag` from the `utilities._common` package

### DIFF
--- a/src/tiledb/cloud/bioimg/ingestion.py
+++ b/src/tiledb/cloud/bioimg/ingestion.py
@@ -23,6 +23,7 @@ def ingest(
     threads: Optional[int] = 8,
     resources: Optional[Mapping[str, Any]] = None,
     compute: bool = True,
+    local: bool = False,
     namespace: Optional[str],
     verbose: bool = False,
     exclude_metadata: bool = False,
@@ -150,6 +151,8 @@ def ingest(
                         verbose=verbose,
                         **kwargs,
                     )
+
+    submit = graph.submit_local if local else graph.submit
 
     if isinstance(source, str):
         # Handle only lists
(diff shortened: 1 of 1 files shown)


src/tiledb/cloud/bioimg/ingestion.py
```diff
@@ -92,35 +95,36 @@ def ingest_tiff_udf(
...
     logger.info("Building graph")
-    graph = tiledb.cloud.dag.DAG(
+    graph = dag.DAG(
         name=dag_name,
-        mode=tiledb.cloud.dag.Mode.BATCH,
+        mode=dag.Mode.BATCH,
         max_workers=max_workers,
         namespace=namespace,
     )
+    submit = graph.submit_local if local else graph.submit
```
thetorpedodog, line 104: If you’re running a graph server-side, then `submit_local` will never work; you will have to configure that also.

ktsitsi: What do you mean by server-side? This functionality is introduced the same way we use it here https://github.com/TileDB-Inc/TileDB-Cloud-Py/blob/6ca6404280758982606546df11f633a7494e8c5f/src/tiledb/cloud/vcf/ingestion.py#L920

thetorpedodog: "batch mode", i.e. the coordinator which runs the graph is on the server, rather than on the client. a key difference in that code is:

https://github.com/TileDB-Inc/TileDB-Cloud-Py/blob/6ca6404280758982606546df11f633a7494e8c5f/src/tiledb/cloud/vcf/ingestion.py#L910-L920

ihnorton (approved):
LGTM, thanks for the update.

PR #474, opened 2023-10-12
Kt/sc 30105/logger for bioimg

ktsitsi:
This PR:

- Complements that PR in [TileDB-Bioimaging](https://github.com/TileDB-Inc/TileDB-BioImaging/pull/105)
- introduces a variable `verbose=False` that will initialize a DEBUG mode for the logger.
- The logger is being initialised by the same logger wrapper that VCF uses although we could discuss the merge of the two. I duplicate the code here cause I don't want to import a function in bioimg from `vcf.ingestion.py`. If agreed I could merge and move all of them under the same `get_logger` in `utilities._common`


thetorpedodog (changes requested):
Actually, could you separate the "single-file TileDB URIs" part of this from the logging part of this? I have a few questions about the URI change but the logging part looks OK.

ktsitsi:
> Actually, could you separate the "single-file TileDB URIs" part of this from the logging part of this? I have a few questions about the URI change but the logging part looks OK.

I think this is the rebased code. This part of the code was introduced, reviewed and merged yesterday here #456 . However lets discuss in case you mean a rollback of this PR.

thetorpedodog (approved):

PR #476, opened 2023-10-17
Request UDF results as `tiledb_json` format by default.

thetorpedodog:
This is the final piece of the Pandas cross-compatibility puzzle: we request the results of UDFs and task graphs as `tiledb_json`–encoded values by default.  This means that Pandas dataframes within results will be returned as Arrow and transparently decoded back to Pandas for the user, along with all the other features TileDB JSON encoding gives us.

There are some slight changes elsewhere:

- Tests which expect serialized tuples now get lists.
- We auto-detect the result format when fetching results by default.

---

[sc-34367]


ihnorton (approved):
LGTM, but let's bump the version to 0.11 in order to highlight the change that tuple return values will now be returned as lists (also cc @Shelnutt2 FYI).

PR #482, opened 2023-11-09
Add option for embedding generation upon BioImage Ingestion 

ktsitsi:
This PR:

- Introduces an extra functionality inside the ingestion process of biomedical images. The new ingestion function on the client side as well as the UDF has been enhanced with extra arguments that enable the creation of image embeddings during the ingestion
- The newly added ingestion arguments are the following: 
```python
    :param embedding_model: The model to be used for creating embedding. Supported values are of type class EMBEDDINGS
    :param embedding_level: The resolution level to be used for the embedding. This could be different from the ingestion level
    selected with parameter `level`
    :param embedding_grid: A tuple that represents the (num_of_rows, num_of_cols), in which the image will be split in patches
    for the embedding creation. According to this grid internally the image is being split to fit this requirement.
``` 
- Currently the supported model is RESNET for the embeddings, but with the introduced `types` abstraction we are able to enhance this functionality in the future.
- Another type `SupportedExtensions` has been added concerning the file extensions we support with our ingestion function.
- This additional functionality changes the Data model specification we have set and thus updates the stored image groups to `fmt_version = 3` 

The resulted image is shown below:
Screenshot 2023-11-09 at 3 45 44 PM

Building a local docker image of the UDFs I was able to test it and the results are shown below as well:
Screenshot 2023-11-09 at 3 45 33 PM

This PR should be followed by PR's that will do the following and will be linked upon creation here:
- Update the `python-udf-imaging` dockerfile in TileDB-REST-UDF-DOCKER-IMAGES with the dependencies needed as `tiledb-vector-search` and `tensorflow`
- Also update the tiledb-cloud version in UDFs to match the one after the release of this PR, so functions and modules are being found on the UDF side


src/tiledb/cloud/bioimg/__init__.py
```diff
@@ -1,4 +1,19 @@
-from tiledb.cloud.bioimg.exportation import export
-from tiledb.cloud.bioimg.ingestion import ingest
+from .exportation import export
+from .helpers import batch
+from .helpers import get_embeddings_uris
+from .helpers import scale_calc
+from .helpers import serialize_filter
```
thetorpedodog, line 5: these helpers don’t seem like they would be public-facing.

src/tiledb/cloud/bioimg/helpers.py
```diff
@@ -4,6 +4,18 @@
 
 import tiledb
 
+from .types import SupportedExtensions
+
+
+def get_embeddings_uris(output_file_uri: str) -> Tuple[str, str]:
+    destination = os.path.dirname(output_file_uri)
+    filename = os.path.basename(output_file_uri).split(".")
```
thetorpedodog, line 12: check out `os.path.splitext`

src/tiledb/cloud/bioimg/ingestion.py
```diff
@@ -8,10 +8,185 @@
...
+                                )
+                            )
+                    return patches
+
+                with TileDBOpenSlide(output) as image:
+                    # Filter out patches with [arg] % non-blank coverage
+                    filtered = split_in_patches(image, embedding_level, embedding_grid)
+
+                    patches_array = np.array([])
+                    for patch in filtered:
```
ihnorton, line 142: (also seems slow..)

src/tiledb/cloud/bioimg/ingestion.py
```diff
@@ -8,10 +8,185 @@
...
+                        embedding_level
+                    ]
+                    # Calculate the region size based on the desired
+                    # number of rows and columns
+                    grid_row_num, grid_col_num = embedding_grid
+                    region_height = level_shape_h // grid_row_num
+                    region_width = level_shape_w // grid_col_num
+                    # Loop through the image and extract each region
+                    patches = []
+                    for i in range(grid_row_num):
```
ihnorton, line 125: This seems like it is going to be very slow?

src/tiledb/cloud/bioimg/ingestion.py
```diff
@@ -8,10 +8,185 @@
...
+
+                    # Create embedding of the image given its patches
+                    embeddings = calculate_model(patches_array, embedding_model)
+                    # Create flat index of the embedding for faster vector search
+                    tmp_features_file = "/tmp/features"
+                    with open(tmp_features_file, "wb") as f:
+                        np.array(embeddings.shape, dtype="uint32").tofile(f)
+                        np.array(embeddings).astype("float32").tofile(f)
+
+                    vs.ingest(
```
NikolaosPapailiou, line 159: Can you explain a bit the use case that you want to support?

From what I understand this is creating one vector index per image, storing the embedding of different patches of the image. Does this mean that you want to find similar patches within the image? Do you want to do any cross image queries?

src/tiledb/cloud/bioimg/ingestion.py
```diff
@@ -8,10 +8,185 @@
...
+                        index_type="FLAT",
+                        index_uri=embeddings_flat_uri,
```
NikolaosPapailiou, line 161: You also need to pass `source_uri` of the input vectors (`tmp_features_file`)

src/tiledb/cloud/bioimg/ingestion.py
ktsitsi: The idea here is to find similar patches across multiple images. We don't have any specific customer request for any of that. The idea behind the patches though is that in this data environment is quite unlikely to find similarities across whole images. Each image is quite different from one another as a whole based on the depicted cell. Also from my knowledge due to the blank space outside the cells captured by sensor it is quite possible to introduce a lot of noise in the model if fed to it as a whole. So my thought was that a future user would like to search for similarities across images of a specific region of the query image, or in a next step to be able to query a region from the viewer and find similar "abnormalities" - that he can justify and I can't- across the images.

PR #484, opened 2023-11-14
Separating supported extensions from #482

ktsitsi:
This PR:

- Splits the #482, which can be reviewed and visited later, to promote the SupportExtensions data type, in order to be able to ingest files that might have different extensions but they are supported by our converters.

### DIFF
--- a/src/tiledb/cloud/bioimg/__init__.py
+++ b/src/tiledb/cloud/bioimg/__init__.py
@@ -1,6 +1,4 @@
 from tiledb.cloud.bioimg.exportation import export
 from tiledb.cloud.bioimg.ingestion import ingest
 
-from .types import SupportedExtensions
-
-__all__ = ("ingest", "export", "SupportedExtensions")
+__all__ = ("ingest", "export")
--- a/src/tiledb/cloud/bioimg/__init__.py
+++ b/src/tiledb/cloud/bioimg/__init__.py
@@ -1,4 +1,6 @@
 from tiledb.cloud.bioimg.exportation import export
 from tiledb.cloud.bioimg.ingestion import ingest
 
-__all__ = ("ingest", "export")
+from .types import SupportedExtensions
+
+__all__ = ("ingest", "export", "SupportedExtensions")
--- a/src/tiledb/cloud/bioimg/helpers.py
+++ b/src/tiledb/cloud/bioimg/helpers.py
@@ -29,6 +29,7 @@ def get_logger_wrapper(
 
     return logger
 
+
 _SUPPORTED_EXTENSIONS = (".tiff", ".tif", ".svs", ".tdb")
 
 
--- a/src/tiledb/cloud/bioimg/helpers.py
+++ b/src/tiledb/cloud/bioimg/helpers.py
@@ -49,7 +49,7 @@ def get_uris(
 
     def iter_paths(sequence) -> Iterator[Tuple]:
         for uri in sequence:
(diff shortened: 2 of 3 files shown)


src/tiledb/cloud/bioimg/helpers.py
```diff
@@ -21,7 +23,7 @@ def create_output_path(input_file, output_dir) -> str:
 
     def iter_paths(sequence) -> Iterator[Tuple]:
         for uri in sequence:
-            if uri.endswith((".tiff", ".tif", ".tdb")):
+            if uri.endswith(tuple([ext.value for ext in SupportedExtensions])):
```
thetorpedodog, line 26: nit: [] not needed here (i.e., just `tuple(ext.value ...)`). Also I must have missed this last time I reviewed—I didn’t realize you could do `.endswith(('multiple', 'things'))`. Neat!

src/tiledb/cloud/bioimg/types.py
```diff
@@ -0,0 +1,8 @@
+import enum
+
+
+class SupportedExtensions(enum.Enum):
+    TIFF: str = ".tiff"
+    TIF: str = ".tif"
+    SVS: str = ".svs"
+    TDB: str = ".tdb"
```
thetorpedodog, line 8: This seems like it could be a module-level variable in helpers.py, just as a tuple.

src/tiledb/cloud/bioimg/__init__.py
```diff
@@ -1,4 +1,6 @@
 from tiledb.cloud.bioimg.exportation import export
 from tiledb.cloud.bioimg.ingestion import ingest
 
-__all__ = ("ingest", "export")
+from .types import SupportedExtensions
```
thetorpedodog, line 4: do we ever expect users to use this? it seems like this doesn't really need to be exported.

src/tiledb/cloud/bioimg/types.py
ktsitsi: Indeed. However in a next PR I am introducing also another type and I thought it would better organised this way. I can change of course.

src/tiledb/cloud/bioimg/__init__.py
ktsitsi: Not really nice catch.

src/tiledb/cloud/bioimg/types.py
thetorpedodog: I would say using an enum is useful for situations where you want to provide a configuration parameter, but it looks like in this case you’re just using it to list out suffixes. In that case, I think just using a tuple like `_SUPPORTED_EXTENSIONS = (".tiff", ".tif", ".svs", ".tdb")` makes more sense.

ktsitsi:
@thetorpedodog Can you take a last look and merge?

src/tiledb/cloud/bioimg/helpers.py
```diff
@@ -46,7 +49,7 @@ def create_output_path(input_file, output_dir) -> str:
 
     def iter_paths(sequence) -> Iterator[Tuple]:
         for uri in sequence:
-            if uri.endswith((".tiff", ".tif", ".tdb")):
+            if uri.endswith(tuple(ext for ext in _SUPPORTED_EXTENSIONS)):
```
thetorpedodog, line 52: this can just be `_SUPPORTED_EXTENSIONS` now since it’s already a tuple!

thetorpedodog (approved):

PR #485, opened 2023-11-15
Pass access_credentials_name to batch UDF functions when present.

gspowley:
This PR allows `access_credentials_name` to be passed to batch UDF functions. 

The goal is to allow a batch UDF function running with `access_credentials_name` to create a new DAG and submit batch UDF functions with the same `access_credentials_name`.


gspowley:
Closing this PR since passing or not passing `access_credentials_name` to arbitrary UDFs is more complicated.

PR #487, opened 2023-11-22
Add xarray as test dependency, partial for sc-37280

ihnorton:
Seeing this in [the tests](https://github.com/TileDB-Inc/TileDB-Cloud-Py/actions/workflows/tiledb-cloud-py.yaml)
```
    raise ImportError(msg)
E   ImportError: Missing optional dependency 'xarray'.  Use pip or conda to install xarray.
```

### DIFF
--- a/.github/workflows/tiledb-cloud-py.yaml
+++ b/.github/workflows/tiledb-cloud-py.yaml
@@ -97,7 +97,7 @@ jobs:
 
       - name: Install TileDB-Cloud-Py
         run: |
-          pip install .
+          pip install .[tests]
 
       - name: Run tests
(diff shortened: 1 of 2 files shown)


thetorpedodog (approved):
I was confused as to why this would be different from the frozen-deps tests, which work. The difference is in the code which generates the `requirements.txt` files we use for the frozen-deps tests:

https://github.com/TileDB-Inc/TileDB-Cloud-Py/blob/9e4eb5de18dc75be7c30b3d08779c042b966428b/update_requirements.sh#L12-L14

If you want to, you can change that file to `pip install .[tests]`, or just submit this as-is.

thetorpedodog:
companion: #488

PR #490, opened 2023-11-30
Expose alternate graph return value on ingest

johnkerl:
This maintains compatibility with the one-click API. Additionally, it allows the caller to get the graph object itself so they can do `grf.compute()` and `grf.wait()` on it, and then directly obtain resultst from the computation.

The alternative is `dag.server_logs` and `tasks.fetch_result`, which are cumbersome at best for users who have synchronous needs.

This pattern will be user in the collection-mapper logic, which _absolutely must_ provide a usable synchronous entrypoint.

### DIFF
--- a/src/tiledb/cloud/soma/ingest.py
+++ b/src/tiledb/cloud/soma/ingest.py
@@ -59,7 +59,7 @@ def run_ingest_workflow(
         manage execution and monitor progress.
     """
 
-    grf = get_ingest_workflow_graph(
+    grf = build_ingest_workflow_graph(
         output_uri=output_uri,
         input_uri=input_uri,
         measurement_name=measurement_name,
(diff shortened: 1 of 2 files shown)


johnkerl:
[sc-37815]

johnkerl:
@thetorpedodog ping 🙏

src/tiledb/cloud/soma/ingest.py
```diff
@@ -143,6 +59,43 @@ def run_ingest_workflow(
...
+        access_credentials_name=access_credentials_name,
+    )
+    grf.compute()
+    return {
+        "status": "started",
+        "graph_id": str(grf.server_graph_uuid),
+    }
+
+
+def get_ingest_workflow_graph(
```
thetorpedodog, line 81: maybe `build_ingest_workflow_graph` would be a better name.

thetorpedodog (approved):

johnkerl:
Thanks @thetorpedodog !

PR #491, opened 2023-12-07
Support Delayed nodes in the VCF distributed query

gspowley:
Update the VCF distributed query to allow passing the `samples` and `regions` args as `Delayed` or `DelayedSQL` nodes.

### DIFF
--- a/src/tiledb/cloud/vcf/query.py
+++ b/src/tiledb/cloud/vcf/query.py
@@ -10,6 +10,8 @@ import pyarrow as pa
 
 import tiledb.cloud
 from tiledb.cloud.compute import Delayed
+from tiledb.cloud.compute import DelayedArrayUDF
+from tiledb.cloud.compute import DelayedMultiArrayUDF
 from tiledb.cloud.compute import DelayedSQL
 from tiledb.cloud.utilities import Profiler
 from tiledb.cloud.utilities import get_logger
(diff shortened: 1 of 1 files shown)


src/tiledb/cloud/vcf/query.py
```diff
@@ -227,11 +234,11 @@ def build_read_dag(
     *,
     config: Optional[Mapping[str, Any]] = None,
     attrs: Optional[Union[Sequence[str], str]] = None,
-    regions: Optional[Union[Sequence[str], str]] = None,
+    regions: Optional[Union[Sequence[str], str, Delayed, DelayedSQL]] = None,
```
Shelnutt2, line 237: Perhaps also check for `DelayedArrayUDF` and `DelayedMultiArrayUDF`?

src/tiledb/cloud/vcf/query.py
```diff
@@ -344,11 +362,11 @@ def read(
...
     bed_file: Optional[str] = None,
     num_region_partitions: int = 1,
     max_workers: int = MAX_WORKERS,
-    samples: Optional[Union[Sequence[str], str]] = None,
+    samples: Optional[Union[Sequence[str], str, Delayed, DelayedSQL]] = None,
```
Shelnutt2, line 369: Perhaps also check for `DelayedArrayUDF` and `DelayedMultiArrayUDF`?

PR #492, opened 2023-12-08
Refactoring and moving io part in UDF lister

ktsitsi:
This PR:

- Moves the i/o related tasks in the server-side task graph of the code following #412
- Refactors the `get_uri` function to a more resilient, readable and maintainable form
- The newly introduced `iter_paths` allows recursive exploration and generalises the way we look up files to be ingested under the given source.
- Removes unused tests
- Deprecates/Removes some helper functions
- Introduces some API arguments to the Exportation API to sync it with the Ingestion one

### DIFF
--- a/src/tiledb/cloud/bioimg/exportation.py
+++ b/src/tiledb/cloud/bioimg/exportation.py
@@ -29,7 +29,7 @@ def build_io_uris_exportation(source: Sequence[str], output_dir: str, output_ext
             if vfs.is_dir(uri) and tiledb.object_type(uri) != "group":
                 # Folder for exploration
                 contents = vfs.ls(uri)
-                yield from tuple(iter_paths(contents))
+                yield from tuple(iter_paths(contents)[1:])
             elif tiledb.object_type(uri) == "group":
                 # For exportation we require the source path to be a tiledb group
                 yield uri, create_output_path(uri, output_dir)
--- a/src/tiledb/cloud/bioimg/exportation.py
+++ b/src/tiledb/cloud/bioimg/exportation.py
@@ -1,14 +1,55 @@
-from typing import Any, Dict, Mapping, Optional, Sequence, Tuple, Union
+import os
+from typing import Any, Dict, Iterator, Mapping, Optional, Sequence, Tuple, Union
 
 import tiledb
-from tiledb.cloud._common.utils import logger
-from tiledb.cloud.bioimg.helpers import batch
-from tiledb.cloud.bioimg.helpers import get_uris
-from tiledb.cloud.bioimg.helpers import scale_calc
+from tiledb.cloud import dag
+from tiledb.cloud.bioimg.helpers import get_logger_wrapper
+from tiledb.cloud.utilities._common import run_dag
 
 DEFAULT_RESOURCES = {"cpu": "8", "memory": "4Gi"}
 DEFAULT_IMG_NAME = "3.9-imaging-dev"
 DEFAULT_DAG_NAME = "bioimg-exportation"
+
+
+def build_io_uris_exportation(source: Sequence[str], output_dir: str, output_ext: str):
+    """Match input uri/s with output destinations
+
+    :param source: A sequence of paths or path to input
+    :param output_dir: A path to the output directory
+    """
+    vfs = tiledb.VFS()
+
+    def create_output_path(input_file, output_dir) -> str:
+        filename = os.path.splitext(os.path.basename(input_file))[0]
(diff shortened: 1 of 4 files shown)


src/tiledb/cloud/bioimg/exportation.py
```diff
@@ -20,14 +61,17 @@ def export(
     num_batches: Optional[int] = None,
     resources: Optional[Mapping[str, Any]] = None,
     compute: bool = True,
+    local: bool = False,
     namespace: Optional[str] = None,
+    verbose: bool = False,
+    output_ext: str = "tiff",
```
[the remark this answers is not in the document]
ktsitsi: We use as a default value in the UDF the `Converters.OMETIFF`, so I thought it would make sense.

```python
to_bioimg(
    input,
    output,
    converter=Converters.OMETIFF,
    config=src_cfg,
    output_config=dest_cfg,
    verbose=verbose,
    **kwargs,
    )
```

src/tiledb/cloud/bioimg/exportation.py
```diff
@@ -20,14 +61,17 @@ def export(
     num_batches: Optional[int] = None,
     resources: Optional[Mapping[str, Any]] = None,
     compute: bool = True,
+    local: bool = False,
```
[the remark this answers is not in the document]
ktsitsi: I tried to use the same naming as https://github.com/TileDB-Inc/TileDB-Cloud-Py/blob/6ca6404280758982606546df11f633a7494e8c5f/src/tiledb/cloud/vcf/ingestion.py#L30C21-L30C21 

There was a follow-up discussion on this in another open ticket https://github.com/TileDB-Inc/TileDB-Cloud-Py/pull/473#discussion_r1392751081

What would you recommend as an alternative?

src/tiledb/cloud/bioimg/exportation.py
```diff
@@ -51,49 +100,69 @@ def export_tiff_udf(
...
         :param config: dict configuration to pass on tiledb.VFS
         """
+        from tiledb.bioimg import Converters
+        from tiledb.bioimg import to_bioimg
+        from tiledb.ctx import default_ctx
 
-        from tiledb.bioimg.converters.ome_tiff import OMETiffConverter
+        # if writer config not given assume same as source
+        src_cfg = default_ctx().config()
+        dest_cfg = src_cfg if not config else config
```
thetorpedodog, line 109: `dest_cfg = config or src_cfg`

src/tiledb/cloud/bioimg/exportation.py
```diff
@@ -51,49 +100,69 @@ def export_tiff_udf(
...
+
+    graph.submit(
+        export_tiff_udf,
+        input_list_node,
+        config,
+        verbose,
+        *args,
+        name=f"{dag_name} exporter",
+        expand_node_output=input_list_node,
+        resources=DEFAULT_RESOURCES if resources is None else resources,
```
thetorpedodog, line 159: likewise, `resources or DEFAULT_RESOURCES` (although this would also use `DEFAULT_RESOURCES` if `resources` were `{}`, so `is None` may be the appropriate check here)

thetorpedodog (approved):
A few minor things but this overall looks good.

src/tiledb/cloud/bioimg/exportation.py
```diff
@@ -1,33 +1,78 @@
...
+        filename = os.path.splitext(os.path.basename(input_file))[0]
+        output_filename = filename + f".{output_ext}" if output_ext else filename
+        return os.path.join(output_dir, output_filename)
+
+    def iter_paths(source: Sequence) -> Iterator[Tuple]:
+        for uri in source:
+            if vfs.is_dir(uri) and tiledb.object_type(uri) != "group":
+                # Folder for exploration
+                contents = vfs.ls(uri)
+                yield from tuple(iter_paths(contents)[1:])
```
thetorpedodog, line 33: Just noticed this—is this what you intend? Iterators don’t generally have indexing available. It seems likely that you would want:

```
yield from iter_paths(contents[1:])
```

Wrapping in `tuple` will also eagerly evaluate the entirety of the inner `iter_paths` call rather than picking from it one by one, which is fine but I would recommend adding a comment explaining *why*.

ktsitsi: After some debugging it seems that after the new release of TileDB the `vfs.ls` drops empty subfolders, so this `[1:]` is not needed anymore. I also changed the direct materialisation of the iterators results and now are fetched one at a time. There was no specific reasoning behind this after all just a design decision on my part.

PR #498, opened 2024-01-04
Add one-line annotation VCF ingestion

gspowley:
Ingestion of annotation VCFs (like ClinVar or gnomAD) differ enough from VCF sample ingestion that we provide a separate ingestion function.

Example for ingesting gnomAD 4.0:

```python

vcf.ingest_annotations(
    "<S3 URI for the new VCF dataset>",
    search_uri="s3://gnomad-public-us-east-1/release/4.0/vcf/genomes/",
    pattern="*.vcf.gz",
    register_name="<gnomAD dataset name in TileDB Cloud>",
    acn="<TileDB Cloud Access Credentials Name>",
)
```

Example for ingesting ClinVar:

```python

vcf.ingest_annotations(
    "<S3 URI for the new VCF dataset>",
    vcf_uri="<S3 URI for the ClinVar vcf.gz file>",
    register_name="<ClinVar dataset name in TileDB Cloud>",
    acn="<TileDB Cloud Access Credentials Name>",
)
```

### DIFF
--- a/src/tiledb/cloud/utilities/_common.py
+++ b/src/tiledb/cloud/utilities/_common.py
@@ -1,11 +1,13 @@
 import configparser
+import functools
+import inspect
 import logging
 import os
 import pathlib
 import subprocess
 import sys
 from concurrent.futures import ThreadPoolExecutor
-from typing import Any, Mapping, Optional, Sequence, Tuple
+from typing import Any, Callable, Mapping, Optional, Sequence, Tuple
 
 import tiledb
 from tiledb.cloud import dag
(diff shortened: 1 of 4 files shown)


src/tiledb/cloud/vcf/ingestion.py
```diff
@@ -1240,6 +1249,161 @@ def group_member_uri(group_uri, group_member, config):
...
+    )
+
+    # Add a node to consolidate the dataset.
+    consolidate_node = graph.submit(
+        consolidate_dataset_udf,
+        dataset_uri,
+        config=config,
+        verbose=verbose,
+        resources=CONSOLIDATE_RESOURCES,
+        name="Consolidate annotations ",
```
Shelnutt2, line 1356: Extra space at the end of name.

src/tiledb/cloud/vcf/ingestion.py
```diff
@@ -1240,6 +1249,161 @@ def group_member_uri(group_uri, group_member, config):
...
+    # Optionally add a node to register the dataset.
+    if register_name:
+        register = graph.submit(
+            register_dataset_udf,
+            dataset_uri=dataset_uri,
+            register_name=register_name,
+            namespace=namespace,
+            config=config,
+            verbose=verbose,
+            name="Register annotations ",
```
Shelnutt2, line 1398: Extra space in `name`.

Shelnutt2 (approved):
Two minor comments, lgtm.

ckrilow (approved):
This looks good to me, great addition!

leipzig (approved):

src/tiledb/cloud/utilities/_common.py
```diff
@@ -272,3 +274,73 @@ def reader(stream):
             stderr = stderr_future.result()
 
             return stdout, stderr
+
+
+def batch(func: Callable) -> Callable:
```
thetorpedodog, line 279: Typing trick: To make functions using this decorator show up correctly in static analysis, you can use a TypeVar:

```python

_CT = TypeVar("_CT", bound=Callable)

def batch(func: _CT) -> _CT:
  ...
```

This specifically indicates to type checkers and static analysis tools that the full details of the type of the incoming function is the exact same as that of the returned value. (It's basically the static analysis equivalent of `functools.wraps`.) You can use this or just leave it as-is; I doubt many people will be using this in the context of a case where people would be using static analysis tools. (For instance, autocomplete and other hints in things like ipython use the function's `__doc__` and `__signature__`, which are unchanged by the change in type annotation.)

Separately:

`as_batch` might be a better name. Also, I would recommend that, rather than using it directly as a decorator, you have the function and expose the wrapped version, maybe by doing:

```python

def ingest_annotations(...) -> ...:
  ...

ingest_annotations_batch = as_batch(ingest_annotations)

# You could also do `_ingest_annotations` for the implementation
# and `ingest_annotations` for the exported version.
```

That way, the mechanism of what is actually run is separated from the policy of where to run it, which may be useful for testing or other scenarios.

src/tiledb/cloud/utilities/_common.py
```diff
@@ -272,3 +274,73 @@ def reader(stream):
...
+        """
+        Filter kwargs to only include valid arguments for the function.
+
+        :param function: function to validate kwargs for
+        :param kwargs: kwargs to filter
+        :return: filtered kwargs
+        """
+        valid_args = inspect.signature(function).parameters
+        filtered_kwargs = {k: v for k, v in kwargs.items() if k in valid_args}
+        return filtered_kwargs
```
thetorpedodog, line 298: Since this is only used with `func` as the `function` parameter, I would recommend either just using the captured `func` value (i.e., `def filter_kwargs(kwargs: ...) -> ...`) or pulling this out into the module level (`def _filter_kwargs(function, kwargs)` at the root.

thetorpedodog (approved):
Overall this looks very good! I have a few small suggestions.

src/tiledb/cloud/vcf/ingestion.py
gspowley: The extra spaces were historically cosmetic. We can remove them now.

src/tiledb/cloud/utilities/_common.py
gspowley: Thanks, made these changes.

src/tiledb/cloud/utilities/_common.py
gspowley: I pulled it out to the root, in case we need to use it in the future.

PR #499, opened 2024-01-05
Use replacing visitor so we actually descend into batch args.

thetorpedodog:
When server-side ("batch") DAG execution was added, the implementation used custom code for finding parent nodes, rather than the existing visitor system. This meant that if a Node was passed in not at the top level, it would not work.

This change replaces it with a visitor-based system that allows users to pass in nodes just as they would in client-side execution.

### DIFF
--- a/src/tiledb/cloud/dag/dag.py
+++ b/src/tiledb/cloud/dag/dag.py
@@ -36,6 +36,7 @@ from .._results import codecs
 from .._results import results
 from .._results import stored_params
 from .._results import tiledb_json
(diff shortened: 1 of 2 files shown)


johnkerl (approved):
🚢 🎉

johnkerl:
[sc-38858]

johnkerl:
@thetorpedodog any reason not to merge this?

johnkerl:
Thanks @thetorpedodog !

PR #503, opened 2024-01-22
Add ingestion argument for converter selection

ktsitsi:
This PR:

- Introduces an already supported feature of the underlying `tiledb-bioimg` for supporting a user-defined converter for the ingestion without automatically inferring it from the file extension of the data source.

### DIFF
--- a/src/tiledb/cloud/bioimg/ingestion.py
+++ b/src/tiledb/cloud/bioimg/ingestion.py
@@ -154,8 +154,6 @@ def ingest(
                         **kwargs,
                     )
 
-    submit = graph.submit_local if local else graph.submit
-
     if isinstance(source, str):
         # Handle only lists
         source = [source]
@@ -189,6 +187,7 @@ def ingest(
         name=f"{dag_name} input collector",
         result_format="json",
     )
+    submit = graph.submit_local if local else graph.submit
 
     # serialize udf arguments
     compressor = kwargs.pop("compressor", None)
--- a/src/tiledb/cloud/bioimg/ingestion.py
+++ b/src/tiledb/cloud/bioimg/ingestion.py
@@ -154,6 +154,8 @@ def ingest(
                         **kwargs,
                     )
 
+    submit = graph.submit_local if local else graph.submit
+
     if isinstance(source, str):
         # Handle only lists
         source = [source]
--- a/src/tiledb/cloud/bioimg/ingestion.py
+++ b/src/tiledb/cloud/bioimg/ingestion.py
@@ -12,6 +12,7 @@ DEFAULT_RESOURCES = {"cpu": "8", "memory": "4Gi"}
 DEFAULT_IMG_NAME = "3.9-imaging-dev"
 DEFAULT_DAG_NAME = "bioimg-ingestion"
 _SUPPORTED_EXTENSIONS = (".tiff", ".tif", ".svs")
+_SUPPORTED_CONVERTERS = ("tiff", "zarr", "svs")
 
 
 def ingest(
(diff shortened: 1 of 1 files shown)


src/tiledb/cloud/bioimg/ingestion.py
```diff
@@ -11,6 +11,7 @@
 DEFAULT_IMG_NAME = "3.9-imaging-dev"
 DEFAULT_DAG_NAME = "bioimg-ingestion"
 _SUPPORTED_EXTENSIONS = (".tiff", ".tif", ".svs")
+_SUPPORTED_CONVERTERS = ("tiff", "zarr", "svs")
```
ihnorton, line 14: Shouldn't `osd` be here?
Why allow `svs` here, it's not a specific converter type...

src/tiledb/cloud/bioimg/ingestion.py
```diff
@@ -162,7 +179,7 @@ def ingest_tiff_udf(
     logger.debug("Building graph")
     graph = dag.DAG(
         name=dag_name,
-        mode=dag.Mode.BATCH,
+        mode=dag.Mode.REALTIME if local else dag.Mode.BATCH,
```
Shelnutt2, line 182: Should there be a distinction between realtime, batch and proper local mode?

src/tiledb/cloud/bioimg/ingestion.py
```diff
@@ -125,6 +132,15 @@ def ingest_tiff_udf(
...
         from tiledb.bioimg import from_bioimg
 
+        converter = kwargs.get("converter", None)
+        user_converter = Converters.OMETIFF
+        if not converter or converter == "tiff":
+            user_converter = Converters.OMETIFF
+        elif converter == "zarr":
+            user_converter = Converters.OMEZARR
+        elif converter == "osd":
+            user_converter = Converters.OSD
```
thetorpedodog, line 142: This could be combined with the check below so it all happens here, in one place, something like

```python
# earlier, at the module level
_SUPPORTED_CONVERTERS = {
    None: Converters.OMETIFF,  # default
    "tiff": Converters.OMETIFF,
    ...
}

# wherever the check ends up
converter_name = kwargs.get("converter")
try:
    converter = _SUPPORTED_CONVERTERS[converter_name]
except KeyError as ke:
    raise ValueError(
        f"{converter_name!r} is not a supported converter;"
        f" please select one of {set(_SUPPORTED_CONVERTERS)}"
    )
```

A couple other small things to note in the above:

- when doing `some_dict.get(...)`, None is the default default so it doesn’t need to be explicitly included
- Instead of using backslashes in strings, you can use the fact that adjacent string literals are concatenated to get nice formatting without weird syntax.

I would also recommend putting this conversion at the outer function level, so that the user gets this feedback *immediately*, before the graph is built and starts trying to be executed.

ktsitsi: I agree with your comments. However the reason I approached it this way is because the Coverters enum is not visible to the Cloud-Py. The Enum is defined in the tiledb-bioimg package and this is why I needed to pass them through the UDF without using them directly. The UDF on the other hand has access to the tiledb.bioimg where the converters are visible, hence the check lies inside the UDF. Is there any alternative which I can't see?

thetorpedodog: I see! That makes sense. Might be worth adding a small comment about this (something like "we might not have tiledb.bioimg installed so we can't directly use the Converters enum here").

thetorpedodog (approved):

PR #505, opened 2024-01-23
Mapping argument for renaming ingested files

ktsitsi:
This PR:

- Enables the user to provide a name mapping between the source files to be ingested and the desired final ingested name. This feature can be used also for de-identification cases. The user should be able to provide a list of output files given a equal length source file list where each of the source files will be renamed to the corresponding output item.
- Adds an extra layer of validation useful not only for the programmatic API but also for the 1-click ingestion feature.

### Changes

- Generalization and new semantics for the provided `output` and `input` ingestor argument. If the user want to point to a folder for storing the ingested data, then the path provided should have a trailing `/`. If so the default behaviour is to ingest the source files with the same name in this directory.
- If the `output` arg does not contain a trailing `/` then it is considered as a full target path for the ingested source. In this case the ingested source file should be one at a time, otherwise errors out with a `ValueError`.

### DIFF
--- a/src/tiledb/cloud/bioimg/helpers.py
+++ b/src/tiledb/cloud/bioimg/helpers.py
@@ -48,22 +48,16 @@ def validate_io_paths(source: Sequence[str], output: Sequence[str]):
     if len(source) == 0:
         raise ValueError("Source list must not be empty.")
 
-    # Case 1: If source is a list of files
-    if all(not is_folder(path) for path in source):
-        # Destination is a folder
-        if len(output) == 1 and is_folder(output[0]):
-            return True
-        # Destination has the same number of paths with source - rename at destination
-        elif len(output) == len(source) and all(not is_folder(path) for path in output):
-            return True
+    if len(output) == 1 and is_folder(output[0]):
+        return True
 
-    # Case 2: If source is a list with length one
-    elif len(source) == 1:
-        if is_folder(source[0]):
-            if len(output) == 1 and is_folder(output[0]):
-                return True
-        else:
-            if len(output) == 1 and (is_folder(output[0]) or not is_folder(output[0])):
+    if all(is_folder(path) for path in output):
+        if all(is_folder(path) for path in source):
+            if len(output) == len(source):
                 return True
 
+    if all(not is_folder(path) for path in output):
+        if all(not is_folder(path) for path in source):
+            if len(output) == len(source):
+                return True
     raise ValueError("Invalid combination of source and output paths.")
--- a/src/tiledb/cloud/bioimg/helpers.py
+++ b/src/tiledb/cloud/bioimg/helpers.py
@@ -36,21 +36,23 @@ def serialize_filter(filter):
         raise TypeError
 
 
+def is_folder(path: str) -> bool:
+    return path.endswith("/")
+
+
 def validate_io_paths(source, output):
     if isinstance(source, str):
-        # Handle only lists
-        if source.endswith("/"):
-            # Source is folder
+        if is_folder(source):
             if not isinstance(output, str):
                 # Output is list of possible 1 or multiple records
-                if not (len(output) == 1 and output[0].endswith("/")):
+                if not (len(output) == 1 and is_folder(output[0])):
                     raise ValueError(
                         "For directory input the output should be also a \
                         directory with a trailing '/'"
                     )
             else:
                 # Output is one record
-                if not output.endswith("/"):
+                if not is_folder(output):
                     raise ValueError(
                         "For directory input the output should \
                         be also a directory with a trailing '/'"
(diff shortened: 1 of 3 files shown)


src/tiledb/cloud/bioimg/helpers.py
```diff
@@ -34,3 +34,89 @@ def serialize_filter(filter):
         return filter_dict
     else:
         raise TypeError
+
+
+def is_folder(path: str) -> bool:
+    return path.endswith("/")
+
+
+def validate_io_paths(source, output):
```
ihnorton, line 43: This is very hard to follow. Please break it down into constituent pieces.

Also, I'm not sure why we are supporting multiple output paths. Unless there is a driving use-case, we may need to take a step back and refactor this API to take separate `source: Union[str, List[str]]` and `dest: str` arguments -- in other words, only allow a single output destination (as a separate argument). I think that will simplify this issue signficantly.

ktsitsi: The behaviour till now was indeed to support a single output. However by supporting multiple outputs given multiple inputs we are able to support the Quest's use-case, where the renaming of the files comes as a direct consequence. I get that the code gets complicated so I will try to simplify it further however this will probably need further restrictions on what the user can give as arguments.

Shelnutt2: Lets move this to a design doc. We are well overdue on delivering this, I'd like to get a design agreed upon and implementation shipped, validated and accepted before end of week. Please create a document today that covers the use cases and the proposal for what we are to support.

sgillies:
@ktsitsi I don't see the mapping argument in the changes yet, am I overlooking it?

I've got a suggestion for how to think about this feature: [tar transforms](https://www.gnu.org/software/tar/manual/html_section/transform.html). Tar is time-tested and its design has been approved (or, at least, tolerated) by lots of developers. Path stripping is tar's simple form, regex is the advanced form.

This might be generally useful for other kinds of ingest, too.

src/tiledb/cloud/bioimg/helpers.py
ktsitsi: https://docs.google.com/document/d/1QDIF0T1mWsaOVswsARy5NKGHXc-Vn98083Sv5dWt6MM/edit?usp=sharing

ktsitsi:
> @ktsitsi I don't see the mapping argument in the changes yet, am I overlooking it?
> 
> I've got a suggestion for how to think about this feature: [tar file name transforms](https://www.gnu.org/software/tar/manual/html_section/transform.html). Tar is time-tested and its design has been approved (or, at least, tolerated) by lots of developers. Path stripping is tar's simple form, regex is the advanced form. Would those cover the use cases?
> 
> This might be generally useful for other kinds of ingest, too.

@sgillies thank you for taking a look. I have updated the PR's description to reflect the latest changes. A mapping argument would be an overkill since the user can utilise list of files as outputs given list of input paths. This will allow them to rename the files being ingested. The API is being used by clients so we any change we do needs to keep that in mind. In case we would like to change the overall thinking we will need to take into account that we need to abstract the implementation from the user. I am open though to explore alternatives as long as the former is satisfied.

ihnorton (changes requested):
Per discussion, we need to add a recursion depth limit (and set it to 1 for cloud ingestion).

src/tiledb/cloud/bioimg/helpers.py
```diff
@@ -34,3 +35,29 @@ def serialize_filter(filter):
...
+
+    if len(output) == 1 and is_folder(output[0]):
+        return True
+
+    if all(is_folder(path) for path in output):
+        if all(is_folder(path) for path in source):
+            if len(output) == len(source):
+                return True
+
+    if all(not is_folder(path) for path in output):
```
ihnorton, line 59: This is the negation of condition on `L54`, use `else:`

src/tiledb/cloud/bioimg/ingestion.py
```diff
@@ -48,19 +52,20 @@ def ingest(
     :param namespace: The namespace where the DAG will run
     :param verbose: verbose logging, defaults to False
     :param output_ext: extension for the output images in tiledb
+    :param traverse: If true then traverse the src paths for exploring images in
```
ihnorton, line 55:
```suggestion
    :param traverse: If true then traverse the src paths recursively to find images to convert in
```

src/tiledb/cloud/bioimg/helpers.py
ktsitsi: The L54 checks if all paths are `folders` where L59 checks if all paths are `files`.  For example a case that hasn't the same result if `else` is used is the following:
For an input like `test3 = [False, True, True]`
```
if all(i for i in test3):
    print('A')
if all(not i for i in test3):
    print('B')
```
prints `nothing` whereas:
```
if all(i for i in test3):
    print('A')
else:
    print('B')
``` 
prints `B`. The test pass even with `else` cause the rest of the code handles it, but it was intended.

ktsitsi:
Closing it in favor of #509 .

PR #507, opened 2024-01-30
Geospatial ingestion - WIP

normanb:
For sc-38757

@sgillies @ktsitsi 

The `get_metadata` groups the source files by destination raster block which is scalable for task graphs. For point clouds and geometries it will return the filtered input files. I checked the test results against expected outputs with GDAL using QGIS.

@sgillies can we review the use of `nodata` in the raster ingest from rasterio? I am not able to override `nodata` for the destination raster and this may need a change in the GDAL driver.

Geometries and point clouds are simpler and I will add these as a separate commit.


src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,331 @@
...
+                            col_off = blk[1][0]
+                            dst_window = rasterio.windows.Window(col_off, row_off,
+                                                                blk[1][1] - col_off, blk[0][1] - row_off)
+                            dst_bounds = rasterio.windows.bounds(dst_window, dst.transform)
+
+                            if len(srcs) == 1:
+                                with rasterio.vrt.WarpedVRT(input_datasets[0], crs=dst.crs, transform=dst.transform, resampling=resampling, width=dst.width, height=dst.height, masked=True) as vrt:
+                                    data = vrt.read(window=dst_window)
+                            else:
+                                data, _ = rasterio.merge.merge(input_datasets, bounds=dst_bounds, res=dst.res,
```
sgillies, line 199: @normanb I think it would be better to use the same merge tool regardless of the number of inputs. Rasterio isn't currently testing that WarpedVRT and merge.merge produce the exact same results.

Does merge.merge work well here? I didn't have TileDB in mind, but I think it ought to be supported well enough.

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,331 @@
...
+                    with rasterio.open(output, "w",
+                                       COMPRESSION=compressor_name[:-6].upper(), # remove trailing `Filter`
+                                       COMPRESSION_LEVEL=compression_level,
+                                       **dst_profile,):
+                        pass
+        else:
+            # srcs and dst have same number of bands
+            resampling = rasterio.enums.Resampling[kwargs.pop("resampling", "bilinear").lower()]
+            with rasterio.open(output, mode="r+") as dst:
+                for b in range(dst.count):
```
sgillies, line 184: @normanb I don't understand why the inputs aren't the outer loop. I'm probably missing something conceptually.

sgillies (commented):
@normanb @ktsitsi I've got some comments, questions, and suggestions.

src/tiledb/cloud/geospatial/helpers.py
```diff
@@ -0,0 +1,188 @@
...
+
+    :param source: The source iterable containing files to be ingested/exported
+    :param num_batches: The number of batches given by the API
+    :return: Tuple batch_size, max_workers
+    """
+    # If num_batches is default create number of images nodes
+    # constraint node max_workers to 20 fully heuristic
+    batch_size = 1 if num_batches is None else math.ceil(len(samples) / num_batches)
+    max_workers = 20 if num_batches is None else None
+    return batch_size, max_workers
```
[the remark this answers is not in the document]
normanb: This was copied over as a pattern from the biomedical ingestion, but I don't mind moving them into the same module. Having a shared helpers class for calculating batch sizes would be useful.

src/tiledb/cloud/geospatial/helpers.py
```diff
@@ -0,0 +1,188 @@
...
+from rtree import index
+
+import tiledb
+
+
+class RasterIngestionWarning(UserWarning):
+    pass
+
+@dataclass
+class GeoExtents:
```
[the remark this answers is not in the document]
normanb: I was trying to think of a way to capture the `extents` metadata across point clouds, rasters and geometries which I introduced this. Would you prefer `GeoMetadata` as a class name?

src/tiledb/cloud/geospatial/helpers.py
```diff
@@ -0,0 +1,188 @@
...
+
+import pdal
+import rasterio
+import shapely
+from rtree import index
+
+import tiledb
+
+
+class RasterIngestionWarning(UserWarning):
```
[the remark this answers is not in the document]
normanb: 👍

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,331 @@
...
+                    for blk, srcs in inputs:
+                        input_datasets = []
+                        try:
```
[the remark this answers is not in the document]
normanb: 👍

src/tiledb/cloud/geospatial/ingestion.py
normanb: I couldn't get it to work with a single input file which is why I used `WarpedVRT`, the return from a single input to `merge.merge` is an array of zeros.

src/tiledb/cloud/geospatial/ingestion.py
normanb: I did this because you can technically have different block sizes per band in GDAL, however in TileDB you cannot so I will move the band loop up a level and reduce the number of times the datasets are opened.

normanb:
Thanks @sgillies, I will make these changes and run `black` before adding more support for point clouds and geometries.

src/tiledb/cloud/geospatial/ingestion.py
sgillies: That feels like a rasterio bug.

Speaking of array of zeros, this line here https://github.com/rasterio/rasterio/blob/main/rasterio/merge.py#L293 means that the "nodata" value of an output array isn't properly configurable like file output is. That's something I should fix, too. Is that what you're experiencing @normanb? That passing `dst.nodata` to `merge()` has no effect?

sgillies: For 1.4.0, I propose to do this in Rasterio: https://github.com/rasterio/rasterio/issues/1867. That would be useful for TileDB raster ingest if "r+" mode was also supported.

src/tiledb/cloud/geospatial/helpers.py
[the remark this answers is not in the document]
sgillies: How about src/tiledb/cloud/batching.py if it will be used in >2-3 places? I'm generally 👎 on helpers.py or utils.py because then they tend to become disorganized and hard to maintain.

src/tiledb/cloud/geospatial/helpers.py
```diff
@@ -0,0 +1,188 @@
...
+            return (self.minx, self.miny, self.minz,
+                    self.maxx, self.maxy, self.maxz)
+        else:
+            return (self.minx, self.miny, self.maxx, self.maxy)
+
+@dataclass
+class GeoMetadata:
+    path: os.PathLike
+    extents: GeoExtents
+    projection: str
```
sgillies, line 38: @normanb how about this?

```suggestion
    bbox: BoundingBox
    crs: str
```

`BoundingBox` is consistent with other uses of the term, including PDAL's and Rasterio's. `GeoExtents` could be renamed to `BoundingBox`, no need to change the class. `crs` is more strictly correct than projection.

src/tiledb/cloud/geospatial/helpers.py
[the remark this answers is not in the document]
normanb: Turns out I added this but don't use it, so for now it is removed.

normanb:
Items `ToDo`;

- [ ] Add more tests
- [ ] singular usage of rasterio.merge rather than `warpedvrt`
- [ ] check `nodata` is propagated for rasters
- [ ] Geometries (need `pyopener` or similar)

Can I get a review of the new code before I add these items?

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1182 @@
...
+                    max_files=max_files,
+                )
+            else:
+                sources = find(
+                    search_uri,
+                    config=config,
+                    excludes=ignore,
+                    includes=pointcloud_match,
+                    max_files=max_files,
+                )
```
ktsitsi, line 836: You could shorten this with the following:
```python
sources = find(
                    search_uri,
                    config=config,
                    excludes=ignore,
                    includes=pattern if pattern else pointcloud_match,
                    max_files=max_files,
                )
```

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1182 @@
...
+            return False
+
+    def geometry_match(f):
+        try:
+            fiona.drvsupport.driver_from_extension(f)
+            return True
+        except ValueError:
+            return False
+
+    if dataset_type == "pointcloud":
```
ktsitsi, line 817: You could simplify code by defining a dict that will store for each key (`datatype`) the functions used for `metadata` and `udfs` and then the rest of the code will be less repeatable and you will need to define just the different kwargs.

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1182 @@
...
+            namespace=namespace,
+        )
+
+        nodes = []
+
+        # create first node
+        ingest_node = graph.submit(
+            fn,
+            dataset_uri=dataset_uri,
+            config=config,
```
ktsitsi, line 945: What is the purpose of this node? how does it take different arguments but calling the same `fn` with the rest below?

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1182 @@
...
+            name=f"{task_prefix} - {0}",
+            mode=tiledb.cloud.dag.Mode.BATCH,
+            resources=DEFAULT_RESOURCES,
+            image_name=DEFAULT_IMG_NAME,
+            access_credentials_name=acn,
+            **kwargs,
+        )
+        # work functions
+        for i, work in enumerate(batch(samples, batch_size)):
+            logger.info(f"Adding batch {i + 1}")
```
ktsitsi, line 959: I would really advise the use of the `expand_node_output` approach instead of a for loop here.

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1182 @@
...
+            fn(
+                work,
+                dataset_uri=dataset_uri,
+                append=True,
+                verbose=verbose,
+                trace=trace,
+                log_uri=log_uri,
+                **kwargs,
+            )
+
```
ktsitsi, line 1034: IDK as said in another comment if we want this code segment to be located here.

ktsitsi (changes requested):
Looks proper in general, just some minor comments to be addressed.

src/tiledb/cloud/utilities/_common.py
```diff
@@ -346,3 +380,84 @@ def wrapper(*args, **kwargs) -> None:
...
+
+        for f in list_files(listing):
+            current_count += 1
+            if max_count and current_count == max_count:
+                return
+            else:
+                yield f
+
+
+def batch(items, chunk_size):
```
gspowley, line 446: Add typing and docstring to be consistent with this codebase.

src/tiledb/cloud/utilities/_common.py
```diff
@@ -346,3 +380,84 @@ def wrapper(*args, **kwargs) -> None:
...
+    # Iterator for providing batches of chunks
+    length = len(items)
+    for ndx in range(0, length, chunk_size):
+        yield items[ndx : min(ndx + chunk_size, length)]
+
+
+def serialize_filter(filter):
```
gspowley, line 453: Add typing and docstring to be consistent with this codebase.

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1182 @@
...
+from typing import Any, Iterable, List, Mapping, Optional, Sequence, Tuple, Union
+
+import fiona
+import laspy
+import rasterio
+import shapely
+from rtree import index
+
+import tiledb
+from tiledb.cloud.utilities import Profiler
```
gspowley, line 13: The `Profiler` placeholders look good. When you're ready to use the profiler, `tiledb.cloud.utilities.create_log_array` will create the log array, which is needed by the functions being profiled.

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1182 @@
...
+    :param config: config dictionary, defaults to None
+    :param id: profiler event id, defaults to "consolidate"
+    :param verbose: verbose logging, defaults to False
+    """
+
+    logger = get_logger_wrapper(verbose)
+
+    with tiledb.scope_ctx(config):
+        with Profiler(array_uri=log_array_uri, id=id):
+            modes = ["commits", "fragment_meta", "array_meta"]
```
gspowley, line 1057: Based on the name of this function (`consolidate_fragment_meta`), check if these are the consolidate/vacuum modes you want.

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1182 @@
...
+        defaults to None
+    :param config: config dictionary, defaults to None
+    :param namespace: TileDB-Cloud namespace, defaults to None
+    :param register_name: name to register the dataset with on TileDB Cloud,
+        defaults to None
+    :param search_uri: URI to search for geospatial dataset files, defaults to None
+    :param pattern: Unix shell style pattern to match when searching for files,
+        defaults to None
+    :param ignore: Unix shell style pattern to ignore when searching for files,
+        defaults to None
```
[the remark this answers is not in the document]
gspowley: There is a use case for both `pattern` and `ignore`. For example, if you want to find all `*.foo` files but ignore `bad.foo`, then `pattern="*.foo"` and `ignore="bad.foo"`.

```

File: src/tiledb/cloud/geospatial/ingestion.py [hunk 18]
```diff
@@ -0,0 +1,1182 @@
...
+        dataset_type=dataset_type,
+        acn=acn,
+        config=config,
+        namespace=namespace,
+        register_name=register_name,
+        search_uri=search_uri,
+        pattern=pattern,
+        ignore=ignore,
+        dataset_list_uri=dataset_list_uri,
+        max_files=max_files,
```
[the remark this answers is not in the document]
gspowley: In VCF ingestion, the `max_files` is primarily used during testing to limit the size of the ingestion.

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1182 @@
...
+    verbose: bool = False,
+    trace: bool = False,
+    log_uri: Optional[str] = None,
+    unit_testing: bool = False,
+) -> None:
+    """
+    Ingest samples into a dataset.
+
+    :param dataset_uri: dataset URI
+    :param dataset_type: Classifier, one of pointcloud, raster or geometry
```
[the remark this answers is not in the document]
normanb: I am going to use an Enum to enforce known input data types.

src/tiledb/cloud/geospatial/ingestion.py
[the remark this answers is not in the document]
normanb: 👍

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1182 @@
...
+    :param batch_size: batch size for dataset ingestion, defaults to BATCH_SIZE
+    :param tile_size: for rasters this is the tile (block) size
+        for the merged destination array defaults to 1024
+    :param chunk_size: for point cloud this is the PDAL chunk size, defaults to 1000000
+    :param nodata: NODATA value for rasters
+    :param res: Tuple[float, float], output resolution in x/y
+    :param verbose: verbose logging, defaults to False
+    :param trace: bool, enable trace for logging, defaults to False
+    :param log_uri: log array URI
+    :param unit_testing: bool internal flag for unit testing the code flow
```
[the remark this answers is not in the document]
normanb: 👍 I have made it an environment variable.

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1182 @@
...
+    """
+
+    # Validate user input
+    if sum([bool(search_uri), bool(dataset_list_uri)]) != 1:
```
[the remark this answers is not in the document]
normanb: 👍 a bitwise `&` is needed to test that not both `search_uri` and `dataset_list_uri` are set.

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1182 @@
...
+        raise ValueError(
+            "Exactly one of `search_uri` or `dataset_list_uri`" " must be provided."
+        )
+
+    if not search_uri and (pattern or ignore):
+        raise ValueError("Only specify `pattern` or `ignore` with `search_uri`.")
+
```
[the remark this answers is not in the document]
normanb: a `pattern` and `ignore` should not be set with a `dataset_list_uri`. This follows the convention in VCF.

src/tiledb/cloud/geospatial/ingestion.py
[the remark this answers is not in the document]
normanb: 👍 This limits large ingestions when testing but this can also be achieved with a `dataset_list_uri`, I am happy leaving the parameter in.

src/tiledb/cloud/geospatial/ingestion.py
normanb: 👍

src/tiledb/cloud/geospatial/ingestion.py
normanb: 👍

src/tiledb/cloud/geospatial/ingestion.py
normanb: This node is the set up node and doesn't take any source files just the metadata needed to construct an empty array.

src/tiledb/cloud/geospatial/ingestion.py
normanb: Why? `expand_node_output` takes the output from a previous node, the graph doesn't have that dependency.

src/tiledb/cloud/utilities/_common.py
normanb: 👍

src/tiledb/cloud/utilities/_common.py
normanb: 👍

src/tiledb/cloud/geospatial/ingestion.py
normanb: 👍

src/tiledb/cloud/geospatial/ingestion.py
normanb: 👍 I have renamed the function to that it matches the consolidate/vacuum modes of all metadata.

normanb:
Ingestion of geometries requires an update to use a tiledb VSI opener (a parameter).

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1287 @@
...
+            crs = None
+
+            for pth in sources:
+                logger.debug("finding metadata for %r", pth)
+                with fiona.open(pth) as src:
+                    if crs is None:
+                        crs = src.crs
+                    else:
+                        if src.crs != crs:
+                            raise ValueError("All datasets need to be in the same CRS")
```
thetorpedodog, line 193: This can be shortened a bit:

```python
crs = crs or src.crs
# Alternately, if `""` or some other falsey value is a valid crs or schema:
crs = src.crs if crs is None else crs
if src.crs != crs:  # this can now run unconditionally
  raise ...
```

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1287 @@
...
+                            raise ValueError(
+                                "All datasets need to have the same schema"
+                            )
+
+                    # 2.5 is supported internally but not needed for ingest
+                    extents = BoundingBox(
+                        minx=src.bounds[0],
+                        miny=src.bounds[1],
+                        maxx=src.bounds[2],
+                        maxy=src.bounds[3],
```
thetorpedodog, line 208: likewise, this becomes `BoundingBox(min=Point.at(src.bounds[:2], max=src.bounds[2:])`

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1287 @@
...
+
+    :param sources: iterator, paths or path to process
+    :param config: dict, configuration to pass on tiledb.VFS
+    :param verbose: bool, enable verbose logging, default is False
+    :param trace: bool, enable trace logging, default is False
+    :param log_uri: Optional[str] = None,
+    :Return: GeoMetadata, a populated GeoMetadata object
+    """
+    if len(sources) == 0:
+        raise ValueError("Input point cloud datasets required")
```
thetorpedodog, line 104: nit: since empty iterables are false-like, this can be

```python
if not sources:
  raise ValueError(...)
```

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1287 @@
...
+             and the files that contribute to each block
+    """
+    if len(sources) == 0:
+        raise ValueError("Input raster datasets required")
+
+    logger = get_logger_wrapper(verbose)
+    full_extents = None
+    crs = None
+    num_bands = None
+    do_res_check = True
```
thetorpedodog, line 256: In the interest of only initializing this variable once, I would recommend moving the `do_res_check = True` into the else-block below.

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1287 @@
...
+        xres, yres = res
+        do_res_check = False
+    else:
+        xres = None
+        yres = None
+
+    def group_by_raster_block(
+        meta: Sequence[GeoMetadata],
+    ) -> Tuple[GeoBlockMetadata, ...]:
+        nonlocal logger
```
thetorpedodog, line 269: Since this doesn’t change the value of `logger` (i.e., it doesn’t assign to it/re-bind it), this statement is not needed; the closure captures `logger`.

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1287 @@
...
+            *full_extents.bounds, width, height
+        )
+        dst_windows = rasterio.windows.window_split(height, width, dst_tile_size**2)
+
+        results = []
+        for _, w in dst_windows:
+            region_bounds = rasterio.windows.bounds(w, dst_transform)
+            # find intersection
+            intersects = tuple(idx.intersection(region_bounds))
+            if len(intersects) > 0:
```
thetorpedodog, line 297: likewise, `if intersects:`

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1287 @@
...
+        if self.minz:
+            return (self.minx, self.miny, self.minz, self.maxx, self.maxy, self.maxz)
+        else:
+            return (self.minx, self.miny, self.maxx, self.maxy)
+
+
+@dataclass
+class GeoBlockMetadata:
+    ranges: Tuple[Tuple[int, int], Tuple[int, int]]
+    files: Tuple[os.PathLike, ...]
```
thetorpedodog, line 64: using `attrs` will also make this a bit nicer, since you can use a converter here:

```python
files: Tuple[os.PathLike, ...] = attrs.field(converter=tuple)
```

This will allow you to pass any Iterable in and it will be converted to a tuple.

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1287 @@
...
+        for _, w in dst_windows:
+            region_bounds = rasterio.windows.bounds(w, dst_transform)
+            # find intersection
+            intersects = tuple(idx.intersection(region_bounds))
+            if len(intersects) > 0:
+                # append the result in pixel coords
+                results.append(
+                    GeoBlockMetadata(
+                        ranges=w.toranges(),
+                        files=tuple([meta[s].paths for s in intersects]),
```
thetorpedodog, line 302: With the `converter=tuple` change to GeoBlockMetadata, this can be changed to `files=meta[s].paths for s in intersects,`, without needing to be explicitly wrapped with `tuple()`.

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1287 @@
...
+    import fiona
+
+    from tiledb import filter
+
+    with tiledb.scope_ctx(config):
+        with Profiler(array_uri=log_uri, id=id, trace=trace):
+            tiledb.VFS()
+            logger = get_logger_wrapper(verbose)
+
+            if append:
```
thetorpedodog, line 437: Looking at how all these if-blocks are run, it seems like in the general case only one of them is ever going to run and they're the last thing in the function (except the logging statement). It might make more sense to structure it with early returns. To maintain the logging, we can use a try-finally:

```python
try:
  if append:
     ...
     return
  if extends and crs:
    ...
    return
finally:
  logger.info("max memory usage: %.3f GiB", max_memory_usage() / (1 << 30))
```

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1287 @@
...
+            logger.info("max memory usage: %.3f GiB", max_memory_usage() / (1 << 30))
+
+
+def ingest_point_cloud_udf(
+    sources: Optional[Sequence[str]] = None,
+    *,
+    dataset_uri: str,
+    template_sample: str,
+    append: bool = False,
+    extents: Optional[Tuple[float, float, float, float, float, float]] = None,
```
thetorpedodog, line 503: This might be worth defining a type alias for, and/or breaking down:

```python
SixTuple = Tuple[float, float, float, float, float, float]
# or maybe:
XYZ = Tuple[float, float, float]
XYZPair = Tuple[XYZ, XYZ]

# so this becomes
def ingest_point_cloud_udf(
    ...,
    offsets: Optional[SixTuple] = None,
    # or maybe
    scales: Optional[XYZPair] = None,
```

(In both cases I'm sure you can come up with a better name.)

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1287 @@
...
+                            ).pipeline(arr)
+                            pipeline.execute()
+                else:
+                    raise ValueError("Insufficient metadata for point cloud ingestion")
+
+            logger.info("max memory usage: %.3f GiB", max_memory_usage() / (1 << 30))
+
+
+def ingest_raster_udf(
+    chunks: Optional[Tuple[GeoBlockMetadata, ...]] = None,
```
thetorpedodog, line 599: In general, for inputs to functions, it's usually better to take in a general Iterable rather than a variable-sized tuple. If you need it to be a tuple itself, you can convert it (which, if it's already a tuple, is free because it will give you back the same object).

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1287 @@
...
+                from rasterio.transform import from_origin
+
+                from tiledb import filter
+
+                if (
+                    extents is None
+                    or res is None
+                    or crs is None
+                    or band_count is None
+                    or dtype is None
```
thetorpedodog, line 667: `None in (extents, res, crs, band_count, dtype)` will make this a bit less repetitive.

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1287 @@
...
+                ):
+                    raise ValueError(
+                        """CRS, band count, extents and resolution are
+                        required to create empty raster array"""
```
thetorpedodog, line 671: To do nice strings when it's longer than will fit in one line, you can use the fact that adjacent literals are automatically concatenated:

```python
raise ValueError(
    "CRS, band count, ..."
    " required to ..."
)
```

That means you end up with one string that looks like "CRS, band count, ... required to ..." without excessive white space.

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1287 @@
...
+                }
+
+                logger.debug("Destination raster profile %r", dst_profile)
+
+                if nodata is not None:
+                    dst_profile.update(
+                        {
+                            "nodata": nodata,
+                        }
+                    )
```
thetorpedodog, line 694: `dst_profile.update(nodata=nodata)`

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1287 @@
...
+                if compressor:
+                    compressor_args = dict(compressor)
+                    compressor_name = compressor_args.pop("_name")
+                    if compressor_name:
+                        compressor_args = {
+                            k: None if not v else v for k, v in compressor_args.items()
+                        }
+                        compression_filter = vars(filter).get(compressor_name)(
+                            **compressor_args
+                        )
```
thetorpedodog, line 705: A few things here:

1. You can directly `**` dict literals.
2. Instead of `vars().get()`, use `getattr`
3. `first or second` is exactly equivalent to `second if not first else first`.

```python
compression_filter = getattr(filter, compressor_name)(
    **{k: v or None for k, v in compressor_args.items()}
)
```

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1287 @@
...
+                        compression_level = compression_filter.level
+
+                    else:
+                        raise ValueError
```
thetorpedodog, line 709: message?

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1287 @@
...
+
+                    with rasterio.open(
+                        dataset_uri,
+                        mode="w",
+                        COMPRESSION=compressor_name[
+                            :-6
+                        ].upper(),  # remove trailing `Filter`
```
thetorpedodog, line 716: I think it would be worth it to add a variable for this to avoid this weird line-wrap.

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1287 @@
...
+                if resampling:
+                    resampling = rasterio.enums.Resampling[resampling.lower()]
+                with rasterio.open(dataset_uri, mode="r+") as dst:
+                    for chunk in chunks:
+                        input_datasets = [
+                            rasterio.open(f, opener=vfs.open) for f in chunk.files
+                        ]
+                        try:
+                            row_off = chunk.ranges[0][0]
+                            col_off = chunk.ranges[1][0]
```
thetorpedodog, line 741: This should probably be rearranged to `col_off` first to match below. (Alternately, pass in the params to `Window` as named arguments.)

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1287 @@
...
+                            for b in range(dst.count):
+                                chunk_arr, _ = rasterio.merge.merge(
+                                    input_datasets,
+                                    bounds=chunk_bounds,
+                                    nodata=nodata,
+                                    indexes=[b + 1],
+                                )
+                                dst.write(chunk_arr, window=chunk_window)
+                        finally:
+                            map(lambda s: s.close(), input_datasets)
```
thetorpedodog, line 761: Since this isn't really a functional transformation, I think that

```python
for s in input_datasets:
  s.close()
```

is probably a better representation for what's going on here.

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1287 @@
...
+                namespace=namespace,
+            )
+
+
+def ingest_datasets_dag(
+    dataset_uri: str,
+    *,
+    dataset_type: DatasetType,
+    acn: Optional[str] = None,
+    config=None,
```
thetorpedodog, line 862: `Dict[str, object]`?

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1287 @@
...
+                    "template_sample": meta.paths[0],
+                    "extents": meta.extents,
+                    "offsets": meta.offsets,
+                    "scales": meta.scales,
+                    "chunk_size": chunk_size,
+                }
+            )
+            samples = meta.paths[1:]
+        else:
+            raise ValueError("Require at least one point cloud file to have been found")
```
thetorpedodog, line 996: insetad of

```python
if something:
  do_stuff()
else:
  raise ...
```

This can be written as
```python
if not something:
  raise ...
do_stuff()
```
to reduce the amount of nesting and make the "happy path" easier to follow.

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1287 @@
...
+    elif dataset_type == DatasetType.RASTER:
+        kwargs.update(
+            {
```
thetorpedodog, line 999: since the keys to this `.update` are all literals, this can be written as
```python
kwargs.update(
    crs=meta.crs,
    ...
)
```

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1287 @@
...
+    :param chunk_size: for point cloud this is the PDAL chunk size, defaults to 1000000
+    :param nodata: NODATA value for rasters
+    :param res: Tuple[float, float], output resolution in x/y
+    :param verbose: verbose logging, defaults to False
+    :param trace: bool, enable trace for logging, defaults to False
+    :param log_uri: log array URI
+    """
+
+    # Validate user input
+    if bool(search_uri) & bool(dataset_list_uri):
```
thetorpedodog, line 1246: Should this be `if bool(search_uri) != bool(dataset_list_uri)`? Otherwise, this will work fine if neither is provided.

src/tiledb/cloud/utilities/_common.py
```diff
@@ -347,3 +382,94 @@ def wrapper(*args, **kwargs) -> Dict[str, object]:
...
+            if max_count and current_count == max_count:
+                return
+            else:
+                yield f
+
+
+T = TypeVar("T")
+
+
+def batch(items: Sequence[T], chunk_size: int) -> Iterator[Sequence[T]]:
```
thetorpedodog, line 451: To avoid the overloaded term `batch`, maybe `chunked`? Alternately, we could [directly use `more_itertools.chunked`](https://more-itertools.readthedocs.io/en/stable/api.html#more_itertools.chunked).

src/tiledb/cloud/utilities/_common.py
```diff
@@ -347,3 +382,94 @@ def wrapper(*args, **kwargs) -> Dict[str, object]:
...
+
+def serialize_filter(filter) -> dict:
+    """Serialize TileDB filter.
+
+    :param filter: TileDB filter to serialize
+    :return: dict, TileDB filter attributes
+    """
+    if isinstance(filter, tiledb.Filter):
+        filter_dict = filter._attrs_()
+        filter_dict["_name"] = type(filter).__name__
```
thetorpedodog, line 472: This is the same format we use in TileDB SOMA, except that we use the key `_type` rather than `_name`. Maybe useful to keep that consistent?

tests/test_geospatial.py
thetorpedodog, line 1: Since these tests can’t run with the basic TileDB install, it might be better to put them in a `geospatial` directory for easy exclusion, and so that it’s obvious that the datafiles belong with the geospatial test.

thetorpedodog (commented):
This change is looking pretty good so far. While I have quite a few comments here, overall it's well-written and fairly easy to follow, and I think these suggestions can make it even more so.

src/tiledb/cloud/utilities/_common.py
thetorpedodog: forgot to put the link here: https://github.com/single-cell-data/TileDB-SOMA/blob/db548eb0d8ef09c941993c56b394aa9cdd6d1777/apis/python/src/tiledbsoma/options/_tiledb_create_options.py#L30-L36

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1182 @@
...
+DEFAULT_DAG_NAME = "geo-ingestion"
+
+# Default values for ingestion parameters
+RASTER_TILE_SIZE = 1024
+POINT_CLOUD_CHUNK_SIZE = 1_000_000
+MAX_WORKERS = 40
+BATCH_SIZE = 10
+
+
+@dataclass
```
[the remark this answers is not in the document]
normanb: 👍

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1287 @@
...
+            for f in sources:
+                logger.debug("finding metadata for %r", f)
+                with vfs.open(f, "rb") as src:
+                    # only open header
+                    las = laspy.open(src)
+                    hdr = las.header
+                    if mins is None:
+                        mins = hdr.mins
+                    else:
+                        mins = [min(e) for e in zip(mins, hdr.mins)]
```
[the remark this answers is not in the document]
normanb: 👍

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1287 @@
...
+
+
+@dataclass
+class BoundingBox:
+    minx: float
+    miny: float
+    maxx: float
+    maxy: float
+    minz: Optional[float] = None
+    maxz: Optional[float] = None
```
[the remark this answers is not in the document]
normanb: The abstraction is the `BoundingBox` class, adding a `Point` class makes the code more complicated and doesn't reflect that these are bounds.

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1287 @@
...
+    ranges: Tuple[Tuple[int, int], Tuple[int, int]]
+    files: Tuple[os.PathLike, ...]
+
+
+@dataclass
+class GeoMetadata:
+    # common properties
+    extents: BoundingBox
+    crs: str = None
+    paths: Optional[Union[List[os.PathLike], os.PathLike]] = None
```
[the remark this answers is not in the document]
normanb: 👍

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1287 @@
...
+    scales: Optional[Tuple[float, float, float]] = None
+    offsets: Optional[Tuple[float, float, float]] = None
+    # geometry properties
+    geometry_schema: Optional[dict] = None
+
+
+def get_pointcloud_metadata(
+    sources: Iterable[os.PathLike],
+    *,
+    config: Mapping[str, Any] = None,
```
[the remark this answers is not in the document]
normanb: 👍

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1287 @@
...
+    res: Optional[Tuple[float, float]] = None
+    block_metadata: Optional[Tuple[GeoBlockMetadata, ...]] = None
+    # point cloud properties
+    scales: Optional[Tuple[float, float, float]] = None
+    offsets: Optional[Tuple[float, float, float]] = None
+    # geometry properties
+    geometry_schema: Optional[dict] = None
+
+
+def get_pointcloud_metadata(
```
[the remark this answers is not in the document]
normanb: 👍

requirements-geospatial-py3.9.txt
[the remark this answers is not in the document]
normanb: I see the `update_requirements_txt` script, it is call pip freeze to generate the `requirements` files, what are you suggesting here? I put the geospatial in a separate requirements file more as a hint as we require a GDAL build from source.

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1287 @@
...
+
+            return GeoMetadata(
+                paths=sources,
+                extents=BoundingBox(
+                    minx=mins[0],
+                    miny=mins[1],
+                    minz=mins[2],
+                    maxx=maxs[0],
+                    maxy=maxs[1],
+                    maxz=maxs[2],
```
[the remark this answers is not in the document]
normanb: I see how this makes the code appear simpler, but minx, miny, etc reflect what we have which is bounds in x/y/z. I could use the constructs in shapely for this if this is a change that you think is important.

src/tiledb/cloud/geospatial/ingestion.py
normanb: as before, lets discuss, because we can use shapely if this is a must have change.

src/tiledb/cloud/geospatial/ingestion.py
normanb: 👍

src/tiledb/cloud/geospatial/ingestion.py
normanb: 👍

src/tiledb/cloud/geospatial/ingestion.py
normanb: 👍

src/tiledb/cloud/geospatial/ingestion.py
normanb: 👍

src/tiledb/cloud/geospatial/ingestion.py
normanb: 👍

src/tiledb/cloud/geospatial/ingestion.py
normanb: 👍

src/tiledb/cloud/geospatial/ingestion.py
normanb: 👍

src/tiledb/cloud/geospatial/ingestion.py
normanb: 👍

src/tiledb/cloud/geospatial/ingestion.py
normanb: 👍

src/tiledb/cloud/geospatial/ingestion.py
normanb: 👍

src/tiledb/cloud/geospatial/ingestion.py
normanb: 👍

src/tiledb/cloud/geospatial/ingestion.py
normanb: 👍

src/tiledb/cloud/geospatial/ingestion.py
normanb: 👍

src/tiledb/cloud/geospatial/ingestion.py
ktsitsi: Yeap exactly. I had in mind that you can have a previous node in your taskgraph that splits the samples in batches and feed the outputs directly to the next node which is for ingestion. This way you avoid batching client side and `for loops` you leave the backend do the job for you and run all these tasks in parallel.

src/tiledb/cloud/geospatial/ingestion.py
normanb: 👍

src/tiledb/cloud/geospatial/ingestion.py
normanb: 👍

src/tiledb/cloud/geospatial/ingestion.py
normanb: 👍

src/tiledb/cloud/geospatial/ingestion.py
normanb: 👍

src/tiledb/cloud/geospatial/ingestion.py
normanb: 👍

src/tiledb/cloud/geospatial/ingestion.py
normanb: this is checking that not both `search_uri` and `dataset_list_uri` are provided. I will simplify this.

src/tiledb/cloud/utilities/_common.py
normanb: 👍

src/tiledb/cloud/utilities/_common.py
normanb: 👍

tests/test_geospatial.py
normanb: 👍

normanb:
Added dynamic node expansion from the result of a previous UDF when ingesting the data.

src/tiledb/cloud/geospatial/ingestion.py
normanb: 👍

normanb:
@thetorpedodog Can you help with https://github.com/TileDB-Inc/TileDB-Cloud-Py/pull/507#discussion_r1493017230 so that the tests pass? We are missing the geospatial library dependencies.

thetorpedodog:
> @thetorpedodog Can you help with [#507 (comment)](https://github.com/TileDB-Inc/TileDB-Cloud-Py/pull/507#discussion_r1493017230) so that the tests pass? We are missing the geospatial library dependencies.

The thing to do here would be to add a new `optional-dependencies` section with the geospatial libraries:

https://github.com/TileDB-Inc/TileDB-Cloud-Py/blob/09240abf512ff6636ad2aa036b2cd7c2bde70cc7/pyproject.toml#L25-L28

Then, we can introduce a new workflow into the GitHub CI that runs just the geospatial tests with those extra dependencies.

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1410 @@
...
+                    log_uri=log_uri,
+                    config=config,
+                    max_files=max_files,
+                )
+
+            if search_uri:
+                sources = find(
+                    search_uri,
+                    config=config,
+                    excludes=ignore,
```
Shelnutt2, line 999: Typo should be `exclude`?

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1410 @@
...
+                    includes=pattern if pattern else fns[dataset_type]["pattern_fn"],
```
Shelnutt2, line 1000: Typo should be `include`?

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1410 @@
...
+    """
+    Ingest samples into a dataset.
+
+    :param dataset_uri: dataset URI
+    :param dataset_type: dataset type, one of pointcloud, raster or geometry
+    :param acn: Access Credentials Name (ACN) registered in TileDB Cloud (ARN type),
+        defaults to None
+    :param config: config dictionary, defaults to None
+    :param namespace: TileDB-Cloud namespace, defaults to None
+    :param register_name: name to register the dataset with on TileDB Cloud,
```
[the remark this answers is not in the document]
normanb: the destination array is not registered, I have updated the docstring.

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1410 @@
...
+        config,
+        resources=DEFAULT_RESOURCES,
+        verbose=verbose,
+        trace=trace,
+        log_uri=log_uri,
+        access_credentials_name=acn,
+    ).depends_on(process_node)
+
+    # Register the dataset on TileDB Cloud
+    if register_name:
```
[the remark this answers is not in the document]
normanb: The `dataset_uri` can be a `tiledb://` uri and work. This way we can write directly to object storage if needed. This is following VCF but happy to restrict this to just `tiledb://` URIs.

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1410 @@
...
+        register_dataset_udf(
```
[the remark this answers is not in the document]
normanb: Yes, this should be part of the graph, I am fixing all of the comments above related to this function.

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1410 @@
...
+            dataset_uri,
+            namespace=namespace,
+            register_name=register_name,
+            config=config,
+            verbose=verbose,
+            trace=trace,
```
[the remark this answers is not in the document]
normanb: 👍

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1410 @@
...
+            log_uri=log_uri,
```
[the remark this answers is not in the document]
normanb: :thumbs

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1410 @@
...
+            access_credentials_name=acn,
```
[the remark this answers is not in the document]
normanb: 👍

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1410 @@
...
+    resampling: Optional[str] = "bilinear",
+    res: Tuple[float, float] = None,
+    verbose: bool = False,
+    trace: bool = False,
+    log_uri: Optional[str] = None,
+) -> dict[str, object]:
+    """Groups input URIs into batches.
+    :param dataset_uri: dataset URI
+    :param dataset_type: dataset type, one of pointcloud, raster or geometry
+    :param acn: Access Credentials Name (ACN) registered in TileDB Cloud (ARN type),
```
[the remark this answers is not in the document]
normanb: 👍

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1410 @@
...
+        mode=tiledb.cloud.dag.Mode.BATCH,
+        max_workers=workers,
+        namespace=namespace,
+    )
+
+    # find sources
+    input_list_node = graph.submit(
+        build_inputs_udf,
+        dataset_type=dataset_type,
+        acn=acn,
```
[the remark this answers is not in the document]
normanb: 👍

src/tiledb/cloud/geospatial/ingestion.py
normanb: 👍

src/tiledb/cloud/geospatial/ingestion.py
normanb: 👍

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1410 @@
...
+                    max_files=max_files,
+                )
+
+            if search_uri:
+                sources = find(
+                    search_uri,
+                    config=config,
+                    excludes=ignore,
+                    includes=pattern if pattern else fns[dataset_type]["pattern_fn"],
+                    max_files=max_files,
```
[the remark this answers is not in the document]
normanb: yes, it is, I missed this change when writing the `find` function.

src/tiledb/cloud/utilities/_common.py
```diff
@@ -347,3 +382,94 @@ def wrapper(*args, **kwargs) -> Dict[str, object]:
...
+        current_count = 0
+
+        def list_files(listing):
+            for f in listing:
+                # Avoid infinite recursion
+                if f == uri:
+                    continue
+
+                if vfs.is_dir(f):
+                    yield list_files(
```
Shelnutt2, line 417: This should be `yield from` and I believe we want to call `find` not `list_files` to recurse into the sub-folders. This means we also need to pass all the parameters, `config=config, max_count=max_count`.

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1405 @@
...
+                    # only open header
+                    las = laspy.open(src)
+                    hdr = las.header
+                    mins = _fold_in(min, mins, hdr.mins)
+                    maxs = _fold_in(max, maxs, hdr.maxs)
+                    scales = _fold_in(max, scales, hdr.scales)
+                    offsets = _fold_in(min, offsets, hdr.offsets)
+
+            return GeoMetadata(
+                paths=sources,
```
Shelnutt2, line 149: This needs to be `list(sources)`

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1405 @@
...
+                    exclude=ignore,
+                    include=pattern if pattern else fns[dataset_type]["pattern_fn"],
+                    max_count=max_files,
+                )
+
+            if not sources:
+                raise ValueError(f"No {dataset_type.name} datasets found")
+
+            meta_kwargs = fns[dataset_type]["kwargs"]
+            meta = fns[dataset_type]["meta_fn"](sources, config=config, **meta_kwargs)
```
Shelnutt2, line 1005: So in testing this on some larger scale use cases. Getting the metadata in-line is a problem. I believe we should batch the output into jobs of 10 or 100 files, and grab metadata in parallel in another stage of the task graph. I've seen it take over 30m when there is 6000 files.

src/tiledb/cloud/geospatial/ingestion.py
Shelnutt2: Actually we should convert source to list before we consume the generator. something like:
```paths = list(sources)
for f in paths:
...
...
..
return GeoMetadata(
    paths=paths
...
...
```

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -0,0 +1,1405 @@
...
+        verbose=verbose,
+        trace=trace,
+        log_uri=log_uri,
+        name=f"{dag_name} input collector",
+    )
+
+    # schema creation node, returns a sequence of work items
+    ingest_node = graph.submit(
+        fn,
+        **input_list_node,
```
Shelnutt2, line 1188: `input_list_node` is a `Node` type, you can't deconstruct the dictionary that is returned from `input_list_node` like this. The dictionary needs to be passed in-intact and handles inside the functions.

src/tiledb/cloud/utilities/_common.py
normanb: I will add a test case. For `find` and not `list_files`, we retain the current count as a check in the outer function. With `list_files` I have removed the extraneous arguments.

src/tiledb/cloud/geospatial/ingestion.py
normanb: 👍

src/tiledb/cloud/geospatial/ingestion.py
normanb: I think we should change this after this PR has been merged. There are a couple of scalability improvements we can make.

src/tiledb/cloud/geospatial/ingestion.py
normanb: 👍

src/tiledb/cloud/geospatial/ingestion.py
normanb: Also just noting that geospatial datasets tend to be large as opposed to thousands of files. We should anticipate the problem but I would like to make this change as an incremental PR.

Shelnutt2: Sure thing, thanks!

normanb:
The latest code has been tested with rasters and point clouds on TileDB Cloud.
`pyopener` is required for geometries in object storage which will be added in a separate PR but the code is stubbed out for geometry ingestion.
This PR inlines the new utility functions, after the PR is merged another PR will be issued to remove them and use the updated tiledb-cloud-py.

tests/common/test_utils.py
```diff
@@ -42,6 +46,39 @@ def test_datetime_to_msec(self):
...
+    def test_find(self):
+        test_1 = [
+            self.fs.create_file("/var/data.dat"),
+            self.fs.create_file("/var/data/xx1.txt"),
+            self.fs.create_file("/var/data/xx2.txt"),
+        ]
+
+        with mock.patch.object(VFS, "ls", return_value=test_1):
+            with mock.patch.object(VFS, "is_dir", return_value=True) as mock_is_dir:
+                mock_is_dir.side_effect = lambda f: self.fs.isdir(f.name)
```
thetorpedodog, line 62: Though I did notice this while scrolling through: Instead of setting up a bunch of mocks and using a fake filesystem, it seems like it would be simpler to create a temporary directory:

```python
with tempfile.TemporaryDirectory() as tmp_name:
    tmp = pathlib.Path(tmp_name)
    (tmp / "data.dat").write_text("test")
    data_dir = tmp / "data"
    data_dir.mkdir()
    (data_dir / "xx1.txt").write_text("test")
    ...
```

The temporary directory will be cleaned up at the end of the with-block, and there is no need to muck with TileDB internals.

tests/test_geospatial.py
```diff
@@ -0,0 +1,418 @@
...
+import unittest
+import warnings
+from pathlib import Path
+from unittest import mock
+
+import affine
+import fiona
+import numpy as np
+import rasterio
+import shapely
```
thetorpedodog, line 15: You'll need to move the imports of these into only the methods where they are directly used, to avoid having them at the module level.

tests/test_geospatial.py
```diff
@@ -0,0 +1,418 @@
...
+from tiledb.cloud.geospatial import BoundingBox
+from tiledb.cloud.geospatial import DatasetType
+from tiledb.cloud.geospatial import build_inputs_udf
+from tiledb.cloud.geospatial import ingest_geometry_udf
+from tiledb.cloud.geospatial import ingest_point_cloud_udf
+from tiledb.cloud.geospatial import ingest_raster_udf
+from tiledb.cloud.geospatial import load_geometry_metadata
+from tiledb.cloud.geospatial import load_pointcloud_metadata
+from tiledb.cloud.geospatial import load_raster_metadata
+from tiledb.cloud.geospatial import read_uris
```
thetorpedodog, line 27: Likewise with this—importing `tiledb.cloud.geospatial` will end up importing those libraries by proxy, so within the function itself, import the items. To avoid all the repetition, I recommend importing the module rather than anything inside it, so…

tests/test_geospatial.py
```diff
@@ -0,0 +1,418 @@
...
+            **new_kwargs,
+        )
+
+
+class GeospatialTest(unittest.TestCase):
+    def setUp(self):
+        # Ignore warnings
+        warnings.simplefilter("ignore")
+        # Create a temporary directory
+        self.test_dir = Path(tempfile.mkdtemp())
```
thetorpedodog, line 96: Another sidenote:

Something like

```python
self.tempdir_obj = tempfile.TemporaryDirectory()
self.test_dir = pathlib.Path(self.tempdir.name)
```
can be used here, then…

tests/test_geospatial.py
```diff
@@ -0,0 +1,418 @@
...
+
+        create_test_geometries(self.test_dir)
+        for g in GEOM_NAMES:
+            self.assertTrue(os.path.exists(self.test_dir.joinpath(g)))
+
+        self.out_path = "out"
+
+    def tearDown(self):
+        # Remove the directory after the test
+        shutil.rmtree(self.test_dir)
```
thetorpedodog, line 173: …`self.tempdir_obj.cleanup()` will do all the cleanup work for you.

tests/test_geospatial.py
```diff
@@ -0,0 +1,418 @@
...
+
+class GeospatialTest(unittest.TestCase):
+    def setUp(self):
+        # Ignore warnings
+        warnings.simplefilter("ignore")
+        # Create a temporary directory
+        self.test_dir = Path(tempfile.mkdtemp())
+
+        def create_test_geometries(tmp_path: os.PathLike):
+            radius = 1.0
```
thetorpedodog, line 99: add `import shapely` to the top of this function rather than having that import at the module level.

Additionally, since this doesn’t appear to depend upon anything in its closure (i.e., the only thing it uses is the `tmp_path` passed in as a parameter), I recommend moving it to the module level so that the `setUp` function isn’t so long. Likewise with similar functions.

tests/test_geospatial.py
```diff
@@ -0,0 +1,418 @@
...
+            for i in range(1, 4):
+                g = shapely.Point(i, i).buffer(radius)
+                with open(tmp_path.joinpath(GEOM_NAMES[i - 1]), "w") as dst:
+                    json.dump(shapely.geometry.mapping(g), dst)
+
+        def create_test_rasters(tmp_path: os.PathLike):
+            kwargs = {
```
thetorpedodog, line 106: Likewise, at the top of this function, do `import affine; import rasterio`, etc.

This means that when this file is imported by `pytest`, it won’t immediately import any of the libraries, and by default all these tests will be skipped.

tests/test_geospatial.py
```diff
@@ -0,0 +1,418 @@
...
+    for _, work in enumerate(samples):
+        fn(
+            sources=work,
+            dataset_uri=dataset_uri,
+            append=True,
+            **new_kwargs,
+        )
+
+
+class GeospatialTest(unittest.TestCase):
```
thetorpedodog, line 91: You can mark this test with `@pytest.mark.geospatial`, then

thetorpedodog (commented):
To set your tests up so that they only run when geospatial tests are called for:

Part 1 is already complete—you can use `@pytest.mark.geospatial` on your test cases that use geospatial code. This ensures that they don’t *run* when not called for. However, they are still collected as part of the test discovery process, which means the test Python files themselves should not need things installed.

I will add notes on the applicable files below, but don’t have time to do a full read-through of the codebase.

tests/common/test_utils.py
normanb: I agree for `find` this is easier than mocking the file system.

sgillies (approved):

PR #508, opened 2024-01-30
Registration upon ingestion

ktsitsi:
This PR:

- Introduces a flag `register`, which when `True` creates an extra node in the ingestion graph for registering the ingested data to TileDB-Cloud. The behaviour supported for now is that the registration name will be the filename of the ingested data. 

- This implementation comes as an extra pathway as until now the following was also feasible if the destination path was a tiledb URI.

Upon Merging this PR, the following will be also covered as the initial base branch was #509 .
#509 #505


src/tiledb/cloud/bioimg/ingestion.py
```diff
@@ -24,6 +25,8 @@ def ingest(
     threads: Optional[int] = 8,
     resources: Optional[Mapping[str, Any]] = None,
     compute: bool = True,
+    register: bool = False,
```
Shelnutt2, line 28: Register should be the default.

Shelnutt2 (approved):

Shelnutt2 (changes requested):
Request we change the defaults.

src/tiledb/cloud/bioimg/helpers.py
```diff
@@ -40,83 +41,29 @@ def is_folder(path: str) -> bool:
...
-                    raise ValueError(
-                        "For multiple inputs in source list \
-                        the inputs cannot contain a dir"
-                    )
-                if not is_folder(output):
-                    raise ValueError(
-                        "For multiple inputs the output should \
-                        correspond to a dir with a trailing '/'"
-                    )
+def validate_io_paths(source: Sequence[str], output: Sequence[str]):
```
ihnorton, line 44: Seems a bit cleaner now, thanks.

src/tiledb/cloud/bioimg/ingestion.py
```diff
@@ -88,33 +97,47 @@ def iter_paths(source: Sequence[str], output: Sequence[str]) -> Iterator[Tuple]:
...
+        )
         # If the user didn't specify a number of batches, run every import
         # as its own task.
+        logger.debug(f"The io pairs for ingestion: {uri_pairs}")
         my_num_batches = num_batches or len(uri_pairs)
         # If they specified too many batches, don't create empty tasks.
         my_num_batches = min(len(uri_pairs), my_num_batches)
-        return [uri_pairs[n::my_num_batches] for n in range(my_num_batches)]
+        batches = [uri_pairs[n::my_num_batches] for n in range(my_num_batches)]
+        logger.debug(f"Splitted batchs : {batches}")
```
ihnorton, line 139: `Splitted` -> `Split`?

ihnorton (approved):

src/tiledb/cloud/bioimg/helpers.py
```diff
@@ -34,3 +35,35 @@ def serialize_filter(filter):
...
+    if len(source) == 0:
+        raise ValueError("Source list must not be empty.")
+
+    # Case 1: If source is a list of files
+    if all(not is_folder(path) for path in source):
+        # Destination is a folder
+        if len(output) == 1 and is_folder(output[0]):
+            return True
+        # Destination has the same number of paths with source - rename at destination
+        elif len(output) == len(source) and all(not is_folder(path) for path in output):
```
thetorpedodog, line 57: nit: no else when the other branch unconditionally returns

src/tiledb/cloud/bioimg/helpers.py
```diff
@@ -34,3 +35,35 @@ def serialize_filter(filter):
...
+        raise ValueError("Both source and output must be lists.")
+
+    if len(source) == 0:
+        raise ValueError("Source list must not be empty.")
+
+    # Case 1: If source is a list of files
+    if all(not is_folder(path) for path in source):
+        # Destination is a folder
+        if len(output) == 1 and is_folder(output[0]):
+            return True
```
thetorpedodog, line 55: this looks like it either returns True or raises; in that case I would make this `-> None` and use `return` rather than `return True`

src/tiledb/cloud/bioimg/ingestion.py
```diff
@@ -150,19 +186,84 @@ def ingest_tiff_udf(
...
+            register_name = os.path.splitext(os.path.basename(output_path))[0]
+            register_map[output_path] = register_name
+            tiledb_uris[output_path] = f"tiledb://{namespace}/{register_name}"
+
+        with tiledb.scope_ctx(config):
+            for dataset_uri, tiledb_uri in tiledb_uris.items():
+                found = False
+                try:
+                    object_type = tiledb.object_type(tiledb_uri)
+                    logger.debug(f"Object type of {tiledb_uri} : {object_type}")
```
thetorpedodog, line 228: `"Object of type %s: %s", tiledb_uri, object_type`

The idea behind doing it that way is so that it doesn't have to go through the work of constructing the output string if it's not going to be logged. This isn't a huge deal here, but it can make a difference when logging in hot loops of complex operations.

tests/test_bioimg.py
```diff
@@ -0,0 +1,94 @@
+import unittest
+
+from tiledb.cloud.bioimg.helpers import validate_io_paths
+
+
+class BioimgTest(unittest.TestCase):
+    def setUp(self) -> None:
+        self.out_path = "out"
+        self.accepted_pairs = {
```
thetorpedodog, line 9: These look like they can just be variables in the `test_validate_io_paths` function, avoiding the need for a separate `setUp` function.

tests/test_bioimg.py
```diff
@@ -0,0 +1,94 @@
...
+            ),
+            # Multiple files in multiple files out - equal length
+            # a -> b & c -> d
+            "test15": (
+                ["s3://test_in/a.tiff", "s3://test_in/c.tiff"],
+                ["s3://test_out/b", "s3://test_out/d"],
+            ),
+        }
+
+        return super().setUp()
```
thetorpedodog, line 52: this line is commonly forgotten! :)

src/tiledb/cloud/bioimg/helpers.py
```diff
@@ -34,3 +35,35 @@ def serialize_filter(filter):
...
     else:
         raise TypeError
+
+
+def is_folder(path: str) -> bool:
+    return path.endswith("/")
+
+
+def validate_io_paths(source: Sequence[str], output: Sequence[str]):
+    if not isinstance(source, list) or not isinstance(output, list):
```
thetorpedodog, line 45: instead of `not isinstance(source, list)`, it's probably better to check for `isinstance(source, str)`, so that you're not *exclusively* allowing lists (e.g., you could pass a tuple).

tests/test_bioimg.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def test_validate_io_paths(self):
+        # Accepted cases
+        for test_name, io_tuple in self.accepted_pairs.items():
+            source, dest = io_tuple
+            source = [source] if isinstance(source, str) else source
+            dest = [dest] if isinstance(dest, str) else dest
+            print(f"{test_name}:", source, dest)
+            validate_io_paths(source, dest)
```
thetorpedodog, line 61: you can use `subTest` to isolate each of the sub-tests so that if one fails the rest still run:

```python
with self.subTest(test_name):
    source, dest = io_tuple
```

You also might be able to write the loop as
```python
for test_name, (source, dest) in self.accepted_pairs.items():
  ...
```

Also, since the function now only accepts lists, I would remove the pre-processing list-wrapping step and only pass in the values that you directly want into the function.

src/tiledb/cloud/bioimg/helpers.py
ktsitsi: Will totally remove after all before the validation we cast to list if the input is a str so the check is not useful.

src/tiledb/cloud/bioimg/ingestion.py
```diff
@@ -132,13 +166,10 @@ def ingest_tiff_udf(
...
 
-        converter = kwargs.get("converter", None)
         user_converter = Converters.OMETIFF
-        if not converter or converter == "tiff":
-            user_converter = Converters.OMETIFF
-        elif converter == "zarr":
+        if converter == "zarr":
             user_converter = Converters.OMEZARR
-        elif converter == "osd":
+        if converter == "osd":
```
thetorpedodog, line 172: you could reduce repetition with an internal dict here:

```python
user_converter = {
    "zarr": converters.OMEZARR,
    "osd": converters.OSD,
}.get(converter, converters.OMETIFF)
```

thetorpedodog (approved):
One minor suggestion. Looks good!

PR #511, opened 2024-02-07
Completely drop support for Python 3.7.

thetorpedodog:
- We no longer mark ourselves as Python 3.7 compatible.
- We no longer test against Python 3.7 locally or server-side.
- Python 3.7–related code and warnings are dropped.

Future Python 3.7 users will still be able to install the old version of the tiledb-cloud package, but new versions are fully incompatible.


ihnorton (approved):
LGTM. Needs rebase.

PR #513, opened 2024-02-14
Hotfix for cloud 1-Click ingestion go args

ktsitsi:


thetorpedodog:
This looks fine without any context, but with the context I don't follow this—it looks like the task is to remove `**kwargs` and this adds `**kwargs`.

ktsitsi:
> This looks fine without any context, but with the context I don't follow this—it looks like the task is to remove `**kwargs` and this adds `**kwargs`.

You are right @thetorpedodog sorry for the misleading title. At first indeed I thought of removing the kwargs from passing in the register_udf but then I thought that enriching the udf with kwargs even though that are not used inside of it, would be future proof in case some kwargs eventually reach this register UDF and become useful at some point. They don't affect it in any case.

thetorpedodog (approved):

PR #520, opened 2024-02-21
Fixing 1-Click ingestion `access_credentials_name` kwargs

ktsitsi:
This PR:

- modifies the function signature for ingest so `access_credentials_name` is not a positional argument but a kwarg. 
This way the following line in wrapper:
```python
acn = kwargs.get("acn", kwargs.get("access_credentials_name", None))
```
will be allowed to set it in the `acn` variable and feed it to the `func` as a kwarg argument explicitly. Due to the fact that `access_credentials_name` will not be part of the original `ingest` signature the `_filter_kwargs` will filter it out and won't be given as a duplicate argument in the `graph.submit`. Also with this change the `access_credentials_name` will not be a positional argument anymore and thus we will avoid the error from inside the wrapper.
```bash
TypeError: ingest() missing 1 required positional argument: 'access_credentials_name'
```

**The following PR should be tested in staging using the QA notebook.**

### DIFF
--- a/src/tiledb/cloud/bioimg/ingestion.py
+++ b/src/tiledb/cloud/bioimg/ingestion.py
@@ -23,6 +23,7 @@ def ingest(
     output: Union[Sequence[str], str],
     config: Mapping[str, Any],
     *args: Any,
+    access_credentials_name: str,
     taskgraph_name: Optional[str] = None,
     num_batches: Optional[int] = None,
     threads: Optional[int] = 0,
@@ -298,8 +299,7 @@ def ingest(
         ),
     )
 
-    acn = kwargs.pop("access_credentials_name", None)
-    if not acn:
+    if not access_credentials_name:
         raise ValueError(
             "Ingestion graph requires `access_credentials_name` to be set."
         )
@@ -313,7 +313,7 @@ def ingest(
         _SUPPORTED_EXTENSIONS,
         *args,
         verbose=verbose,
-        access_credentials_name=acn,
+        access_credentials_name=access_credentials_name,
         name=f"{dag_name} input collector",
         result_format="json",
     )
@@ -337,7 +337,7 @@ def ingest(
         image_name=DEFAULT_IMG_NAME,
         max_workers=threads,
         compressor=compressor_serial,
-        access_credentials_name=acn,
+        access_credentials_name=access_credentials_name,
         **kwargs,
     )
 
@@ -347,7 +347,7 @@ def ingest(
             ingest_list_node,
             config=config,
             verbose=verbose,
-            acn=acn,
+            acn=access_credentials_name,
             namespace=namespace,
             name=f"{dag_name} registrator ",
             expand_node_output=ingest_list_node,
--- a/src/tiledb/cloud/bioimg/ingestion.py
+++ b/src/tiledb/cloud/bioimg/ingestion.py
@@ -22,7 +22,6 @@ def ingest(
     source: Union[Sequence[str], str],
     output: Union[Sequence[str], str],
     config: Mapping[str, Any],
-    access_credentials_name: str,
     *args: Any,
     taskgraph_name: Optional[str] = None,
     num_batches: Optional[int] = None,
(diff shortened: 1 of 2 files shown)


src/tiledb/cloud/bioimg/ingestion.py
```diff
@@ -22,7 +22,6 @@ def ingest(
     source: Union[Sequence[str], str],
     output: Union[Sequence[str], str],
     config: Mapping[str, Any],
-    access_credentials_name: str,
```
Shelnutt2, line 25: This should still be defined as a parameter to the function. Let's not hide things behind generic kwargs.

Shelnutt2 (dismissed):

src/tiledb/cloud/bioimg/ingestion.py
Shelnutt2: I believe all that is required is setting the type to.optional with a default of None?

ktsitsi: Nope this is not sufficient. Type checking in python is nothing more than just syntax sugar. The alternative I pushed is to keep it as a non positional argument and pop from the wrapper instead of just getting it.

Shelnutt2: Type checking is used by language servers and linters, I'd like us to keep the argument as a parameter. It can be handled, we do this in other functions. Please consult with @thetorpedodog on how we can handle this.

src/tiledb/cloud/utilities/_common.py
```diff
@@ -314,7 +314,7 @@ def wrapper(*args, **kwargs) -> Dict[str, object]:
 
         name = kwargs.get("name", func.__name__)
         namespace = kwargs.get("namespace", None)
-        acn = kwargs.get("acn", kwargs.get("access_credentials_name", None))
+        acn = kwargs.get("acn", kwargs.pop("access_credentials_name", None))
```
thetorpedodog, line 317: Why this change?

src/tiledb/cloud/bioimg/ingestion.py
thetorpedodog: Defining it this way means it remains a full-fledged parameter, just one that _has_ to be passed as a named argument. Language servers and linters are able to handle this, and moving the name after `*args` is the right way to accomplish it.

```python

def some_func(first, this_arg, *args, that_arg="z"): ...

# In both of these cases, first="f", this_arg="a", args=(1, 2, 3), that_arg="z".
some_func("f", "a", 1, 2, 3)
some_func("f", 1, 2, 3, this_arg="a")

# first="f", this_arg=1, args=(2, 3), that_arg="b"
some_func("f", 1, 2, 3, that_arg="b")

def other_func(first, *args, this_arg, that_arg="z"): ...

# Neither of these is allowed, since `this_arg` must be explicitly named.
other_func("f", "a", 1, 2, 3)
other_func("f", 1, 2, 3, that_arg="b")

# This one works, with first="f", this_arg="a", args=(1, 2, 3), that_arg="z"
other_func("f", 1, 2, 3, this_arg="a")
```

ktsitsi: In my last commit I returned the access_credentials_name as a named argument as @Shelnutt2 recommended. My answer to @Shelnutt2 was about the problem that setting it to `Optional` or setting a default value does not resolve the issue I am trying to solve. Hence I added the `pop`. 
```python
import inspect
def some_func(first, this_arg, *args, that_arg="z"):
    print(a)
valid_args = inspect.signature(some_func).parameters
valid_args
>>>mappingproxy({'first': <Parameter "first">,
              'this_arg': <Parameter "this_arg">,
              'args': <Parameter "*args">,
              'that_arg': <Parameter "that_arg='z'">})
```
As you notice even when passed as a named argument it is returned from the `inspect` and this where the issue lies. _filter_dict function does not filter out the `access_credentials_name` and this is fed to `graph.submit`, where we already explicitly set the `access_credentials_name`.

thetorpedodog (approved):
Sorry for not coming back to this sooner! This looks good.

Shelnutt2 (approved):

PR #521, opened 2024-02-23
Improve UX when querying annotation VCFs

gspowley:
When querying a sample-less annotation VCF file with `tiledb.cloud.vcf.read`, the `samples` parameter must be provided.

Prior to this PR, the samples had to be provided as a list:
```python
samples=[""],
```

This PR allows the samples to also be provided as a string:
```python
samples="",
```

Input validation is also improved with more useful feedback to the user.


leipzig (approved):
yep

ckrilow (approved):
Looks great, I really like this functionality.

PR #522, opened 2024-02-27
Parallelize Cloud tests.

JohnMoutafis:
## Method
- Uses [pytest-split](https://pypi.org/project/pytest-split/)
- Splits tests in 6 parrallel groups
- Utilizes `.test_durations` file produced by [pytest-split](https://pypi.org/project/pytest-split/) to split the groups in roughly similar duration groups

### Steps
- Before the tests begin [attempt to fetch the `.test_durations` file from the actions cache](https://github.com/TileDB-Inc/TileDB-Cloud-Py/blob/ed59d7d8e1a459a0029bcba34ec7fbfff738ad3b/.github/workflows/tiledb-cloud-py.yaml#L110) for each group.
  - If this step fails to retrieve a caches object, tests will revert to the [`.test_durations` file stored in the repo](https://github.com/TileDB-Inc/TileDB-Cloud-Py/blob/ed59d7d8e1a459a0029bcba34ec7fbfff738ad3b/.test_durations).
- [Run the grouped tests](https://github.com/TileDB-Inc/TileDB-Cloud-Py/blob/ed59d7d8e1a459a0029bcba34ec7fbfff738ad3b/.github/workflows/tiledb-cloud-py.yaml#L123) and [store the updated durations for each](https://github.com/TileDB-Inc/TileDB-Cloud-Py/blob/ed59d7d8e1a459a0029bcba34ec7fbfff738ad3b/.github/workflows/tiledb-cloud-py.yaml#L131)
- Upload each groups updated test durations [as artifacts](https://github.com/TileDB-Inc/TileDB-Cloud-Py/blob/ed59d7d8e1a459a0029bcba34ec7fbfff738ad3b/.github/workflows/tiledb-cloud-py.yaml#L139) with a name papttern: `partial-test-durations-<group-number>`
- After the tests have passed, [download the partial-duration artifacts](https://github.com/TileDB-Inc/TileDB-Cloud-Py/blob/ed59d7d8e1a459a0029bcba34ec7fbfff738ad3b/.github/workflows/tiledb-cloud-py.yaml#L160) and [combine them into a single `.test_durations` file](https://github.com/TileDB-Inc/TileDB-Cloud-Py/blob/ed59d7d8e1a459a0029bcba34ec7fbfff738ad3b/.github/workflows/tiledb-cloud-py.yaml#L167) and then [store it to cache for the next run](https://github.com/TileDB-Inc/TileDB-Cloud-Py/blob/ed59d7d8e1a459a0029bcba34ec7fbfff738ad3b/.github/workflows/tiledb-cloud-py.yaml#L172).

## Issues created with this PR
- The PR removes testing for `macos` because there is [a hard limit of 5 concurrent instances running at once](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration) and that lead to queuing times of up to 3:30 hours in some cases.
- `macos` testing should be transfered in a cronjob/daily test for sanity check that we still support `macos` systems without issues.
- A `squash-merge` is necessary as it needed to many commits to test behavior.


.github/workflows/tiledb-cloud-py.yaml
```diff
@@ -107,10 +120,27 @@ jobs:
...
+            --store-durations --splitting-algorithm least_duration --durations-path=.test_durations.${{ matrix.pytest-split-group }} --clean-durations
         env:
           TILEDB_REST_TOKEN: ${{ secrets.TILEDB_CLOUD_HELPER_VAR }}
 
+  save-test-durations-to-cache:
+    needs: run-tests
+    runs-on: ubuntu-latest
+    continue-on-error: true
+    steps:
+      - name: "Combine test duration cache files in .test_durations"
```
ihnorton, line 134: - Do we need to update this in every run?
- Is the action here going to commit back to `.test_durations` on this branch, or is that just a cache key?

.github/workflows/tiledb-cloud-py.yaml
```diff
@@ -107,10 +120,27 @@ jobs:
             pytest -sv \
             --last-failed \
             --tb short \
-            --color yes
+            --color yes \
+            --splits ${{ env.PYTEST_SPLIT_GROUPS }} --group ${{ matrix.pytest-split-group }} \
+            --store-durations --splitting-algorithm least_duration --durations-path=.test_durations.${{ matrix.pytest-split-group }} --clean-durations
```
thetorpedodog, line 125: Please put these one-per-line for easier reading.

.test_durations
thetorpedodog, line 1: This seems like it should not be committed.

.github/workflows/tiledb-cloud-py.yaml
thetorpedodog: My understanding here:

- this saves the timings from the latest run to the cache, and does not modify what is in the repository
- we do this so pytest has an idea of how to partition tests so that the N shards run in about the same length of time

It seems reasonable from a "how" perspective, but having a better explanation in a comment would be useful.

It also seems like maybe we should include the matrix variables in the cache key, so that all the different matrix entries aren’t continually overwriting each other’s timing values.

.test_durations
JohnMoutafis: I am still trying to fine tune the behavior of the tests. Is there another way to dry-run/test the actions without commiting?

.github/workflows/tiledb-cloud-py.yaml
JohnMoutafis: Will do

.github/workflows/tiledb-cloud-py.yaml
JohnMoutafis: @ihnorton @thetorpedodog  I will update the commit with a detailed explanation when it is ready for a final review.

.github/workflows/tiledb-cloud-py.yaml
```diff
@@ -11,6 +11,10 @@ on:
     # 01:30 UTC = 20:30/21:30 America/New_York = 03:30/04:30 Europe/Athens
     - cron: "30 1 * * *"
 
+env:
+  PYTEST_SPLIT_GROUPS: 6 # How many parallel groups to create for the tests.
+  PYTEST_TEST_DURATIONS_CACHE_KEY: "test_durations"
```
thetorpedodog, line 16: Since these variables are specific to the run-tests pipeline, I think it would make sense to define them as part of that job.

.github/workflows/tiledb-cloud-py.yaml
JohnMoutafis: @ihnorton @thetorpedodog I have updated the PR text to explain the process.

ihnorton (approved):

.test_durations
thetorpedodog, line 1: I have one remaining question: Since we will have these test durations in the cache, is this file necessary? There will be a one-time bootstrapping cost but after that it seems like we should be able to rely exclusively on the cache. (This also reduces the maintenance burden—no need to manually update the file when tests are added/removed/changed.)

JohnMoutafis: Split doesn't ultimately need the file to work for the first time.
If the file is not present, it will just split into equal size groups.
So I can remove it!

thetorpedodog (approved):

PR #524, opened 2024-02-27
Append-mode SOMA ingest [WIP]

johnkerl:
This PR is not ready for review. Please do not review it.

Also, this needs #512 as a prerequisite.


johnkerl:
Please see Linear for current ownership.

PR #525, opened 2024-02-28
Update registration name when multiple files present

ktsitsi:
This PR:

- Fixes a bug that came up during a batch registration. Although the registration name mapping was being constructed, the registration name was not getting updated according to the mapping. 
- The bug was not visible in the cases where even multiple files are getting registered but each task is being given just one io pair from the previous step.

### DIFF
--- a/src/tiledb/cloud/bioimg/ingestion.py
+++ b/src/tiledb/cloud/bioimg/ingestion.py
@@ -262,10 +262,15 @@ def ingest(
                 if found:
                     logger.info("Dataset already registered at %r.", tiledb_uri)
                 else:
-                    logger.info("Registering dataset %s at %s", dataset_uri, tiledb_uri)
+                    logger.info(
+                        "Registering dataset %s at %s with name %s",
+                        dataset_uri,
+                        tiledb_uri,
+                        register_map[dataset_uri],
+                    )
                     tiledb.cloud.groups.register(
                         dataset_uri,
-                        name=register_name,
+                        name=register_map[dataset_uri],
                         namespace=namespace,
                         credentials_name=acn,
                     )


gspowley (approved):

src/tiledb/cloud/bioimg/ingestion.py
```diff
@@ -262,10 +262,15 @@ def register_dataset_udf(
                 if found:
                     logger.info("Dataset already registered at %r.", tiledb_uri)
                 else:
-                    logger.info("Registering dataset %s at %s", dataset_uri, tiledb_uri)
+                    logger.info(
+                        "Registering dataset %s at %s with name %s",
+                        dataset_uri,
+                        tiledb_uri,
+                        register_map[dataset_uri],
+                    )
```
sgillies, line 270: @ktsitsi congratulations on finding and fixing the bug!

I've got a suggestion about logging, for next time. As use of this package grows in scale, we'll eventually start using sophisticated tools for analysis, and then more structured logging will help us a lot. At one extreme of structured logging are practitioners who advocate using JSON formatted log messages, like here: https://newrelic.com/blog/how-to-relic/structured-logging. I've found that what [structlog does](https://www.structlog.org/en/stable/why.html) is good enough. The gist of structlog's structured logging is replacing a prose log message that reads like a sentence, like

`Registering dataset X at Y with name Z`

With one containing an event name and keys and values, like

`Registering dataset: dataset_uri='X', tiledb_uri='Y', register_map={...}`

If we do that, we'll have log messages that are easier to parse and analyse. Maybe easier for non-native English speakers to read and write, too?

But at the very least, I recommend using `%r` in log formatting because `%s` can hide some important details, like whether strings are unicode or binary.

antalakas:
[sc42264]

PR #526, opened 2024-03-01
Add pytest-explicit to skip geospatial tests by default.

thetorpedodog:
This change adds the `pytest-explicit` pytest plugin, which allows us to mark certain tests as "explicit-only", i.e., they should only be run when specifically requested to. This will initially be used to mark geospatial tests, which require specific libraries to be installed.

---

This is preparation for https://github.com/TileDB-Inc/TileDB-Cloud-Py/pull/507.

### DIFF
--- a/.github/workflows/tiledb-cloud-py.yaml
+++ b/.github/workflows/tiledb-cloud-py.yaml
@@ -88,7 +88,14 @@ jobs:
 
       - name: Install prerequisites
         run: |
-          pip install --upgrade pip wheel pytest pytest-cov setuptools setuptools-scm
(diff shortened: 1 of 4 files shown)


normanb (approved):
tests/common/test_pickle_compat.py seems unrelated, but 👍

thetorpedodog:
> tests/common/test_pickle_compat.py seems unrelated, but 👍

That’s intended so that you can run `pytest` even if you don’t have `xarray` installed. It’s not directly related (and in fact I think I intended to not have it in this change) but it’s there now and it doesn’t hurt so it can stay.

PR #528, opened 2024-03-05
Supporting tiledb-bioimg 0.2.10 new tile_scale arg

ktsitsi:
This PR:

- Introduces a new argument to the bioimg ingestor for supporting the tiledb-bioimg [0.2.10](https://github.com/TileDB-Inc/TileDB-BioImaging/releases/tag/v0.2.10). The latter will allow a more efficient way to ingest + register using TileDB URI directly at the ingestion stage. 
- This change requires the update of TileDB-Bioimg package to UDF docker images and JupyterHub from 0.2.9 to 0.2.10.

### DIFF
--- a/src/tiledb/cloud/bioimg/helpers.py
+++ b/src/tiledb/cloud/bioimg/helpers.py
@@ -41,10 +41,18 @@ def is_folder(path: str) -> bool:
     return path.endswith("/")
 
 
-def validate_io_paths(source: Sequence[str], output: Sequence[str]) -> None:
+def validate_io_paths(
+    source: Sequence[str], output: Sequence[str], register: bool
+) -> None:
     if len(source) == 0 or len(output) == 0:
         raise ValueError("Source/Output list must not be empty.")
 
+    if register:
+        if any(o.startswith("tiledb://") for o in output):
+            raise ValueError(
+                "Invalid value of argument 'register'. \
+                             TileDB URIs require register argument to be set to False"
+            )
     if len(source) == 1 and len(output) == 1:
         if is_folder(source[0]) and not is_folder(output[0]):
             raise ValueError("Invalid combination of source and output paths.")
--- a/src/tiledb/cloud/bioimg/helpers.py
+++ b/src/tiledb/cloud/bioimg/helpers.py
@@ -50,8 +50,7 @@ def validate_io_paths(
     if for_registration:
         if any(o.startswith("tiledb://") for o in output):
             raise ValueError(
-                "Invalid value of argument 'register'"
-                "TileDB URIs require register argument to be set to False"
+                "Output sequence contains a tiledb URI and this cannot be re-registered."
             )
     if len(source) == 1 and len(output) == 1:
         if is_folder(source[0]) and not is_folder(output[0]):
--- a/src/tiledb/cloud/bioimg/helpers.py
+++ b/src/tiledb/cloud/bioimg/helpers.py
@@ -42,16 +42,16 @@ def is_folder(path: str) -> bool:
 
 
 def validate_io_paths(
-    source: Sequence[str], output: Sequence[str], register: bool
+    source: Sequence[str], output: Sequence[str], *, for_registration: bool
 ) -> None:
     if len(source) == 0 or len(output) == 0:
         raise ValueError("Source/Output list must not be empty.")
 
-    if register:
+    if for_registration:
         if any(o.startswith("tiledb://") for o in output):
             raise ValueError(
-                "Invalid value of argument 'register'. \
-                             TileDB URIs require register argument to be set to False"
+                "Invalid value of argument 'register'"
+                "TileDB URIs require register argument to be set to False"
             )
     if len(source) == 1 and len(output) == 1:
         if is_folder(source[0]) and not is_folder(output[0]):
--- a/src/tiledb/cloud/bioimg/helpers.py
+++ b/src/tiledb/cloud/bioimg/helpers.py
@@ -50,7 +50,8 @@ def validate_io_paths(
     if for_registration:
         if any(o.startswith("tiledb://") for o in output):
             raise ValueError(
-                "Output sequence contains a tiledb URI and this cannot be re-registered."
+                "Output sequence contains a tiledb URI"
+                "and this cannot be re-registered."
             )
     if len(source) == 1 and len(output) == 1:
         if is_folder(source[0]) and not is_folder(output[0]):
--- a/src/tiledb/cloud/bioimg/ingestion.py
+++ b/src/tiledb/cloud/bioimg/ingestion.py
@@ -292,7 +292,7 @@ def ingest(
         )
     source = [source] if isinstance(source, str) else source
     output = [output] if isinstance(output, str) else output
-    validate_io_paths(source, output, register)
+    validate_io_paths(source, output, for_registration=register)
 
     # Build the task graph
     dag_name = taskgraph_name or DEFAULT_DAG_NAME
--- a/src/tiledb/cloud/bioimg/ingestion.py
+++ b/src/tiledb/cloud/bioimg/ingestion.py
@@ -36,6 +36,7 @@ def ingest(
     exclude_metadata: bool = False,
     converter: Optional[str] = None,
     output_ext: str = "",
+    tile_scale: int = 1,
     **kwargs,
 ) -> tiledb.cloud.dag.DAG:
     """The function ingests microscopy images into TileDB arrays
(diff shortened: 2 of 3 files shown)


src/tiledb/cloud/bioimg/helpers.py
```diff
@@ -41,10 +41,18 @@ def is_folder(path: str) -> bool:
...
+) -> None:
     if len(source) == 0 or len(output) == 0:
         raise ValueError("Source/Output list must not be empty.")
 
+    if register:
+        if any(o.startswith("tiledb://") for o in output):
+            raise ValueError(
+                "Invalid value of argument 'register'. \
+                             TileDB URIs require register argument to be set to False"
+            )
```
sgillies, line 55: @ktsitsi do you think we could design the ingestion so that `tiledb://` URIs and `register=True` condition can't occur?

tests/test_bioimg.py
```diff
@@ -19,9 +19,11 @@ def test_validate_io_paths_accepted(self):
                 ["s3://test_out/b", "s3://test_out/d"],
             ),
         }
-        for test_name, (source, dest) in accepted_pairs.items():
-            with self.subTest(f"case: {test_name}"):
-                validate_io_paths(source, dest)
+        register = {"true": True, "false": False}
+        for _, r in register.items():
+            for test_name, (source, dest) in accepted_pairs.items():
```
sgillies, line 24: @ktsitsi what would you think about moving these loops to **outside** the test function using `pytest.parameterize()`? It's more powerful and more clear than `subTest`.

sgillies (commented):
@ktsitsi the `register` and `tiledb://` conditions feels to me like something that we could avoid. But I could be wrong!

Also I've got a suggestion about making the tests easier to read, but that's not a blocker. Using `subTest()` is fine.

src/tiledb/cloud/bioimg/helpers.py
ktsitsi: We offer here alternative paths to registration. When a user provides tiledb uris as outputs, then the registration is being performed during ingestion, as we pre-register the assets before writing to them. In this case we do not need the extra registration stage  inside the DAG. Even though these two could co-exist, in this scenario we would be adding extra registation nodes to the graph that would complete with `Dataset already registered`. The registration stage will not fail but will be useless so thus I am deactivating when tiledb uris are present and we inform the user.

tests/test_bioimg.py
ktsitsi: Indeed @sgillies and we are using `pytest.parametrize` in other repos. The tests for bioimg in this repo need to be redesigned in any case and there is a ticket for that. I followed this path so the tests match the test style of other files and subtest was a suggestion in a previous PR comment. But I am positive in changing them in a separate PR.

thetorpedodog: Agreed on this. The tests in this package are written in classic `unittest` style mostly as an artifact of history and we could generally get a lot of mileage out of moving things to use more pytest infrastructure.

tests/test_bioimg.py
```diff
@@ -19,9 +19,11 @@ def test_validate_io_paths_accepted(self):
                 ["s3://test_out/b", "s3://test_out/d"],
             ),
         }
-        for test_name, (source, dest) in accepted_pairs.items():
-            with self.subTest(f"case: {test_name}"):
-                validate_io_paths(source, dest)
+        register = {"true": True, "false": False}
+        for _, r in register.items():
```
thetorpedodog, line 23: `for r in (True, False)`?

src/tiledb/cloud/bioimg/helpers.py
```diff
@@ -41,10 +41,18 @@ def is_folder(path: str) -> bool:
     return path.endswith("/")
 
 
-def validate_io_paths(source: Sequence[str], output: Sequence[str]) -> None:
+def validate_io_paths(
+    source: Sequence[str], output: Sequence[str], register: bool
```
thetorpedodog, line 45: When there’s a boolean parameter, I find it almost always better to make it keyword-only, because in general there’s no way at the callsite to tell what `do_something(x, y, True)` means vs. `do_something(x, y, False)`. `for_registration` might be a better name for this parameter as well.

tests/test_bioimg.py
ktsitsi: Yeap just wanted to make this more readable for the one that reads the tests.

thetorpedodog: Given that, in this case, the "name" is literally just "true" or "false", I'm not sure this does that; I find it just adds a layer of indirection. Maybe making the `r` value be named `register` (or `for_registration` as above) would communicate the intent better?

ktsitsi: No worries will change.

src/tiledb/cloud/bioimg/helpers.py
ktsitsi: Indeed there will be eventually a change of design, where we will require only TileDB URIs and the register stage could be dropped. So you are correct here, but since this is a customer facing API we need to keep it as backwards compatible as possible for the time being and don't affect customer usage, before we communicate changes.

src/tiledb/cloud/bioimg/helpers.py
ktsitsi: Sure I missed this one. I thought that since validate is an internal usage function the caller that passes register as value would make it self explanatory but I will make it a named arg.

src/tiledb/cloud/bioimg/helpers.py
```diff
@@ -42,16 +42,16 @@ def is_folder(path: str) -> bool:
...
         raise ValueError("Source/Output list must not be empty.")
 
-    if register:
+    if for_registration:
         if any(o.startswith("tiledb://") for o in output):
             raise ValueError(
-                "Invalid value of argument 'register'. \
-                             TileDB URIs require register argument to be set to False"
+                "Invalid value of argument 'register'"
+                "TileDB URIs require register argument to be set to False"
```
sgillies, line 54: @ktsitsi since this check is to catch tiledb URIs, let's say that in the exception message.

```suggestion
                "Output sequence contains a tiledb URI and this cannot be re-registered."
```

sgillies (approved):
@ktsitsi I made one last suggestion for clarity in the exception message. That's all for me.

src/tiledb/cloud/bioimg/ingestion.py
```diff
@@ -48,13 +49,14 @@ def ingest(
     :param taskgraph_name: Optional name for taskgraph, defaults to None
     :param num_batches: Number of graph nodes to spawn.
         Performs it sequentially if default, defaults to 1
-    :param threads: Number of threads for node side multiprocessing, defaults to 8
+    :param threads: Number of threads for node side multiprocessing, defaults to 0
```
ihnorton, line 52: Should this be 1?

ihnorton: (also, I wish we had called this `processes` rather than `threads`, but alas - deprecate and replace in the future)

ihnorton (approved):
Minor nit, otherwise lgtm

src/tiledb/cloud/bioimg/ingestion.py
ktsitsi: The `threads` value is given to the `max_workers` argument in `ingestion_udf`. The reason for this rename is that DAG API has its own `max_workers` argument and I had to separate them without breaking the API. Successively this is passed to https://github.com/TileDB-Inc/TileDB-BioImaging/blob/f8365af24eb00714f851fef4a5b29993df7283b0/tiledb/bioimg/converters/base.py#L299. The default value there is `0`, thus is why kept it `0` in the upper level. When `0` we handle it as `None` inside the `bioimg.to_tiledb` https://github.com/TileDB-Inc/TileDB-BioImaging/blob/f8365af24eb00714f851fef4a5b29993df7283b0/tiledb/bioimg/converters/base.py#L589 so it will default to the number of processors on the machine. I called it `threads` cause we are using a `ThreadpoolExecutor`. I am open to renaming though I just tried to be consistent without breaking the API.

sgillies: @ktsitsi Would it be useful to follow https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor as an example of paramererizing the number of workers? There the number of workers parameter defaults to None and then the documentation explains that the default is the number of processors * 5.

In our docstring above "defaults to 0" has the potential to mislead a user.

PR #529, opened 2024-03-06
Invites SDK [WIP]

JohnMoutafis:
Add support for handling of invites in TileDB-Cloud-Py


src/tiledb/cloud/invites.py
```diff
@@ -0,0 +1,201 @@
...
+    Accept an invitation.
+
+    Args:
+        invitation_id (str): the ID of invitation about to be accepted.
+
+    Raises:
+        TileDBCloudError
+
+    Returns:
+        None
```
thetorpedodog, line 20: We use Sphinx-style docstrings in here:

```
Accepts an invitation.

:param invitation_id: The UUID of the invitation to be accepted.
```

The `raises` section is probably not useful in this case.

src/tiledb/cloud/invites.py
```diff
@@ -0,0 +1,201 @@
...
+    """
+    invitation_api = client.build(rest_api.InvitationApi)
+    try:
+        return invitation_api.join_organization(organization, email_invite)
```
thetorpedodog, line 78: I don’t think we want people to have to be constructing their own internal structures to call this API. Instead, I recommend taking in a sequence of invitees (email addresses or usernames) and the role to provide them. I also don’t feel like there is any obvious order that the parameters should go in (you could easily think that either who is being invited, or the org to be invited to should go first), and because of that it might be best to make them kwargs-only, so that users can’t accidentally mess it up.

This might look like:

```python
def invite_to_organization(
    *,
    organization: str,
    recipients: Sequence[str],
    role: str,
) -> None:
    ...
```

src/tiledb/cloud/invites.py
```diff
@@ -0,0 +1,201 @@
...
+    """
+    invitation_api = client.build(rest_api.InvitationApi)
+    try:
+        return invitation_api.cancel_join_organization(invitation_id, organization)
+    except rest_api.ApiException as exc:
+        raise tiledb_cloud_error.check_exc(exc)
+
+
+def invite_to_array(
+    namespace: str, array: str, email_invite: InvitationOrganizationJoinEmail
```
thetorpedodog, line 105: Most of our functions accept the URI of the array and then use `utils.split_uri` to split it into namespace and array, so a user could call:

```python
invite_to_array("tiledb://some_ns/some_arr", recipients=["..."], roles=["..."])
```

thetorpedodog (commented):
I have quite a few comments here, but they’re more about getting used to the style and conventions of the library than anything else. This is a good start!

src/tiledb/cloud/invites.py
```diff
@@ -0,0 +1,165 @@
...
+    InvitationOrganizationJoinEmail,
+)
+from tiledb.cloud.rest_api.models.organization_roles import OrganizationRoles
+
+
+def accept_invitation(invitation_id: str) -> None:
+    """
+    Accept an invitation.
+
+    :param str invitation_id: the ID of invitation about to be accepted.
```
thetorpedodog, line 17: Since we have type annotations in the signature, there is no need to include them in the params, so this can just be `:param invitation_id:`

src/tiledb/cloud/invites.py
```diff
@@ -0,0 +1,165 @@
...
+    Fetches a paginated list of invitations.
+
+    :param str organization: name or ID of organization to filter
+    :param str array: name/uri of array that is url-encoded to filter
+    :param str group: name or ID of group to filter
+    :param int start: start time for tasks to filter by
+    :param int end: end time for tasks to filter by
+    :param int page: pagination offset
+    :param int per_page: pagination limit
+    :param str type: invitation type, \"ARRAY_SHARE\", \"JOIN_ORGANIZATION\"
```
thetorpedodog, line 38: no need to escape these in a triple-quoted string

src/tiledb/cloud/invites.py
```diff
@@ -0,0 +1,165 @@
...
+    """
+    invitation_api = client.build(rest_api.InvitationApi)
+    try:
+        return invitation_api.fetch_invitations(**filters)
+    except rest_api.ApiException as exc:
+        raise tiledb_cloud_error.check_exc(exc)
+
+
+def invite_to_organization(
+    organization: str, *, recipients: List[str], role: OrganizationRoles
```
thetorpedodog, line 52: `OrganizationRoles` unfortunately isn’t an actual type that you can instantiate (like an `enum.Enum`), it’s more just a container for strings, so it should be annotated as `str`. Also, `recipients` could be more or less any `Sequence[str]`, not just a `List`.

src/tiledb/cloud/invites.py
```diff
@@ -0,0 +1,165 @@
...
+) -> None:
+    """
+    Sends email to multiple recipients with joining information
+    regarding an organization.
+
+    :param str organization: name or UUID of organization.
+    :param List[str] recipients: list of recipient emails
```
thetorpedodog, line 59: each recipient can be either an email address or a TileDB Cloud username.

src/tiledb/cloud/invites.py
```diff
@@ -0,0 +1,165 @@
...
+    :param str array: name/uri of array that is url-encoded
+    :param List[str] recipients: list of recipient emails
+    :param OrganizationRoles role: role assigned to the recipient.
+    :return: None
+    """
+    invitation_api = client.build(rest_api.InvitationApi)
+    namespace, array = utils.split_uri(array)
+    email_invite = InvitationOrganizationJoinEmail(
+        organization_role=role, invitee_email=recipients
+    )
```
thetorpedodog, line 103: I think you got a bit too excited with copy/paste here…

src/tiledb/cloud/invites.py
```diff
@@ -0,0 +1,165 @@
...
+    """
+    invitation_api = client.build(rest_api.InvitationApi)
+    try:
+        return invitation_api.cancel_join_organization(invitation_id, organization)
+    except rest_api.ApiException as exc:
+        raise tiledb_cloud_error.check_exc(exc)
+
+
+def invite_to_array(
+    array: str, *, recipients: List[str], role: OrganizationRoles
```
thetorpedodog, line 89: to match existing APIs, I would recommend calling the array `uri`.

src/tiledb/cloud/invites.py
```diff
@@ -0,0 +1,165 @@
...
+) -> None:
+    """
+    Share array by email invite.
+
+    :param str array: name/uri of array that is url-encoded
```
thetorpedodog, line 94: and likewise, this might be something like

    URI of the array, in the form ``tiledb://namespace/...``

src/tiledb/cloud/invites.py
JohnMoutafis: Not really, although `InvitationOrganizationJoinEmail`'s `__init__` seems to suggest that `organizations_role` is optional, I am getting the following error when omitting it: `ValueError: Invalid value for 'organization_role', must not be 'None'`

thetorpedodog: (this function is for inviting to an array)

JohnMoutafis: yeap, I knew I should be more caffeinated!!

PR #532, opened 2024-03-08
Add indexing function for indexing files to vector search

NikolaosPapailiou:
This adds one-click ingestion function for ingesting files to vector search. 

Tested in cloud:
- [Driver](https://cloud.tiledb.com/compute/logs/taskgraphs/TileDB-Inc/9aa94bfe-f403-49fa-bea3-6949110f4486)
- [ingestion](https://cloud.tiledb.com/compute/logs/taskgraphs/TileDB-Inc/94d38f5d-d293-4882-8067-12bfdef58bda)


src/tiledb/cloud/vector_search/file_ingestion.py
```diff
@@ -0,0 +1,326 @@
...
+        # TODO(nikos) remove the pacakge installations after the image is updated
+        def install_extra_worker_modules():
+            import os
+            import subprocess
+            import sys
+
+            sys.path.insert(0, "/home/udf/.local/bin")
+            sys.path.insert(0, "/home/udf/.local/lib/python3.9/site-packages")
+            os.environ["PATH"] = f"/home/udf/.local/bin:{os.environ['PATH']}"
+            subprocess.check_call(
```
Shelnutt2, line 199: This needs to be removed. Add packages required to default images.

src/tiledb/cloud/vector_search/file_ingestion.py
```diff
@@ -0,0 +1,326 @@
...
+    glob: str = "**/[!.]*",
+    exclude: Sequence[str] = (),
+    suffixes: Optional[Sequence[str]] = None,
+    text_splitter: str = "RecursiveCharacterTextSplitter",
+    text_splitter_kwargs: Optional[Dict] = {
+        "chunk_size": 500,
+        "chunk_overlap": 50,
+    },
+    # SentenceTransformersEmbedding params
+    model_name_or_path: str = "BAAI/bge-small-en-v1.5",
```
Shelnutt2, line 30: We had discussed using openai as the default model and having openai api token passed in. What has prompted this change?

src/tiledb/cloud/vector_search/file_ingestion.py
```diff
@@ -0,0 +1,326 @@
...
+        reader = DirectoryTextReader(
+            uri=file_dir_uri,
+            glob=f"**/{file_name}" if file_name is not None else glob,
+            exclude=exclude,
+            suffixes=suffixes,
+            text_splitter=text_splitter,
+            text_splitter_kwargs=text_splitter_kwargs,
+        )
+
+        embedding = SentenceTransformersEmbedding(
```
Shelnutt2, line 241: Before this the huggingface location needs to be set so any model used is loaded into `/dev/shm` not the home directory.

src/tiledb/cloud/vector_search/file_ingestion.py
```diff
@@ -0,0 +1,326 @@
...
+            model_name_or_path=model_name_or_path,
+        )
+        if create_index:
```
Shelnutt2, line 244: We should detect if the index_uri exists or not so the UX is cleaner for the user and the UI,

src/tiledb/cloud/vector_search/file_ingestion.py
```diff
@@ -0,0 +1,326 @@
...
+            namespace=namespace,
+            verbose=verbose,
+            trace_id=trace_id,
+            **index_update_kwargs,
+        )
+
+    graph = dag.DAG(
+        name="file-vector-search-ingestion",
+        mode=dag.Mode.BATCH,
+        max_workers=1,
```
Shelnutt2, line 282: Why are we launching a batch task graph of 1 worker? Why do we not use a single dag for the whole process?

src/tiledb/cloud/vector_search/file_ingestion.py
```diff
@@ -0,0 +1,326 @@
+from typing import Dict, List, Optional, Sequence
+
+from tiledb.cloud import dag
+from tiledb.cloud.utilities import as_batch
+
+
+def ingest_files_dag(
```
Shelnutt2, line 7: Need parameters for passing in OpenAI API tokens and using OpenAI embeddings.

src/tiledb/cloud/vector_search/file_ingestion.py
```diff
@@ -0,0 +1,326 @@
...
+    :param config: config dictionary, defaults to None.
+    :param namespace: TileDB-Cloud namespace, defaults to None.
+    :param verbose: verbose logging, defaults to False.
+    :param trace_id: trace ID for logging, defaults to None.
+    # Index creation params
+    :param create_index: If true, creates a new vector search index.
+    :param index_type: Vector search index type ("FLAT", "IVF_FLAT").
+    :param index_creation_kwargs: Arguments to be passed to the index creation method
+    # DirectoryTextReader params.
+    :param glob: Glob pattern relative to the specified path by default set to pick up
```
Shelnutt2, line 64: Lets standardize the parameters, we use `include` and `exclude` in vcf, imaging and geospatial ingestion functions.

Shelnutt2 (changes requested):
Several comments in first pass.

This is also completely missing interacting with TileDB Files. The requirement is to support loading into TileDB Files store and performing indexing. Not just indexing.

src/tiledb/cloud/vector_search/file_ingestion.py
NikolaosPapailiou: OpenAi embeddings are nor supported yet in the object api ingestion. If we want OpenAI embeddings we will need to add an implementation and also figure out how to pass the credentials to the taskgraph.

src/tiledb/cloud/vector_search/file_ingestion.py
NikolaosPapailiou: This is just a temporary solution to test before the release of the new UDF image. If we remove this the PR will need to block on the UDF image release.

Shelnutt2: We cannot ship as is, this is a blocking change request. Please add the packages needed to the default UDF images.

src/tiledb/cloud/vector_search/file_ingestion.py
Shelnutt2: The original plan that was discussed was to pass the OpenAI API token as a parameter to start with. Do you feel that is longer viable?

NikolaosPapailiou: In this case, users need to pass it every time the index a file in plain text. We don't have a way to store and retrieve this. 

Apart from the key we need to implement the embedding function that would execute the API calls

src/tiledb/cloud/vector_search/file_ingestion.py
NikolaosPapailiou: We need this because `as_batch` doesn't allow us to specify the image that is used on the first taskgraph. In our case, we need to start with `vectorsearch` image which contains the packages required by `ingest_files_udf`

src/tiledb/cloud/vector_search/file_ingestion.py
Shelnutt2: This sounds like what we had discussed in planning two weeks ago. I suggest we move forward with supporting openai.

src/tiledb/cloud/vector_search/file_ingestion.py
NikolaosPapailiou: I have tried this with larger models and I think it worked by default using the container memory. Will check again. If needed we can also set the `cache_folder` param of `SentenceTransformersEmbedding`

NikolaosPapailiou:
> This is also completely missing interacting with TileDB Files. The requirement is to support loading into TileDB Files store and performing indexing. Not just indexing.

I don't think we have discussed requirements for this. This needs to be designed in collaboration with cloud and we need to understand who owns the file ingestion code and implementation.

src/tiledb/cloud/vector_search/file_ingestion.py
Shelnutt2: Yes the UI will list our existing ones or let a user select a new one. That doesn't mean the code has to require the user to toggle it themselves. In other ingestors like VCF we have a simple check to see if the uri already exists or not to make it a better UX instead of having the user perform that check.

src/tiledb/cloud/vector_search/file_ingestion.py
NikolaosPapailiou: We don't have an OpenAI embeddings implementation yet for the object api.

src/tiledb/cloud/vector_search/file_ingestion.py
NikolaosPapailiou: This will require some changes in the vector search library. Is this ok if I address it in a separate PR?

src/tiledb/cloud/vector_search/file_ingestion.py
NikolaosPapailiou: Removed

src/tiledb/cloud/vector_search/file_ingestion.py
NikolaosPapailiou: Added support for OpenAI embeddings

src/tiledb/cloud/vector_search/file_ingestion.py
NikolaosPapailiou: Added support for OpenAI embeddings

src/tiledb/cloud/vector_search/file_ingestion.py
NikolaosPapailiou: Consolidated the arguments. This is blocking on the release of https://github.com/TileDB-Inc/TileDB-Vector-Search/pull/283

Shelnutt2:
> I don't think we have discussed requirements for this. This needs to be designed in collaboration with cloud and we need to understand who owns the file ingestion code and implementation.

When you are back we need to sync on this. I believe we had discussed this and the example POC code I provided handled all of this. The goal was to take the POC we used and re-implement the features in a production fashion. The first goal is to support file ingestion and index creation as part of one pipeline.

src/tiledb/cloud/vector_search/file_ingestion.py
```diff
@@ -0,0 +1,354 @@
...
+        :param vector_indexing_mode: TaskGraph execution mode for the vector indexing.
+        :param index_update_kwargs: Extra arguments to pass to the index update job.
+        """
+        import importlib
+
+        import tiledb
+        from tiledb.vector_search.object_api import object_index
+        from tiledb.vector_search.object_readers import DirectoryTextReader
+
+        def index_exists(
```
JohnMoutafis, line 220: From an engineering perspective, can we avoid nesting methods like this that is also only used once?
It adds unnecessary overhead and potentially makes debugging more convoluted.

src/tiledb/cloud/vector_search/file_ingestion.py
```diff
@@ -0,0 +1,354 @@
...
+from tiledb.cloud.utilities import as_batch
+
+
+def ingest_files_dag(
+    file_dir_uri: str,
+    index_uri: str,
+    file_name: Optional[str] = None,
+    acn: Optional[str] = None,
+    config=None,
+    environment_variables: Mapping[str, str] = {},
```
thetorpedodog, line 13: Including mutable objects as default parameters will have unexpected results, since the *reference* is used as the default parameter:

```python
def do_something(it: dict = {}):
  return it, len(it)

a_dict, size = do_something()
a_dict["my value"] = 1

dict_two, size_two = do_something()
# dict_two == {"my value": 1}; size_two == 1
# a_dict is dict_two
```

Instead, I would recommend using an Optional:

```python
def do_something(it: Optional[dict] = None):
  it = {} if it is none else it
  return it, len(it)

a_dict, size = do_something()
a_dict["my_value"] = 1

dict_two, size_two = do_something()
# dict_two == {}; size_two == 0
# a_dict is not dict_two
```

src/tiledb/cloud/vector_search/file_ingestion.py
```diff
@@ -0,0 +1,354 @@
...
+        embeddings_generation_mode=embeddings_generation_mode,
+        embeddings_generation_driver_mode=embeddings_generation_driver_mode,
+        vector_indexing_mode=vector_indexing_mode,
+        index_update_kwargs=index_update_kwargs,
+    )
+    graph.compute()
+    graph.wait()
+
+
+ingest_files = as_batch(ingest_files_dag)
```
JohnMoutafis, line 354: This could be named `ingest` to follow the naming convention of the VCF implementation as a guide.

src/tiledb/cloud/vector_search/file_ingestion.py
```diff
@@ -0,0 +1,354 @@
...
+    namespace: Optional[str] = None,
+    verbose: bool = False,
+    trace_id: Optional[str] = None,
+    # Index creation params
+    create_index: bool = False,
+    index_type: str = "IVF_FLAT",
+    index_creation_kwargs: Dict = {},
+    # DirectoryTextReader params
+    include: str = "*",
+    exclude: Sequence[str] = ["[.]*", "*/[.]*"],
```
thetorpedodog, line 23: This is simpler; you can replace the list with a tuple; i.e., `("[.]*", "*/[.]*")`.

src/tiledb/cloud/vector_search/file_ingestion.py
```diff
@@ -0,0 +1,354 @@
...
+    :param max_tasks_per_stage: Number of maximum udf tasks per computation stage.
+    :param embeddings_generation_mode: TaskGraph execution mode for embeddings
+        generation.
+    :param embeddings_generation_driver_mode: TaskGraph execution mode for the ingestion
+        driver.
+    :param vector_indexing_mode: TaskGraph execution mode for the vector indexing.
+    :param index_update_kwargs: Extra arguments to pass to the index update job.
+    """
+
+    def ingest_files_udf(
```
thetorpedodog, line 110: Since this is a purely-internal function and the parent function is guaranteed to pass in all values, the default values are not needed here.

JohnMoutafis (changes requested):
We should avoid nesting functions especially when they are only used once through out the body of the method.
Also functions like `index_exists` that only work as a "passthrough" to only call another function should be avoided as they are unnecessary, they add overhead and they complicate the code and any potential debugging process.

src/tiledb/cloud/vector_search/file_ingestion.py
thetorpedodog: This seems like a parameter that could be added to `as_batch`.

src/tiledb/cloud/vector_search/file_ingestion.py
NikolaosPapailiou: Applied

src/tiledb/cloud/vector_search/file_ingestion.py
NikolaosPapailiou: Applied

src/tiledb/cloud/vector_search/file_ingestion.py
NikolaosPapailiou: Removed

src/tiledb/cloud/vector_search/file_ingestion.py
NikolaosPapailiou: Simplified

Changed as_batch to pass the `image_name` argument and now we don't require all the nesting of the functions.

src/tiledb/cloud/vector_search/file_ingestion.py
NikolaosPapailiou: Added `image_name` as argument to `as_batch`

src/tiledb/cloud/vector_search/file_ingestion.py
NikolaosPapailiou: Applied

src/tiledb/cloud/vector_search/file_ingestion.py
```diff
@@ -0,0 +1,326 @@
...
+        :param embeddings_generation_driver_mode: TaskGraph execution mode for the
+            ingestion driver.
+        :param vector_indexing_mode: TaskGraph execution mode for the vector indexing.
+        :param index_update_kwargs: Extra arguments to pass to the index update job.
+        """
+        import tiledb
+
+        # Install extra packages not yet available in vectorsearch UDF image.
+        # TODO(nikos) remove the pacakge installations after the image is updated
+        def install_extra_worker_modules():
```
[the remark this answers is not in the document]
NikolaosPapailiou: Removed

JohnMoutafis (approved):

src/tiledb/cloud/vector_search/file_ingestion.py
```diff
@@ -0,0 +1,201 @@
...
+    embeddings_module = importlib.import_module("tiledb.vector_search.embeddings")
+    embedding_class_ = getattr(embeddings_module, embedding_class)
+    embedding = embedding_class_(**embedding_kwargs)
+
+    with tiledb.scope_ctx(config):
+        index_uri_exists = tiledb.object_type(index_uri) == "group"
+    if create_index:
+        if index_uri_exists:
+            raise ValueError(f"{index_uri} allready exists and `create_index` was set.")
+        else:
```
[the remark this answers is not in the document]
NikolaosPapailiou: Applied

src/tiledb/cloud/vector_search/file_ingestion.py
```diff
@@ -0,0 +1,201 @@
...
+        exclude=exclude,
+        suffixes=suffixes,
+        max_files=max_files,
+        text_splitter=text_splitter,
+        text_splitter_kwargs=text_splitter_kwargs,
+    )
+
+    if openai_key is not None:
+        environment_variables["OPENAI_API_KEY"] = openai_key
+    embeddings_module = importlib.import_module("tiledb.vector_search.embeddings")
```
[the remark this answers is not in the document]
NikolaosPapailiou: Changed

src/tiledb/cloud/vector_search/file_ingestion.py
```diff
@@ -0,0 +1,201 @@
...
+    if index_update_kwargs is None:
+        index_update_kwargs = {}
+    if embedding_kwargs is None:
+        embedding_kwargs = {}
+    if text_splitter_kwargs is None:
+        text_splitter_kwargs = {}
+
+    reader = DirectoryTextReader(
+        search_uri=file_dir_uri,
+        include=f"{file_name}" if file_name is not None else include,
```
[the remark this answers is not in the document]
NikolaosPapailiou: Done

src/tiledb/cloud/vector_search/file_ingestion.py
```diff
@@ -0,0 +1,201 @@
...
+    # DirectoryTextReader params
+    include: str = "*",
+    exclude: Optional[Sequence[str]] = ("[.]*", "*/[.]*"),
+    suffixes: Optional[Sequence[str]] = None,
+    max_files: Optional[int] = None,
+    text_splitter: str = "RecursiveCharacterTextSplitter",
+    text_splitter_kwargs: Optional[Dict] = {
+        "chunk_size": 500,
+        "chunk_overlap": 50,
+    },
```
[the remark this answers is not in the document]
NikolaosPapailiou: What do you mean by pulled out? I don't see any other comment about `embedding_kwargs ` and `driver_resources `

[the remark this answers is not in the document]
thetorpedodog: These are also mutable default values, so their setting needs to be moved into the function itself and the default here replaced with `None`.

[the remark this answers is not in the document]
NikolaosPapailiou: Applied

thetorpedodog (approved):

NikolaosPapailiou:
@Shelnutt2 are your comments addressed by the changes? This requires your approval to continue with merging.

src/tiledb/cloud/vector_search/file_ingestion.py
```diff
@@ -0,0 +1,200 @@
...
+        text_splitter=text_splitter,
+        text_splitter_kwargs=text_splitter_kwargs,
+    )
+
+    if openai_key is not None:
+        environment_variables["OPENAI_API_KEY"] = openai_key
+    embedding_class_ = getattr(embeddings_module, embedding_class)
+    embedding = embedding_class_(**embedding_kwargs)
+
+    with tiledb.scope_ctx(config):
```
Shelnutt2, line 144: This should be done as the first stage in the graph, not local to the caller.

NikolaosPapailiou: I am not sure I understand this, `ingest_files_udf` is running within the taskgraph

src/tiledb/cloud/vector_search/__init__.py
```diff
@@ -0,0 +1,5 @@
+from .file_ingestion import ingest_files
+
+__all__ = [
+    "ingest_files",
```
Shelnutt2, line 4: This doesn't appear to exist?

NikolaosPapailiou: Fixed

src/tiledb/cloud/vector_search/file_ingestion.py
```diff
@@ -0,0 +1,200 @@
...
+        index_uri_exists = tiledb.object_type(index_uri) == "group"
+    if create_index:
+        if index_uri_exists:
+            raise ValueError(f"{index_uri} allready exists and `create_index` was set.")
```
Shelnutt2, line 148: `already` has a typo.

src/tiledb/cloud/vector_search/file_ingestion.py
```diff
@@ -0,0 +1,200 @@
...
+    index_uri: str,
+    file_name: Optional[str] = None,
+    acn: Optional[str] = None,
+    config=None,
+    environment_variables: Optional[Mapping[str, str]] = None,
+    namespace: Optional[str] = None,
+    verbose: bool = False,
+    trace_id: Optional[str] = None,
+    # Index creation params
+    create_index: bool = False,
```
Shelnutt2, line 18: What's the reason for having the user check if the index exists or not and settings this parameter vs detecting it internally?

src/tiledb/cloud/vector_search/file_ingestion.py
NikolaosPapailiou: Fixed

src/tiledb/cloud/vector_search/file_ingestion.py
NikolaosPapailiou: This was just to make sure that we don't accidentally write data in an existing dataset.
Removed

src/tiledb/cloud/vector_search/file_ingestion.py
NikolaosPapailiou: I created an alternative version of the PR that uses an extra taskgraph and creates the dataset as a first node in the taskgraph  https://github.com/TileDB-Inc/TileDB-Cloud-Py/pull/547 Is this what you are expecting the ingestion structure to look like?

NikolaosPapailiou: @Shelnutt2 is the alternative version matching your expectation? Let me know if you still have concerns

NikolaosPapailiou: Applied the alternative taskgraph structure in this PR.

NikolaosPapailiou: @Shelnutt2 this PR needs your approval to move forward. Let me know if you need any more changes.

src/tiledb/cloud/vector_search/file_ingestion.py
```diff
@@ -0,0 +1,355 @@
...
+    index_timestamp: Optional[int] = None,
+    workers: int = -1,
+    worker_resources: Optional[Dict] = None,
+    worker_image: Optional[str] = None,
+    extra_worker_modules: Optional[List[str]] = None,
+    driver_resources: Optional[Dict] = None,
+    driver_image: Optional[str] = None,
+    extra_driver_modules: Optional[List[str]] = None,
+    max_tasks_per_stage: int = -1,
+    embeddings_generation_mode: dag.Mode = dag.Mode.LOCAL,
```
Shelnutt2, line 41: Everything must default to batch mode. Running this in local is unexpected. The goals are that like all other verticals we support and default to batch ingestion capabilities.

src/tiledb/cloud/vector_search/file_ingestion.py
```diff
@@ -0,0 +1,355 @@
+from typing import Dict, List, Mapping, Optional, Sequence
+
+from tiledb.cloud import dag
+from tiledb.cloud.utilities import as_batch
+
+
+def ingest_files(
+    file_dir_uri: str,
```
Shelnutt2, line 8: This does not work as expected. Passing in a TileDB file URI get ignored. Please test this and add unit tests for the relevant cases. Currently this does not cover the required use cases.

src/tiledb/cloud/vector_search/file_ingestion.py
```diff
@@ -0,0 +1,355 @@
...
+    index_uri: str,
+    file_name: Optional[str] = None,
```
Shelnutt2, line 10: This along with include/exclude don't make sense. How is this to work with TileDB files? There is no check if the TileDB file name or any parsing of the TileDB URIs. The goal again as outlined in the requirements is to use this with TileDB files, either standalone or from a group.

src/tiledb/cloud/vector_search/file_ingestion.py
```diff
@@ -0,0 +1,355 @@
...
+    text_splitter: str = "RecursiveCharacterTextSplitter",
+    text_splitter_kwargs: Optional[Dict] = None,
+    # Embedding params
+    embedding_class: str = "LangChainEmbedding",
+    embedding_kwargs: Optional[Dict] = None,
+    openai_key: Optional[str] = None,
+    # Index update params
+    index_timestamp: Optional[int] = None,
+    workers: int = -1,
+    worker_resources: Optional[Dict] = None,
```
Shelnutt2, line 34: This is not plumbing all the different resource parameters, is there a reason?

src/tiledb/cloud/vector_search/file_ingestion.py
```diff
@@ -0,0 +1,355 @@
...
+            embedding = embedding_class_(**embedding_kwargs)
+
+            if tiledb.object_type(index_uri) == "group":
+                logger.info("Existing dataset: %r. Updating reader.", index_uri)
+                index = object_index.ObjectIndex(
+                    uri=index_uri,
+                    environment_variables=environment_variables,
+                    load_embedding=False,
+                    load_metadata_in_memory=False,
+                    memory_budget=1,
```
Shelnutt2, line 207: Why is this set to 1? Please add inline code comments. There should be a decent amount of comments explaining the purpose of values set such as this one. The goal is for others to be able to read the code + comments and understand the code and be able to work on it. If request.

src/tiledb/cloud/vector_search/file_ingestion.py
```diff
@@ -0,0 +1,355 @@
...
+    # DAG
+    # --------------------------------------------------------------------
+
+    graph = dag.DAG(
+        name="vector-search-file-indexing",
+        namespace=namespace,
+        mode=dag.Mode.BATCH,
+    )
+    if worker_resources is None:
+        driver_resources = {"cpu": "2", "memory": "8Gi"}
```
Shelnutt2, line 298: Did you mean worker or driver here?

Shelnutt2 (changes requested):
Several comments that need to be fixed. This code also needs to be aligned with the goals of handling TileDB FileStore files as a primary source.

Additionally there are some pylint errors related to variables that aren't passed through. Please address all lint errors.

src/tiledb/cloud/vector_search/file_ingestion.py
NikolaosPapailiou: Document indexing has multiple execution steps that can be run in different modes:
- `ingest_files` creates a `BATCH` taskgraph that runs all the indexing. This means that all processing happens within a `BATCH` taskgraph with `access_credentials` even if the options here are set to `LOCAL`
- `embeddings_generation`: reads the documents and creates text embeddings. This can spawn its own taskgraph.
- `vector_indexing`: creates a vector index from the produced embeddings. This can spawn its own taskgraph.

The default configuration at the moment is:
- `ingest_files` creates a `BATCH` taskgraph that runs all the indexing.
- `embeddings_generation, vector_indexing` run in `LOCAL` mode within a UDF of the `ingest_files` taskgraph. Both of these tasks can leverage the available parallelism within the single worker. 

This is expected to be a good default execution configuration for cost and latency even for sets of thousands of documents. 

Do you want all the execution steps to be executed in `BATCH` mode by default?

src/tiledb/cloud/vector_search/file_ingestion.py
NikolaosPapailiou: As it is implemented atm this should be the group URI and we pick up the file from the group(applying regexp patterns if provided). What are the cases you are looking for supporting here?

NikolaosPapailiou: We have some testcases for this here https://github.com/TileDB-Inc/TileDB-Vector-Search/blob/main/apis/python/test/test_directory_reader.py

src/tiledb/cloud/vector_search/file_ingestion.py
Shelnutt2: The requirement again, as we discussed and as spelled out in the story is to have a robust batch mode ingestion that can scale to millions of documents. Local mode is a bad default and does not meet our intended goals.

src/tiledb/cloud/vector_search/file_ingestion.py
Shelnutt2: The requirements are to support TileDB FileStore files or a group of files. This has been a hard requirement from day one and is outlined in our planning document.

src/tiledb/cloud/vector_search/file_ingestion.py
NikolaosPapailiou: Removed the `file_name` option. FileStore URIs work directly using the `search_uri` param.

src/tiledb/cloud/vector_search/file_ingestion.py
NikolaosPapailiou: You mean the vector indexing resources? These can be passed in `index_update_kwargs`

src/tiledb/cloud/vector_search/file_ingestion.py
NikolaosPapailiou: Added comment for this, this avoids loading vector data in memory since we don't want to perform queries.

src/tiledb/cloud/vector_search/file_ingestion.py
NikolaosPapailiou: Changed

JohnMoutafis:
@Shelnutt2 what is the *P0*'s definition of done for this PR?
Since File ingestion is now in place and working, we would like to push this forward and combine it eventually with ingestion.
@NikolaosPapailiou has tested this, using notebooks.

antalakas:
> @Shelnutt2 what is the _P0_'s definition of done for this PR? Since File ingestion is now in place and working, we would like to push this forward and combine it eventually with ingestion. @NikolaosPapailiou has tested this, using notebooks.

@JohnMoutafis We need to add unit tests and prove things work as expected in the context of the change. Integration style (notebook) tests is good but they will have to integrate in later steps as well.

src/tiledb/cloud/_common/utils.py
```diff
@@ -112,3 +113,10 @@ def release_connection(resp: urllib3.HTTPResponse) -> None:
 
 def datetime_to_msec(t: Union[datetime.datetime, int, None]) -> Optional[int]:
     return int(t.timestamp() * 1000) if isinstance(t, datetime.datetime) else t
+
+
+def sanitize_enum_argument(enum_class: Type[Enum], value: str) -> Type[Enum]:
+    try:
+        return enum_class(value)
+    except ValueError:
+        raise ValueError(f"{value} is not a valid {enum_class.__name__}")
```
thetorpedodog, line 122: This should be either a `raise ValueError(...) from None` or `raise ValueError(...) from ve` so that the error you’re raising here is treated as the “real” error and not a "while error A was being handled, error B happened" error

pyproject.toml
```diff
@@ -41,9 +41,7 @@ requires = ["setuptools>=42", "wheel", "setuptools_scm>=6"]
 
 [tool.pytest.ini_options]
 explicit-only = ["geospatial"]
-markers = [
-  "geospatial: tests that require the geospatial libraries",
-]
+markers = ["geospatial: tests that require the geospatial libraries"]
```
thetorpedodog, line 44: maybe add a marker for these file-related tests and add it to the explicit-only list?

tests/test_file.py
```diff
@@ -263,3 +267,78 @@ def test_add_array_to_group_udf_non_existing_group_raises_value_error(self):
                 config=client.Ctx().config().dict(),
                 verbose=True,
             )
+
+
+class TestFileIndexing(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls) -> None:
+        """Setup test files, group and destinations once before the file tests start."""
+        cls.input_file_location = "s3://tiledb-unittest/groups/file_indexing_test_files"
```
thetorpedodog, line 276: To avoid having a hard dependency on files that it’s impossible for users who aren’t us to access, would it make sense to do something like:

 1. Put these files in the testdata folder
 2. Copy them to a new directory in cloud storage when the test starts
 3. Add a way to override this path (environment variable?)

?

This would also let us do something like putting all the data/results in a path that looks like

<code><var>cloud storage base</var>/run-<var>timestamp</var>-<var>random suffix</var>/...</code>

to make automatic cleanup easier if the test crashes or is interrupted in the middle and can't clean itself up.

src/tiledb/cloud/file/indexing.py
```diff
@@ -0,0 +1,405 @@
...
+
+    logger = get_logger_wrapper(verbose)
+    logger.debug("tiledb-vector-search=%s", vs.__version__)
+
+    # Check if the dataset already exists
+    with tiledb.scope_ctx(config):
+        if environment_variables is None:
+            environment_variables = {}
+        if index_creation_kwargs is None:
+            index_creation_kwargs = {}
```
thetorpedodog, line 56: `environment_variables = environment_variables or {}` here. (this construction won’t work for the others below but will make these shorter)

src/tiledb/cloud/file/indexing.py
```diff
@@ -0,0 +1,405 @@
...
+    """
+
+    from tiledb.cloud.utilities import run_dag
+
+    # Argument preparation
+    DEFAULT_IMG_NAME = "vectorsearch"
+    worker_resources = worker_resources or {"cpu": "2", "memory": "8Gi"}
+    driver_resources = driver_resources or {"cpu": "1", "memory": "4Gi"}
+    ## Vector Search BATCH Embedding Resources:
+    index_update_kwargs.update(
```
thetorpedodog, line 319: the named args form of this might make it easier to read:

```python
index_update_kwargs.update(
    ingest_resources=ingest_resources or ...,
    consolidate_partition_resources=...,
    ...,
)
```

thetorpedodog (commented):
Overall looking pretty good. A few suggestions to go

tests/test_file.py
JohnMoutafis: I have tried that for the ingestion tests, but the CI runner has no VFS access so we have abandoned that idea for now.

Shelnutt2: > We have tried that for the ingestion tests, but the CI runner has no VFS access so we have abandoned that idea for now.

We could enable this. Did we consider adding those permissions? I agree with @thetorpedodog here that making this reproducible would go a long way to future support-ability. It doesn't have to block this PR but we should consider an immediate follow up. `s3://tiledb-unittest` as a bucket also need followups, @antalakas lets discuss offline.

JohnMoutafis: Sure I am onboard for that. Will also make the ingestion&indexing testing less flaky.
I will add a followup for that in the ticket.

thetorpedodog (approved):

PR #533, opened 2024-03-08
Adding supported extension for ndpi files

ktsitsi:
This PR:

- Extends ingestion support for `.ndpi` images via cloud ingestor.

**Important**. This PR heavily **depends** on the deployment of a new TileDB-Py version, which will include https://github.com/TileDB-Inc/TileDB-Py/pull/1915

### DIFF
--- a/src/tiledb/cloud/bioimg/ingestion.py
+++ b/src/tiledb/cloud/bioimg/ingestion.py
@@ -14,7 +14,7 @@ from tiledb.cloud.utilities._common import run_dag
 DEFAULT_RESOURCES = {"cpu": "8", "memory": "4Gi"}
 DEFAULT_IMG_NAME = "3.9-imaging-dev"
 DEFAULT_DAG_NAME = "bioimg-ingestion"
-_SUPPORTED_EXTENSIONS = (".tiff", ".tif", ".svs")
+_SUPPORTED_EXTENSIONS = (".tiff", ".tif", ".svs", ".ndpi")
 _SUPPORTED_CONVERTERS = ("tiff", "zarr", "osd")
 
 


Shelnutt2 (approved):

src/tiledb/cloud/bioimg/ingestion.py
```diff
@@ -14,7 +14,7 @@
 DEFAULT_RESOURCES = {"cpu": "8", "memory": "4Gi"}
 DEFAULT_IMG_NAME = "3.9-imaging-dev"
 DEFAULT_DAG_NAME = "bioimg-ingestion"
-_SUPPORTED_EXTENSIONS = (".tiff", ".tif", ".svs")
+_SUPPORTED_EXTENSIONS = (".tiff", ".tif", ".svs", ".ndpi")
```
ihnorton, line 17: Possible follow-up item: we should list the supported extensions in tiledb-bioimg so we maintain in only one place.

ihnorton (approved):

src/tiledb/cloud/bioimg/ingestion.py
ktsitsi: The idea behind listing them here is that we pass them in the "Path Listing UDF", this way we fail in the first node of ingestion. If we allow the lister to parse all extension and move the validation for the supported file extensions to tiledb-bioimg this will lead the `Ingestion` stage to fail with `unsupported type`. I am ok with that if we agree I can create a ticket for this one.

ihnorton: No change for now, just keep in mind for the future - eventually we'll need "is this supported" and "give me supported extension" functions in -bioimg.

JohnMoutafis (approved):

thetorpedodog:
Do you want this to be merged now, or should we wait until the new TileDB core/tiledb-py library is released?

ktsitsi:
> Do you want this to be merged now, or should we wait until the new TileDB core/tiledb-py library is released?

The latter.

PR #534, opened 2024-03-12
List groups flat value should be defaulted to True

vlasisPit:
According to https://tiledb.slack.com/archives/CB0NNLS6B/p1709908178253559 , when listing groups default flat value should be True, to ignore the nesting of the groups and searches all of them.


thetorpedodog (approved):

thetorpedodog:
The build failures are not your fault; TileDB core just doesn't do 3.7 anymore.

PR #536, opened 2024-03-13
Update Group Info.

JohnMoutafis:
SDK to allow Group Info updating, using the v2 of the API.

### DIFF
--- a/src/tiledb/cloud/groups.py
+++ b/src/tiledb/cloud/groups.py
@@ -10,9 +10,7 @@ from tiledb.cloud import client
 from tiledb.cloud import rest_api
 from tiledb.cloud._common import api_v2
 from tiledb.cloud._common import utils
-from tiledb.cloud._common.api_v2.models.group_metadata_update_request import (
-    GroupMetadataUpdateRequest,
-)
+from tiledb.cloud.rest_api.models import group_update
 
 
 def create(
@@ -101,6 +99,7 @@ def update_info(
     *,
     description: Optional[str] = None,
     name: Optional[str] = None,
+    logo: Optional[str] = None,
     tags: Optional[List[str]] = None,
 ) -> None:
     """
@@ -109,12 +108,13 @@ def update_info(
     :param uri: URI of the group in the form 'tiledb://<namespace>/<group>'
     :param description: Group description, defaults to None
     :param name: Group's name, defaults to None
+    :param logo: Group's logo, defaults to None
     :param tags: Group tags, defaults to None
     :return: None
     """
     namespace, group_name = utils.split_uri(uri)
-    groups_v2_client = client.build(api_v2.GroupsApi)
-    metadata = {}
+    groups_v1_client = client.build(rest_api.GroupsApi)
+    info = {}
     for kw, arg in inspect.signature(update_info).parameters.items():
         if arg.kind != inspect.Parameter.KEYWORD_ONLY:
             # Skip every non-keyword-only argument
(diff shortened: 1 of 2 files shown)


src/tiledb/cloud/groups.py
```diff
@@ -92,6 +94,45 @@ def info(uri: str) -> object:
...
+def update_info(
+    uri: str,
+    *,
+    description: Optional[str] = None,
+    name: Optional[str] = None,
+    logo: Optional[str] = None,
+    tags: Optional[List[str]] = None,
+) -> None:
+    """
+    Update Group Metadata
```
snagles, line 106: nit: Maybe "Update Group Attributes", metadata has a very specific meaning in the groups/array context. The term generally refers to the on-disk metadata stored alongside arrays/group

src/tiledb/cloud/groups.py
```diff
@@ -92,6 +94,45 @@ def info(uri: str) -> object:
...
+
+        value = locals()[kw]
+        if value is None:
+            # Explicitly update metadata
+            continue
+        info[kw] = value
+
+    info = group_update.GroupUpdate(**info)
+    try:
+        return groups_v1_client.update_group(namespace, group_name, group_update=info)
```
snagles, line 131: LGTM as long as this points to https://github.com/TileDB-Inc/TileDB-Cloud-API-Spec/blob/master/openapi-v1.yaml#L9615

snagles (approved):

src/tiledb/cloud/groups.py
JohnMoutafis: yy, it points to that endpoint: https://github.com/TileDB-Inc/TileDB-Cloud-Py/blob/92c6cab79f7b461fad0fd65c5cf6d4540a00367a/src/tiledb/cloud/rest_api/api/groups_api.py#L2515

src/tiledb/cloud/groups.py
JohnMoutafis: Updated in the latest commit from "metadata" to "attributes"

tests/test_groups.py
```diff
@@ -42,26 +66,24 @@ def test_create_deregister(self):
...
-        for _ in range(TRIES):
-            try:
-                api.get_group_contents(group_namespace=self.namespace, group_name=name)
-            except rest_api.ApiException as apix:
-                if apix.status in (400, 404):
-                    return
-            time.sleep(1)
-        self.fail(f"group {self.namespace}/{name} still exists")
+        description = "this is a test description"
+        logo = "testLogo"
```
Shelnutt2, line 78: Logo is expected to be a base64 encoded image. If you are looking for a test to validate it end to end perhaps we should hardcode an image value?

tests/test_groups.py
```diff
@@ -42,26 +66,24 @@ def test_create_deregister(self):
...
+        tags = ["tag", "othertag"]
+        groups.update_info(group_uri, description=description, logo=logo, tags=tags)
+        group_info = groups.info(group_uri)
+
+        self.assertEqual(group_info.description, description)
+        self.assertEqual(group_info.logo, logo)
+        self.assertCountEqual(group_info.tags, tags)
+
+        # Cleanup
+        groups.deregister(group_uri)
```
Shelnutt2, line 88: I think we should delete not just deregister so we cleanup storage too.

JohnMoutafis: I have just copied the "delete"/deregister from the existing test.
It has a TODO about that same issue. Do we have a real delete method?

PR #538, opened 2024-03-18
Add workflow_dispatch CI trigger, and run 'fresh' workflow when so ca…

ihnorton:
Add workflow_dispatch CI trigger, and run 'fresh' workflow when so called.


ihnorton:
I'd like to merge this even with failing CI (known, see slack thread) so that we can run the "fresh" workflow and get more information.

davejrt (approved):

PR #542, opened 2024-03-21
Enable renaming in bioimg exporter

ktsitsi:
This PR:

- Replicates the logic of `bioimg.ingest` for allowing renaming of files during exportation.
- Adds `access_credentials_name` as a keyword argument and not as arbitrary
- Adds the `verbose` argument and activates a logger for internal UDFs
- Updates the name of `export_udf` to `export_batch` using the `as_batch`

### DIFF
--- a/src/tiledb/cloud/bioimg/exportation.py
+++ b/src/tiledb/cloud/bioimg/exportation.py
@@ -1,26 +1,31 @@
-from typing import Any, Dict, Iterator, Mapping, Optional, Sequence, Tuple, Union
+import logging
+from typing import Any, Iterator, Mapping, Optional, Sequence, Tuple, Union
 
 import tiledb
 from tiledb.cloud import dag
 from tiledb.cloud.bioimg.helpers import get_logger_wrapper
+from tiledb.cloud.bioimg.helpers import validate_io_paths
+from tiledb.cloud.dag.mode import Mode
+from tiledb.cloud.rest_api.models import RetryStrategy
+from tiledb.cloud.utilities._common import as_batch
 from tiledb.cloud.utilities._common import run_dag
 
 DEFAULT_RESOURCES = {"cpu": "8", "memory": "4Gi"}
 DEFAULT_IMG_NAME = "3.9-imaging-dev"
 DEFAULT_DAG_NAME = "bioimg-exportation"
-_RUNNING_PROFILES = ("client", "server")
 
 
 def export(
     source: Union[Sequence[str], str],
-    output: str,
+    output: Union[Sequence[str], str],
     *args: Any,
+    access_credentials_name: str,
     config: Optional[Mapping[str, Any]] = None,
     taskgraph_name: Optional[str] = None,
     num_batches: Optional[int] = None,
     resources: Optional[Mapping[str, Any]] = None,
     compute: bool = True,
-    run_on: Optional[str] = None,
+    mode: Optional[Mode] = Mode.BATCH,
     namespace: Optional[str] = None,
     verbose: bool = False,
     output_ext: str = "tiff",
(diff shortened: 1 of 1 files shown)


src/tiledb/cloud/bioimg/exportation.py
```diff
@@ -64,36 +76,66 @@ def build_io_uris_exportation(
...
-
-        if len(source) == 0:
-            raise ValueError("The source files cannot be empty")
-        return tuple(iter_paths(source))
+        def create_output_path(input_file: str, output: str) -> str:
+            # Check if output is dir
+            if output.endswith("/"):
+                filename = os.path.splitext(os.path.basename(input_file))[0]
+                output_filename = (
+                    filename + f".{output_ext}" if output_ext else filename
```
thetorpedodog, line 84: minor nit: `f"{filename}.{output_ext}"`. no need to + when you already have the f-string right there!

src/tiledb/cloud/bioimg/exportation.py
```diff
@@ -64,36 +76,66 @@ def build_io_uris_exportation(
...
+                )
+                return os.path.join(output, output_filename)
+            else:
```
thetorpedodog, line 87: no else after an unconditional return. you might rework this function as

```python
if not output.endswith("/"):
  return output
# the rest
```

just so most of the code isn’t indented

src/tiledb/cloud/bioimg/exportation.py
```diff
@@ -64,36 +76,66 @@ def build_io_uris_exportation(
...
+        verbose: bool,
     ):
+        logger = get_logger_wrapper(verbose)
+
         """Groups input URIs into batches."""
-        uri_pairs = build_io_uris_exportation(source, output, out_ext)
+        uri_pairs = build_io_uris_exportation(source, output, out_ext, logger)
         # If the user didn't specify a number of batches, run every import
         # as its own task.
+        logger.debug(f"Input batches:{uri_pairs}")
```
thetorpedodog, line 130: nit: when logging, use %-style strings and suffixes, so

```python
logger.debug("Input batches: %s", uri_pairs)
```

thetorpedodog (approved):

src/tiledb/cloud/bioimg/exportation.py
sgillies: @ktsitsi please use `%r` instead of `%s` so that we get the maximum information (like the difference between `b"1"` and `"1"`).

src/tiledb/cloud/bioimg/exportation.py
```diff
@@ -127,41 +169,45 @@ def export_tiff_udf(
...
+            limit=3,
+            retry_policy="Always",
+        ),
     )
 
+    if not access_credentials_name:
+        raise ValueError(
+            "Ingestion graph requires `access_credentials_name` to be set."
+        )
+
```
sgillies, line 195: @ktsitsi Is this the soonest we can validate the argument? Let's move it as high in the function as we can so that we can fail as quickly as possible.

src/tiledb/cloud/bioimg/exportation.py
```diff
@@ -1,26 +1,31 @@
...
 DEFAULT_DAG_NAME = "bioimg-exportation"
-_RUNNING_PROFILES = ("client", "server")
 
 
 def export(
     source: Union[Sequence[str], str],
-    output: str,
+    output: Union[Sequence[str], str],
     *args: Any,
+    access_credentials_name: str,
```
sgillies, line 22: Breaking this out of `kwargs` seems good to me 👍

sgillies (approved):
@ktsitsi I haven't run the code or any static checkers, but the changes look good to me. I made a suggestion and asked a question. I trust you to follow up, and I'll check approve 👍

src/tiledb/cloud/bioimg/exportation.py
ktsitsi: Addressed!

ihnorton (approved):

PR #544, opened 2024-03-29
Support using passed config for VFS listing in bioimage ingestion

Shelnutt2:
The bioimage ingestion needs to use the config for both ingestion/read (currently supported) and for the bucket listing (what this PR adds). This allows parameters like region to be supported.


src/tiledb/cloud/bioimg/ingestion.py
```diff
@@ -90,7 +91,8 @@ def build_io_uris_ingestion(
 
         import tiledb
 
-        vfs = tiledb.VFS()
+        ctx = tiledb.Ctx(config)
+        vfs = tiledb.VFS(ctx=ctx)
```
ktsitsi, line 95: Right now this VFS instance is getting the `access_credentials_name` (given during the node creation) and is being initialized with the node's default context including these access_credentials. Would not this overwrite the context of the instance, leading to misconfiguration?

src/tiledb/cloud/bioimg/ingestion.py
```diff
@@ -195,7 +198,7 @@ def ingest_tiff_udf(
                 raise ValueError
 
         write_context = tiledb.Ctx(config)
-        vfs = tiledb.VFS()
+        vfs = tiledb.VFS(ctx=write_context)
```
ktsitsi, line 201: Pretty similar here the reason we have this `write_context` is basically to allow the user to give a config for the destination path of the ingested file, if for some reason the destination S3 account is different from the one we use for reading the datasource with VFS. Again here the VFS is initialized with the default context of the node after node has the `access_credentials_name`.If `write_context` is None right now we assume that the source and destination S3 paths can be accessed with the credentials_name given in the pod, since the scope_ctx(None) does not create any issue.

ktsitsi (changes requested):

src/tiledb/cloud/bioimg/ingestion.py
Shelnutt2: Users shouldn't be passing credentials in to the config, only other settings like region. I don't think it'll be an issue for auth.

ihnorton (approved):

ktsitsi (approved):
Approving it but let's thoroughly test it in staging.

JohnMoutafis (approved):
Seems ok to me, but I do also agree with @ktsitsi about testing this and maybe adding unit and/or integration tests for that.

src/tiledb/cloud/bioimg/ingestion.py
```diff
@@ -340,7 +343,8 @@ def register_dataset_udf(
         _SUPPORTED_EXTENSIONS,
         *args,
         verbose=verbose,
-        access_credentials_name=acn,
```
spencerseale, line 343: This is breaking the main ingest. We list `acn` in the ingest function as an arg, but don't pass it into our first node. So this throws an access denied error if an access credential error is passed into `acn` rather than `access_credentials_name` @ktsitsi @ihnorton @Shelnutt2

PR #546, opened 2024-04-03
File ingestion implementation

JohnMoutafis:
Implements File ingestion into TileDB from VFS available sources.


src/tiledb/cloud/file_ingestion/ingestion.py
```diff
@@ -0,0 +1,177 @@
...
+    :param taskgraph_name: Optional name for taskgraph, defaults to "file-ingestion"
+    :param batch_size: Batch size for file ingestion, defaults to 100
+    :param resources: Configuration for node specs,
+        defaults to {"cpu": "2", "memory": "8Gi"}
+    :param verbose: Verbose logging, defaults to False
+    :return tiledb.cloud.dag.DAG: The resulting ingestion graph
+    """
+    # Preparation and argument cleanup
+    config = config or dict()
+    namespace = namespace or tiledb.cloud.user_profile().default_namespace_charged
```
Shelnutt2, line 117: I think we have a function about figuring out the default namespace

src/tiledb/cloud/file_ingestion/ingestion.py
```diff
@@ -0,0 +1,177 @@
...
+    resources = resources or DEFAULT_RESOURCES
+    dataset_uri = dataset_uri.rstrip("/")
+    logger = get_logger_wrapper(verbose)
+    if isinstance(source, str):
+        source = [source]
+
+    # logger.debug("Building the graph...")
```
Shelnutt2, line 124: Remove?

Shelnutt2 (commented):
Great start! A few minor comments

src/tiledb/cloud/file_ingestion/ingestion.py
JohnMoutafis: I am searching for that with no luck, do you remember how it is called?

src/tiledb/cloud/file_ingestion/ingestion.py
```diff
@@ -0,0 +1,177 @@
...
+    :param batch_size: Resulting chunk size, defaults to 100.
+    :return List[List[str]]: A list of chunks as lists.
+    """
+    # Reduce batch size if there are fewer sample URIs
+    batch_size = min(batch_size, len(udf_results))
+    num_chunks = ceil(len(udf_results) / batch_size)
+
+    flattened_results = [result for udf_result in udf_results for result in udf_result]
+    return [
+        flattened_results[n * batch_size : (n + 1) * batch_size]
```
NikolaosPapailiou, line 32: Do we have to check if we go beyond the size of `flattened_results`?

src/tiledb/cloud/file_ingestion/ingestion.py
```diff
@@ -0,0 +1,177 @@
...
+
+    :param dataset_uri: The dataset URI.
+    :param file_uris: An iterable of file URIs.
+    :param namespace: TileDB-Cloud namespace, defaults to None.
+    :param verbose: Verbose logging, defaults to False.
+    """
+    logger = get_logger_wrapper(verbose)
+
+    for file_uri in file_uris:
+        filename = file_uri.replace(",", "").replace(" ", "_")
```
NikolaosPapailiou, line 55: This flattening the entire path into a file name. Should we also replace `/`?

Also do we need to replicate the source directory structure using tiledb groups?

src/tiledb/cloud/file_ingestion/ingestion.py
```diff
@@ -0,0 +1,177 @@
...
+        filestore_array_uri = f"{dataset_uri}/{filename}"
+        filestore_array_s3_uri = f"s3{filestore_array_uri.split('s3')[1]}"
+        namespace = (
+            namespace or filestore_array_uri.split("tiledb://")[1].split("/s3")[0]
```
NikolaosPapailiou, line 59: Why do we assume that all `tiledb://` are of the format `tiledb://namespace/s3://..`?

src/tiledb/cloud/file_ingestion/ingestion.py
JohnMoutafis: Not really because even if we go out of size while slicing, the last slice will be as big as possible without errors.

src/tiledb/cloud/file_ingestion/ingestion.py
Shelnutt2, line 55: Should this use basename?

src/tiledb/cloud/file_ingestion/ingestion.py
Shelnutt2, line 59: We have functions to do this, if we want it.

src/tiledb/cloud/file_ingestion/ingestion.py
```diff
@@ -0,0 +1,177 @@
...
+        )
+        logger.debug(
+            f"""
+            ---------------------------------------------\n
+            - Filename: {filename}\n
+            - Namespace: {namespace}\n
+            - Array URI: {filestore_array_uri}\n
+            - S3 Array URI: {filestore_array_s3_uri}\n
```
Shelnutt2, line 67: This must support all permutations of backends not just s3. Please make this more generic.

src/tiledb/cloud/file_ingestion/ingestion.py
```diff
@@ -0,0 +1,177 @@
...
+            ---------------------------------------------
+            """
+        )
+
+        if not tiledb.array_exists(filestore_array_uri):
```
Shelnutt2, line 72: Let's have a message if the array already exists

src/tiledb/cloud/file_ingestion/ingestion.py
```diff
@@ -0,0 +1,177 @@
...
+    :param config: Config dictionary, defaults to None
+    :param namespace: TileDB-Cloud namespace, defaults to None
+    :param include: UNIX shell style pattern to filter files in the search,
+        defaults to None
+    :param exclude: UNIX shell style pattern to filter files out of the search,
+        defaults to None
+    :param taskgraph_name: Optional name for taskgraph, defaults to "file-ingestion"
+    :param batch_size: Batch size for file ingestion, defaults to 100
+    :param resources: Configuration for node specs,
+        defaults to {"cpu": "2", "memory": "8Gi"}
```
Shelnutt2, line 111: I think we can do smaller resources. Since you aren't ingesting the file client side this is really just a listing operation.

src/tiledb/cloud/file_ingestion/ingestion.py
Shelnutt2: https://github.com/TileDB-Inc/TileDB-Cloud-Py/blob/cf5749baea26267e7b448518cb3879c572b0749f/src/tiledb/cloud/client.py#L151

src/tiledb/cloud/file_ingestion/ingestion.py
```diff
@@ -0,0 +1,220 @@
...
+    # Preparation and argument cleanup
+    config = config or dict()
+    namespace = namespace or tiledb.cloud.user_profile().default_namespace_charged
+    resources = resources or DEFAULT_RESOURCES
+    dataset_uri = dataset_uri.rstrip("/")
+    logger = get_logger_wrapper(verbose)
+    if isinstance(source, str):
+        source = [source]
+
+    with tiledb.scope_ctx(config):
```
JohnMoutafis, line 158: @Shelnutt2 As I understand this, `config` should be a needed argument. Or is there a default `namespace` configuration that can be accessed by a method?

JohnMoutafis: @NikolaosPapailiou helped me understand this better. So it is resolved.

src/tiledb/cloud/file_ingestion/ingestion.py
```diff
@@ -0,0 +1,226 @@
...
+            logger.debug(
+                f"""
+                ---------------------------------------------
+                - Filename: {filename}
+                - Namespace: {namespace}
+                - Destination URI: {filestore_array_uri}
+                ---------------------------------------------"""
+            )
+
+            if not tiledb.array_exists(filestore_array_uri):
```
NikolaosPapailiou, line 108: I think we need to expose file overwrite options to the user or at least make sure that the behaviour is well documented. At the moment, this silently skips file overwrites without any documentation that this should be expected.

src/tiledb/cloud/file_ingestion/ingestion.py
```diff
@@ -0,0 +1,226 @@
...
+    config: Optional[dict] = None,
+    namespace: Optional[str] = None,
+    verbose: bool = False,
+) -> None:
+    """
+    Ingest files.
+
+    :param dataset_uri: The dataset URI.
+    :param file_uris: An iterable of file URIs.
+    :param destination: URI to ingest the files into.
```
NikolaosPapailiou, line 70: Can `destination` be inferred by the `dataset_uri`? Assuming that `dataset_uri` is a group we can get its S3 path and use it for `destination`. If there is no `dataset_uri` we can write the files to the default storage location of the user.

JohnMoutafis: Current UI implementation provides a destination though.

NikolaosPapailiou: This can still be empty and get some default value from the provided group or default namespace

Shelnutt2: The user should be required to pass the storage location. Let's keep all the ingestors consistent. All other ingestors have this value passed today.

Future improvements to all ingestors can be using defaults.

src/tiledb/cloud/file_ingestion/ingestion.py
JohnMoutafis: @Shelnutt2 The current implementation produces a warning in the logs and moves on to the next file of the chunk (if any):  https://github.com/TileDB-Inc/TileDB-Cloud-Py/blob/55e4625a5435eee9ca2a4ddd5a0e8f27bd182714/src/tiledb/cloud/file_ingestion/ingestion.py#L122

src/tiledb/cloud/file_ingestion/ingestion.py
```diff
@@ -0,0 +1,226 @@
...
+        config=config,
+        verbose=verbose,
+        # Expand list of results into multiple node operations
+        expand_node_output=chunks,
+        name="Ingest file URIs",
+        resources=resources,
+        access_credentials_name=acn,
+    )
+
+    return graph
```
ktsitsi, line 223: This will fail with the `as_batch`.  We might need a conditional here.

src/tiledb/cloud/file_ingestion/ingestion.py
```diff
@@ -0,0 +1,226 @@
+import os
```
Shelnutt2, line 1: Lets rename the package `file` or `files`. We can deprecate the existing `file.py` and move packages to this new submodule. I'd like to standardize the naming.

Shelnutt2:
Please also add a description to this PR. A well formed description is required for all pull requests.

src/tiledb/cloud/file_ingestion/ingestion.py
```diff
@@ -0,0 +1,226 @@
...
+                tiledb.cloud.file.create_file(
+                    namespace=namespace,
+                    name=filename,
+                    input_uri=file_uri,
+                    output_uri=filestore_array_uri,
+                    access_credentials_name=acn,
+                )
+
+                if is_group:
+                    with tiledb.Group(dataset_uri, mode="w") as group:
```
Shelnutt2, line 118: Don't add to the group one at a time. Do it in a single batch at the end.

JohnMoutafis: After researching a bit this request, I concluded that I can do that by using the V2 API's `update_group_contents` from [here](https://github.com/TileDB-Inc/TileDB-Cloud-Py/blob/62524dd5c7cba945c754e4a7e7a65e7fa3302d05/src/tiledb/cloud/_common/api_v2/api/groups_api.py#L938).
Is this the correct way to approach that?

src/tiledb/cloud/file_ingestion/ingestion.py
JohnMoutafis: Although this change makes sense, I don't think this PR should do that on top of the new functionality.

antalakas (approved):
Looks good to me after tests are added

antalakas (changes requested):
After syncing with @JohnMoutafis we agreed we need to rework the signature of the ingestion

> def ingest_files(
    dataset_uri: str,
    destination: str,
   
  It should be possible to make it as close to other ingestors, standardizing as much as possible the ingestion mechanism

antalakas (approved):
lgtm, thanks @JohnMoutafis for the recent changes

PR #549, opened 2024-04-05
Add DAG._find_leaf_nodes helper

Shelnutt2:
This function is helpful for returning the leaves for the DAG. Its effectively the opposite of `_find_root_nodes`. It can be helpful for users that want to add a new node(s) that depend on the current leaves.

### DIFF
--- a/src/tiledb/cloud/dag/dag.py
+++ b/src/tiledb/cloud/dag/dag.py
@@ -1139,6 +1139,13 @@ class DAG:
 
         return roots
 
+    def _find_leaf_nodes(self):
+        """
+        Find all leaf nodes
+        :return: list of leaf nodes
+        """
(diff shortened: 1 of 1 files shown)


src/tiledb/cloud/dag/dag.py
```diff
@@ -1139,6 +1139,18 @@ def _find_root_nodes(self):
...
+        """
+        Find all leaf nodes
+        :return: list of leaf nodes
+        """
+        leaves = []
+        for node in self.nodes.values():
+            if node.children is None or len(node.children) == 0:
+                leaves.append(node)
+
+        return leaves
```
thetorpedodog, line 1152: `return [n for n in self.nodes.values() if not node.children]`

PR #550, opened 2024-04-10
Fix UnboundLocalError when accessing kwargs dict that doesn't exist.

JohnMoutafis:
Fixes reported error: `Cannot access local variable 'kwargs' when setting verify_ssl in cloud.login`


Shelnutt2 (approved):

antalakas:
[sc-44931]

antalakas (approved):

PR #552, opened 2024-04-12
Graph should not be returned when computed

ktsitsi:
This PR:

- Adds an `else` statement in ingestion function avoiding the need to `pickle` the results of the ingestion function, when the latter is wrapped with `as_batch` and called as `ingest_batch`. The error comes from the fact that the results of a UDF are serialized to return. The DAG returned before the change is not serializable though as it contains  `_thread.lock` object.
- The change does NOT break the API - used by customers -  since the graph is getting is still returned when the `compute` arg is set to `False`. It will only require for the caller of the `ingest_batch` function to set `compute=True` to follow the correct code path.
- Fixes the following error:
```bash
Unexpected error in the UDF environment:
  File "/opt/conda/lib/python3.9/site-packages/tdbudf/batch_udf_main.py", line 359, in main
    ret = real_main()
  File "/opt/conda/lib/python3.9/site-packages/tdbudf/batch_udf_main.py", line 351, in real_main
    _serialize(result, result_format, expand_output, stage_part_id)
  File "/opt/conda/lib/python3.9/site-packages/tdbudf/batch_udf_main.py", line 96, in _serialize
    serialized_tgargs = _serialize_to_tgargs(result, result_format)
  File "/opt/conda/lib/python3.9/site-packages/tdbudf/batch_udf_main.py", line 52, in _serialize_to_tgargs
    cloudpickle.dumps(result, protocol=_PICKLE_PROTOCOL)
  File "/opt/conda/lib/python3.9/site-packages/cloudpickle/cloudpickle_fast.py", line 73, in dumps
    cp.dump(obj)
  File "/opt/conda/lib/python3.9/site-packages/cloudpickle/cloudpickle_fast.py", line 632, in dump
    return Pickler.dump(self, obj)
TypeError: cannot pickle '_thread.lock' object
```

### DIFF
--- a/src/tiledb/cloud/bioimg/ingestion.py
+++ b/src/tiledb/cloud/bioimg/ingestion.py
@@ -386,7 +386,8 @@ def ingest(
         )
     if compute:
         run_dag(graph, debug=verbose)
-    return graph
+    else:
+        return graph
 
 
 # Wrapper function for batch VCF ingestion


sgillies (commented):
@ktsitsi I've got a question about my understanding of the situation and a suggestion that may or may not be good, depending on my understanding.

src/tiledb/cloud/bioimg/ingestion.py
```diff
@@ -386,7 +386,8 @@ def register_dataset_udf(
         )
     if compute:
         run_dag(graph, debug=verbose)
-    return graph
+    else:
+        return graph
```
sgillies, line 390: @ktsitsi to sum things up: a graph that has been run can't be returned, but a graph that hasn't been run can be pickled and returned. Is that right?

I'm uncomfortable with this if/else from an API design perspective. I may be missing something, but it looks like unnecessary complexity. Could we do the following instead when we want to run the graph?

```python
graph = register_dataset_udf(...)
run_dag(graph)
```

It's less code and has less ambiguity about intent.

sgillies (commented):
Oops, I forgot to activate the comment.

src/tiledb/cloud/bioimg/ingestion.py
Shelnutt2: @sgillies I believe the problem is there are some components that get created during execution, i.e a monitor thread: https://github.com/TileDB-Inc/TileDB-Cloud-Py/blob/9a14c4e9710ee9d60e0d3f46fdd40eddc3ad49d7/src/tiledb/cloud/dag/dag.py#L1173 . So a graph that is running or has completed might contain components that cloudpickle can't handle.

A separate work item should be to cleanup these classes so we don't need to differentiate this.

ktsitsi:
So guys is there any other comment we need to address?

sgillies (approved):
@ktsitsi thanks for giving me more context about the need for this.

PR #555, opened 2024-04-23
Support BED arrays in VCF queries

gspowley:
Add support for TileDB BED arrays in VCF queries.


gspowley:
This PR is no longer required after a change in the VCF read API allows BED arrays to be passed with the existing `bed_file` parameter.

PR #559, opened 2024-04-30
Rename folder file_ingestion to files

alekshs:
Renamed `file_ingestion` folder to `files` to match the rest of the folder structure. `file_ingestion` refers to a specific feature while `files` is more generic and can include more code in the future related to files but not necessarily related to 1-click ingestion.


Shelnutt2:
I agree with the change, but I'd prefer someone more involved to approve for merge.

alekshs:
#560 includes this change. closing

PR #560, opened 2024-05-01
File ingestion follow ups.

JohnMoutafis:
- Fixes exclude/include filters
- Introduces a file ingestion `find_uris_udf` that utilizes `utilities/_common.py -> find`
- Renames `file_ingestion` to `file` and moves the `file.py` in that folder, as `utils.py` for semantic/cohesion reasons
- Refactors `find_uris_udf` and `chunk_udf` as common UDFs in the `file/udfs.py` file
- Modifies and adds unit tests accordingly.


src/tiledb/cloud/file/ingestion.py
```diff
@@ -10,54 +8,13 @@
...
-    :param fname: A filename to sanitize
-    :return str: The sanitized string
-    """
-    name, suffix = os.path.splitext(fname)
-    name = re.sub(r"[^\w\s-]", "", name)
-    name = re.sub(r"[-_\s]+", "_", name).strip("-_")
-    return name + suffix
-
-
-def chunk_results_udf(
```
ktsitsi, line 35: Has the move of these functions to another module been tested? Does it work properly with `cloudpickle`?

ktsitsi (commented):
LGTM overall just a minor question in comments.

src/tiledb/cloud/file/ingestion.py
JohnMoutafis: We will test this in a review app soon

src/tiledb/cloud/file/udfs.py
```diff
@@ -0,0 +1,88 @@
...
+from math import ceil
+from typing import Any, List, Mapping, Optional, Sequence, TypeVar
+
+from tiledb.cloud.file import utils as file_utils
+from tiledb.cloud.utilities import chunk
+from tiledb.cloud.utilities import find
+from tiledb.cloud.utilities import get_logger_wrapper
+
+DEFAULT_BATCH_SIZE = 100
+T = TypeVar("T")
```
thetorpedodog, line 11: nit: to avoid these looking like they should be exportable, I tend to call them with an underscore, like `_T`.

src/tiledb/cloud/file/udfs.py
```diff
@@ -0,0 +1,88 @@
...
+    :param include: include pattern used in the search, defaults to None
+    :param exclude: exclude pattern applied to the search results, defaults to None
+    :param max_files: maximum number of URIs returned, defaults to None
+    :param verbose: verbose logging, defaults to False
+    :return: list of URIs
+    """
+    logger = get_logger_wrapper(verbose)
+
+    search_uri = search_uri.rstrip("/") + "/"
+    logger.info(f"Searching URI: {search_uri}...")
```
thetorpedodog, line 39: For logging, we should use % formatting to avoid building the message when it's unused (and preserve better semantic info): `logger.info("Searching URI: %s", search_uri)`

src/tiledb/cloud/file/udfs.py
```diff
@@ -0,0 +1,88 @@
...
+
+    if include:
+        include = partial(file_utils.basename_match, pattern=include)
+    if exclude:
+        exclude = partial(file_utils.basename_match, pattern=exclude)
+
+    results = [
+        *find(
```
thetorpedodog, line 47: `[*whatever]` strikes me as odd. `list(whatever)` instead?

src/tiledb/cloud/file/udfs.py
```diff
@@ -0,0 +1,88 @@
...
+
+    :param udf_results: An iterable of iterables containing UDF results.
+    :param batch_size: Resulting chunk size, defaults to None.
+    :return List[List[str]]: A list of chunks as lists.
+    """
+    logger = get_logger_wrapper(verbose)
+
+    if flatten_items:
+        # Flatten the results into a list.
+        items = [elem for item in items for elem in item]
```
thetorpedodog, line 77: `itertools.chain.from_iterable` might make this simpler to read.

src/tiledb/cloud/file/utils.py
```diff
@@ -10,6 +13,31 @@
...
+def basename_match(file_uri: str, pattern: Optional[str] = None) -> bool:
+    """
+    Checks if the basename of a given file uri matches
+    the given UNIX shell style pattern.
+
+    :param file_uri: A file URI.
+    :param pattern: A UNIX shell style pattern, defaults to None
+    :return bool: Pattern matches the file basename or not.
+    """
+    return pattern is not None and fnmatch(os.path.basename(file_uri), pattern)
```
thetorpedodog, line 38: If the pattern is None, do we want to reject or accept?

thetorpedodog (commented):
Nothing major here, just a few notes and a question.

src/tiledb/cloud/file/udfs.py
JohnMoutafis: Sounds good, will add that!

src/tiledb/cloud/file/utils.py
JohnMoutafis: If the pattern is `None` then it cannot match or we don't care about matching (this is for the method to be more general). 
Also guards against a type error being raised from fnmatch in that case.

thetorpedodog: My thought here is that if we don't care about matching, then we would want to always say yes.

src/tiledb/cloud/file/udfs.py
JohnMoutafis: I think this is usually named that way in order to make sense to C/C++ engineers, but I get your point.

src/tiledb/cloud/file/udfs.py
JohnMoutafis: I was today years old when I found out about that for logging!
Will fix.

src/tiledb/cloud/file/utils.py
JohnMoutafis: We do care about matching, that is why if the `pattern` is `None` we return `False`.
Also semantically, the method's name is `basename_match` so with `None` I would expect not to match

src/tiledb/cloud/file/ingestion.py
JohnMoutafis: @ktsitsi We have tested the whole process in a review app without issues with pickling (or other issues thus far).

ktsitsi (approved):
LGTM! Thank you @JohnMoutafis 🚀

src/tiledb/cloud/file/ingestion.py
```diff
@@ -10,54 +8,13 @@
 from tiledb.cloud._common import utils
 from tiledb.cloud.array import info
 from tiledb.cloud.dag.mode import Mode
+from tiledb.cloud.file import udfs as file_udfs
+from tiledb.cloud.file import utils as file_utils
 from tiledb.cloud.utilities import as_batch
 from tiledb.cloud.utilities import get_logger_wrapper
-from tiledb.cloud.vcf.ingestion import find_uris_udf
```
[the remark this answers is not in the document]
JohnMoutafis: True, I believe that this PR can work as the starting point for that consolidation.

src/tiledb/cloud/file/ingestion.py
```diff
@@ -82,7 +39,7 @@ def add_arrays_to_group_udf(
             if tiledb.object_type(group_uri) == "group":
                 with tiledb.Group(group_uri, mode="w") as group:
                     logger.debug(
-                        f"Adding to Group {group_uri} the following: {array_uris}"
+                        "Adding to Group %s the following: %s" % (group_uri, array_uris)
```
thetorpedodog, line 42: Like `logger.debug("Adding to group %s the following: %s", group_uri, array_uris)` (i.e., the %-formatting gets done inside `logger.debug`, so it doesn’t have to do anything at all if the log message will never be emitted)

src/tiledb/cloud/file/ingestion.py
```diff
@@ -124,22 +83,23 @@ def ingest_files_udf(
...
+            """
             ---------------------------------------------
-            - Input URI: {file_uri}
-            - Sanitized Filename: {filename}
-            - Array URI: {array_uri}
-            - Destination URI: {filestore_array_uri}
+            - Input URI: %s
+            - Sanitized Filename: %s
+            - Array URI: %s
+            - Destination URI: %s
```
thetorpedodog, line 96: I would just make this four separate log statements.

thetorpedodog (approved):
one minor thing, but other than this it looks good!

thetorpedodog:
In the future, when we do something like this, we should keep around the old name of any functions used in the UI as an alias, to avoid the need for exactly synchronized releases. Once the UI change is complete, then we can remove the alias.

Shelnutt2:
> In the future, when we do something like this, we should keep around the old name of any functions used in the UI as an alias, to avoid the need for exactly synchronized releases. Once the UI change is complete, then we can remove the alias.

Agreed. We also offer compatibility, any breaking changes we need to go through a deprecation cycle on.

PR #561, opened 2024-05-02
Fix Tasks API mismatch between client and server for array parameter

JohnMoutafis:
The server expects `array` to be a TileDB URI, but in the `tasks.tasks` method, if the `array` param was present, it was overwritten by the array name isntead, causing an error response:
`TileDBCloudError: unable to determine array namespace/name from URI (expected "tiledb://<namespace>/<name>"): quickstart_sparse - Code: 24`


Shelnutt2 (approved):

nickvigilante (approved):
Thank you for the quick fix!

PR #566, opened 2024-05-21
An asset module and simple API

sgillies:
To allow abstract asset management and querying.

More methods are useful, but `delete()` and `info()` are the only ones (I think) that currently exist in `tiledb.cloud.array` and `tiledb.cloud.groups` in the same form. Fleshing out `tiledb.cloud.groups` and filling other gaps is work for another PR.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,0 +1,6 @@
+# Changes
+
+## Next (YYYY-MM-DD)
+
+- Add a tiledb.cloud.asset module with functions that allow management of
+  assets of any type (gh-566).
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,22 +29,21 @@ viz-plotly = ["networkx>=2", "plotly>=4", "pydot"]
 all = ["networkx>=2", "plotly>=4", "pydot", "tiledb-plot-widget>=0.1.7"]
 
 dev = ["black", "pytest", "ruff"]
-tests = ["xarray", "pytest-split"]
+tests = ["xarray", "pytest-cov", "pytest-explicit", "pytest-split"]
 
 [project.urls]
 homepage = "https://tiledb.com"
 repository = "https://github.com/TileDB-Inc/TileDB-Cloud-Py"
 
-
 [build-system]
 requires = ["setuptools>=42", "wheel", "setuptools_scm>=6"]
 
 [tool.pytest.ini_options]
+explicit-only = ["geospatial"]
 markers = [
   "geospatial: tests that require the geospatial libraries",
 ]
 norecursedirs = ["tiledb/cloud"]
-
 
 [tool.setuptools]
 zip-safe = false
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,7 +40,6 @@ repository = "https://github.com/TileDB-Inc/TileDB-Cloud-Py"
 requires = ["setuptools>=42", "wheel", "setuptools_scm>=6"]
 
 [tool.pytest.ini_options]
-explicit-only = ["geospatial"]
 markers = [
   "geospatial: tests that require the geospatial libraries",
 ]
--- a/src/tiledb/cloud/asset.py
+++ b/src/tiledb/cloud/asset.py
@@ -3,9 +3,12 @@
 from functools import partial
 from typing import Union
 
-import tiledb.cloud  # type: ignore
-from tiledb.cloud.rest_api.models import ArrayInfo  # type: ignore
-from tiledb.cloud.rest_api.models import GroupInfo  # type: ignore
+import tiledb  # type: ignore
+
+from . import array  # type: ignore
+from . import groups  # type: ignore
+from .rest_api.models import ArrayInfo  # type: ignore
+from .rest_api.models import GroupInfo  # type: ignore
 
 
 def delete(uri: str, recursive: bool = False) -> None:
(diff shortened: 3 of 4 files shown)


tests/test_asset.py
```diff
@@ -0,0 +1,36 @@
...
+@mock.patch("tiledb.cloud.array.info")
+def test_asset_info_array(array_info, object_type):
+    """Dispatch to array.info when URI is an array."""
+    object_type.return_value = "array"
+    array_info.return_value = {"tiledb_uri": "tiledb://a"}
+
+    def info(uri):
+        asset_map = {"array": tiledb.cloud.array, "group": tiledb.cloud.groups}
+        mod = asset_map[tiledb.object_type(uri)]
+        return getattr(mod, "info")(uri)
```
sgillies, line 19: Implementation is in the test at this stage.

src/tiledb/cloud/asset.py
```diff
@@ -0,0 +1,40 @@
...
+
+    :param uri: URI of the asset.
+    :return: None.
+    """
+    delete_map = {
+        "array": array.delete_array,
+        "group": partial(groups.delete, recursive=recursive),
+    }
+    asset_type = tiledb.object_type(uri, ctx=tiledb.cloud.Ctx())
+    func = delete_map[asset_type]
```
sgillies, line 25: Instead of, e.g.,

```python
if asset_type == "array":
    return array.delete_array(uri)
elif asset_type == "group":
    return groups.delete(uri)
else:
    raise Exception(...)
```

Note that moving these maps up to the module scope complicates unit testing. Maps inside the function allows us to patch public API tiledb.cloud functions as late as we want.

CHANGES.md
```diff
@@ -0,0 +1,6 @@
+# Changes
+
+## Next (YYYY-MM-DD)
+
+- Add a tiledb.cloud.asset module with functions that allow management of
+  assets of any type (gh-566).
```
sgillies, line 6: Start of a change log.

pyproject.toml
```diff
@@ -29,13 +29,12 @@ viz-plotly = ["networkx>=2", "plotly>=4", "pydot"]
 all = ["networkx>=2", "plotly>=4", "pydot", "tiledb-plot-widget>=0.1.7"]
 
 dev = ["black", "pytest", "ruff"]
-tests = ["xarray", "pytest-split"]
+tests = ["xarray", "pytest-cov", "pytest-explicit", "pytest-split"]
```
sgillies, line 32: We have some config that depends on pytest-explicit and are using pytest-cov in CI.

src/tiledb/cloud/asset.py
```diff
@@ -0,0 +1,40 @@
...
+
+from functools import partial
+from typing import Union
+
+import tiledb  # type: ignore
+
+from . import array  # type: ignore
+from . import groups  # type: ignore
+from .rest_api.models import ArrayInfo  # type: ignore
+from .rest_api.models import GroupInfo  # type: ignore
```
sgillies, line 11: I'm running mypy on this module and its tests locally and it passes.

tests/test_asset.py
```diff
@@ -0,0 +1,66 @@
...
+from tiledb.cloud.rest_api.models import GroupInfo  # type: ignore
+
+
+@mock.patch("tiledb.object_type", return_value="array")
+@mock.patch("tiledb.cloud.array.delete_array")
+def test_asset_delete_array_dispatch(delete_array, object_type):
+    """Dispatch to array.array_delete when URI is an array."""
+    delete("a")
+    # Since delete() doesn't return a value, we assert on the implementation.
+    delete_array.assert_called_once_with("a")
```
sgillies, line 17: Here's a unit test. Every called function is mocked.

tests/test_asset.py
```diff
@@ -0,0 +1,66 @@
...
+    asset_info = info("g")
+    assert isinstance(asset_info, GroupInfo)
+    assert asset_info.tiledb_uri == "tiledb://g"
+
+
+def test_public_array_asset_info():
+    """Get info about a public production array."""
+    asset_info = info("tiledb://TileDB-Inc/cd89a0d6-c262-4729-9e75-d942879e1d7d")
+    assert asset_info.name == "documents"
+    assert asset_info.type == "sparse"
```
sgillies, line 60: This is not a unit test. There are no mocks. It retrieves information about a public array.

sgillies:
@JohnMoutafis @thetorpedodog I'm building out a new module with a high-level API for managing assets. Let me know if you have questions. The tests are a mix of fast unit tests and slower tests on actual data. I don't think live data tests for `delete()` are crucial right now. They're mostly an exercise in configuration and I think the unit tests are good for now. I'll add something when I get to implementing sharing/unsharing/registration, etc.

src/tiledb/cloud/asset.py
JohnMoutafis: I totally support your arguments here.
Another option here to avoid recreating the mappings in every function call would be to store them in a Class. 
Example just for further discussion: 
```python
class Asset:
    def __init__(self, ...):
        self.delete_map = {
           "array": array.delete_array,
           "group": partial(groups.delete, recursive=recursive),
        }
        self.info_map = {"array": array.info, "group": groups.info}

    def delete(self, uri: str, recursive: bool = False) -> None:
        asset_type = tiledb.object_type(uri, ctx=tiledb.cloud.Ctx())
        func = self.delete_map[asset_type]
        return func(uri)

    ....


asset = Asset()
```

Also mappings etc. can be modified through `__init__`

JohnMoutafis (approved):
lgtm

PR #567, opened 2024-05-29
Rename asset functions, export from tiledb.cloud.

sgillies:
Exports asset_info and delete_asset from `tiledb.cloud` alongside array functions.

### DIFF
--- a/src/tiledb/cloud/asset.py
+++ b/src/tiledb/cloud/asset.py
@@ -11,10 +11,10 @@ from .rest_api.models import ArrayInfo  # type: ignore
 from .rest_api.models import GroupInfo  # type: ignore
 
 
-def delete_asset(uri: str, recursive: bool = False) -> None:
+def delete(uri: str, recursive: bool = False) -> None:
     """Deregister the asset and remove its physical groups and arrays from storage.
 
-    :param uri: URI of the asset.
+    :param uri: tiledb URI of the asset.
     :return: None.
     """
     delete_map: Mapping[str, Callable] = {
@@ -26,10 +26,10 @@ def delete_asset(uri: str, recursive: bool = False) -> None:
     return func(uri)
 
 
-def asset_info(uri: str) -> Union[ArrayInfo, GroupInfo]:
+def info(uri: str) -> Union[ArrayInfo, GroupInfo]:
     """Retrieve information about an asset.
 
-    :param uri: URI of the asset.
+    :param uri: tiledb URI of the asset.
     :return: ArrayInfo or GroupInfo.
     """
     # Note: the URI can be either of the two forms, yes?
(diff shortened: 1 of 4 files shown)


sgillies:
Test failures above are task graph related.

src/tiledb/cloud/asset.py
```diff
@@ -11,30 +11,30 @@
...
 
-def delete(uri: str, recursive: bool = False) -> None:
+def delete_asset(uri: str, recursive: bool = False) -> None:
     """Deregister the asset and remove its physical groups and arrays from storage.
 
     :param uri: URI of the asset.
     :return: None.
     """
-    delete_map = {
+    delete_map: Mapping[str, Callable] = {
```
sgillies, line 20: I don't plan to add type hints to everything, but this is a key piece of data. Mypy, for example, will warn us if we misuse it.

src/tiledb/cloud/asset.py
```diff
@@ -11,30 +11,30 @@
...
         "group": partial(groups.delete, recursive=recursive),
     }
-    asset_type = tiledb.object_type(uri, ctx=tiledb.cloud.Ctx())
+    asset_type: str = tiledb.object_type(uri, ctx=tiledb.cloud.Ctx())
     func = delete_map[asset_type]
     return func(uri)
 
 
-def info(uri: str) -> Union[ArrayInfo, GroupInfo]:
+def asset_info(uri: str) -> Union[ArrayInfo, GroupInfo]:
```
sgillies, line 29: As soon as I realized we want to export this from tiledb.cloud for academy examples it had to be renamed to avoid a clash with `array.info`.

sgillies:
@JohnMoutafis @thetorpedodog Yes? No? Questions?

src/tiledb/cloud/asset.py
Shelnutt2: What was the clash? Aren't they separated by `tiledb.cloud.asset.info` and `tiledb.cloud.array.info` (and `tiledb.cloud.group.info`)? Or is there export issues?

JohnMoutafis: Although this is mostly a detail, I do prefer having the same API for everything and differentiate in the import statements if we are going to use both:
Either:
```python
from tiledb.cloud import asset
from tiledb.cloud import array

asset.info(...)
array.info(...)
```  
Or:
```python
from tiledb.cloud.asset import info as asset_info
from tiledb.cloud.array import info as array_info

asset_info(...)
array_info(...)
```

Shelnutt2 (approved):

Shelnutt2:
I approved but we also had discussed leaving them as `asset.info` and just including the `asset` module on import. I'm good either way, lets just make a decision so we can merge and release as I love the new APIs and they will help a lot with documentation work.

PR #568, opened 2024-05-30
Remove `dataset_type` from registered array metadata for geospatial ingest.

JohnMoutafis:
Removes the `dataset_type` metadata field to enable the UI to read PointClouds and Rasters as arrays or groups.

### DIFF
--- a/src/tiledb/cloud/geospatial/ingestion.py
+++ b/src/tiledb/cloud/geospatial/ingestion.py
@@ -962,6 +962,13 @@ def register_dataset_udf(
                 access_credentials_name=acn,
             )
 
+            # FIXME: This removes `dataset_type` metadata field for internal UI reasons
+            with tiledb.open(dataset_uri, mode="w") as array:
+                try:
+                    del array.meta["dataset_type"]
+                except KeyError:
+                    pass
+
 
 def build_file_list_udf(
     *,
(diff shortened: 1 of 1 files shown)


src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -962,6 +962,13 @@ def register_dataset_udf(
                 access_credentials_name=acn,
             )
 
+            # FIXME: This removes `dataset_type` metadata field for internal UI reasons
```
ihnorton, line 965: We need to check if the output is an array - according to the discussions in the ticket, it will be a group for multi-LOD data.

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -1402,6 +1412,30 @@ def consolidate_meta(
...
 
 
+def remove_dataset_type_from_array_meta(
+    dataset_uri: str,
+    *,
+    verbose: bool = False,
+):
+    """
+    Removes `dataset_type` meta if the ingested result is an array.
+    FIXME: This exists to fix an internal UI issue until formally fixed.
```
ihnorton, line 1422: Let's crossref `sc-nnnnn` here so future readers can track down the discussion.

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -1402,6 +1412,30 @@ def consolidate_meta(
...
+    logger = get_logger_wrapper(verbose)
+    if tiledb.object_type(dataset_uri) == "array":
+        logger.info(
+            "Ingested result is an array. " "Removing dataset_type from metadata..."
+        )
+        with tiledb.open(dataset_uri, mode="w") as array:
+            try:
+                del array.meta["dataset_type"]
+            except KeyError as exc:
+                logger.debug(f"Failed: {str(exc)}")
```
ihnorton, line 1436:
```suggestion
                logger.debug(f"Failed to remove `dataset_type` key: {str(exc)}")
```

ihnorton (approved):

src/tiledb/cloud/geospatial/ingestion.py
```diff
@@ -1402,6 +1412,30 @@ def consolidate_meta(
...
+    Removes `dataset_type` meta if the ingested result is an array.
+    FIXME: This exists to fix an internal UI issue until formally fixed.
+
+    :param dataset_uri: dataset URI
+    :param verbose: verbose logging, defaults to False
+    """
+    logger = get_logger_wrapper(verbose)
+    if tiledb.object_type(dataset_uri) == "array":
+        logger.info(
+            "Ingested result is an array. " "Removing dataset_type from metadata..."
```
sgillies, line 1430: What would you think about some structured data at the end of the log message?

```suggestion
            "Removing non-standard dataset_type from ingested array metadata: dataset_uri=%r", dataset_uri
```

src/tiledb/cloud/geospatial/ingestion.py
sgillies: Maybe info level so that it accompanies the previous message in a log file?

sgillies (approved):
Looks good. I made a couple of minor suggestions.

PR #569, opened 2024-05-31
Make soma an optional dependency

Shelnutt2:
SOMA was added a dependency for TileDB-Cloud-Py as part of ingestion work. Other libraries we treat as optional to have a middle ground for the amount of dependencies and packages required for a base installation of TileDB-Cloud-Py. SOMA should be moved to optional too.

This also lets tiledb-cloud be installed on windows machines where SOMA doesn't currently offer binaries.


JohnMoutafis (approved):
lgtm

antalakas (approved):

PR #571, opened 2024-06-03
Patch rest.py to avoid deprecated urllib3 usage

sgillies:
urllib3's `Response.getheaders()` is deprecated. With this patch, we can successfully run `python -Werrors -m pytest tests/test_asset.py` (turning warnings into errors).

The `update_generated.sh` script seems to be Linux-only. With some modification to the sed commands (`-i ''` instead of `--in-place`), I could get the script to run on my Mac, but the original json_safe patch no longer applies.

I'll look into tuning up that patch so that it will apply before I ask for review.


sgillies:
I have to give up on this for now. The code I'm generating doesn't match the patch.

PR #575, opened 2024-06-04
To groups.py add share_group, unshare_group, list_shared_with, and a test

sgillies:
Coverage of the new code is good (measured by `coverage run --include src/tiledb/cloud/groups.py -m pytest tests/test_groups.py`). We lack tests of errors sent by the API server, but the code around that is implicitly trusted boilerplate.

Once this code is merged, I plan to wrap these up in functions in the new `tiledb.cloud.asset` module.

### DIFF
--- a/src/tiledb/cloud/groups.py
+++ b/src/tiledb/cloud/groups.py
@@ -5,12 +5,16 @@ import posixpath
 import urllib.parse
 from typing import Iterable, List, Optional, Tuple, Union
 
-import tiledb.cloud.tiledb_cloud_error as tce
-from tiledb.cloud import client
-from tiledb.cloud import rest_api
-from tiledb.cloud._common import api_v2
-from tiledb.cloud._common import utils
-from tiledb.cloud.rest_api.models import group_update
+from . import client
+from . import rest_api
+from . import tiledb_cloud_error
+from ._common import api_v2
+from ._common import utils
+from .rest_api import ApiException as GenApiException
+from .rest_api import models
+from .rest_api.models import group_update
+
+split_uri = utils.split_uri
 
 
 def create(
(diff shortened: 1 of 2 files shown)


src/tiledb/cloud/groups.py
```diff
@@ -5,12 +5,16 @@
...
+from . import client
+from . import rest_api
+from . import tiledb_cloud_error
+from ._common import api_v2
+from ._common import utils
+from .rest_api import ApiException as GenApiException
+from .rest_api import models
+from .rest_api.models import group_update
+
+split_uri = utils.split_uri
```
sgillies, line 17: Relative imports like in the array module.

src/tiledb/cloud/groups.py
```diff
@@ -173,7 +177,9 @@ def deregister(
                     grp: rest_api.GroupInfo = m.group
                     deregister(grp.tiledb_uri, recursive=recursive)
                 else:
-                    raise tce.TileDBCloudError("unexpected group member type")
+                    raise tiledb_cloud_error.TileDBCloudError(
+                        "unexpected group member type"
+                    )
```
sgillies, line 182: I used functions in `tiledb.cloud.array` as a template and since they were using `tiledb_cloud_error` unaliased, I dealiased our existing code.

src/tiledb/cloud/groups.py
```diff
@@ -236,3 +242,70 @@ def _add_to(*, namespace: str, name: str, parent_uri: str) -> None:
...
+    :return: None.
+    """
+
+    if not isinstance(permissions, list):
+        permissions = [permissions]
+
+    if set([perm.lower() for perm in permissions]) - {
+        models.GroupActions.READ,
+        models.GroupActions.WRITE,
+    }:
```
sgillies, line 281: I used a set comparison here instead of a loop over individual permissions. A minor optimization.

sgillies:
@JohnMoutafis @thetorpedodog yea or nay? Other suggestions?

JohnMoutafis (approved):
lgtm

PR #576, opened 2024-06-05
Add streaming file uploads.

thetorpedodog:
This adds the ability to upload a file to TileDB Cloud, using a streaming request to avoid loading the whole file in memory at once.

[sc-48678]


src/tiledb/cloud/files/utils.py
```diff
@@ -164,3 +170,78 @@ def export_file(
         )
     except GenApiException as exc:
         raise tiledb_cloud_error.check_exc(exc) from None
+
+
+def upload_file(
```
Shelnutt2, line 175: Like the `export_file` can we also tie this into the current `create_file` (I like the name upload_file btw) but wanting to see if we can switch automatically on "local upload" vs "load from s3"?

JohnMoutafis (approved):
lgtm

src/tiledb/cloud/files/utils.py
thetorpedodog: This is a good idea for a followup.

thetorpedodog:
Testing round-trip exposed a bug in the file reading process, so this has undergone significant revision. I also want to get #580 in before this.

Shelnutt2 (approved):

src/tiledb/cloud/files/utils.py
Shelnutt2: Let's get this merged then, and can we schedule the followup soon, @antalakas ?

PR #577, opened 2024-06-05
Add sharing and registration to tiledb.cloud.asset

sgillies:
Starting with share() and tests.

### DIFF
--- a/src/tiledb/cloud/asset.py
+++ b/src/tiledb/cloud/asset.py
@@ -12,10 +12,13 @@ from .rest_api.models import GroupInfo  # type: ignore
 
 
 def delete(uri: str, recursive: bool = False) -> None:
-    """Deregister the asset and remove its physical groups and arrays from storage.
+    """Deregister an asset and remove its objects from storage.
 
-    :param str uri: tiledb URI of the asset.
-    :return: None.
+    :param uri: tiledb URI of the asset.
+    :type uri: str
+    :param recursive: if True, contents of an asset will be recursively
+        deleted. Default: False.
+    :type recursive: bool
     """
     delete_map: Mapping[str, Callable] = {
         "array": array.delete_array,
@@ -29,7 +32,8 @@ def delete(uri: str, recursive: bool = False) -> None:
 def info(uri: str) -> Union[ArrayInfo, GroupInfo]:
     """Retrieve information about an asset.
 
-    :param str uri: tiledb URI of the asset.
+    :param uri: tiledb URI of the asset.
+    :type uri: str
     :return: ArrayInfo or GroupInfo.
     """
     # Note: the URI can be either of the two forms, yes?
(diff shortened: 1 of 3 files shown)


Shelnutt2 (approved):

Shelnutt2 (approved):

src/tiledb/cloud/asset.py
```diff
@@ -11,11 +11,90 @@
...
+    'your-bucket' with a prefix 'prefix' to your own namespace:
+
+    >>> from tiledb.cloud import asset
+    >>> asset.register(
+    ...     "s3://your-bucket/prefix",
+    ...     "array",
+    ...     name="new-array",
+    ... )
+    """
+    if type == "array":
```
sgillies, line 53: Resorted to `if/else` in this case.

src/tiledb/cloud/asset.py
```diff
@@ -38,3 +118,109 @@ def info(uri: str) -> Union[ArrayInfo, GroupInfo]:
...
+        "array": array.list_shared_with,
+        "group": groups.list_shared_with,
+    }
+    asset_type: str = tiledb.object_type(uri, ctx=tiledb.cloud.Ctx())
+    func = sharing_map[asset_type]
+    return func(uri)
+
+
+def share(
+    uri: str, namespace: str, permissions: Optional[Union[str, list[str]]] = "read"
```
sgillies, line 185: I could be easily talked out of providing a default for permissions.

Shelnutt2: I think default here is good. Additionally at some point we need to coordinate programmatic vs ui. UI uses "roles" which they translate internally to these permissions.

tests/test_asset.py
```diff
@@ -61,3 +73,194 @@ def test_public_group_asset_info():
...
+    "tiledb.cloud.groups.list_shared_with", return_value=[GroupSharing(namespace="bar")]
+)
+def test_asset_list_shared_with_group_dispatch(group_sharing, object_type):
+    """Dispatch to groups.info when URI is a group."""
+    sharing = asset.list_shared_with("g")
+    assert isinstance(sharing[0], GroupSharing)
+    assert sharing[0].namespace == "bar"
+
+
+@pytest.mark.xfail(reason="Server error 610")
```
sgillies, line 124: Oddly there is a server error in this case, but not with the same test of a public group.

sgillies:
@JohnMoutafis how does this look to you? It completes the new asset.py module.

JohnMoutafis (approved):

JohnMoutafis:
@sgillies Is this ready to merge or does it need more work?

sgillies:
@JohnMoutafis this is good to go. I'll merge now.

JohnMoutafis:
Thank you @sgillies

PR #578, opened 2024-06-05
Modify upload_wheel to use tiledb:// URIs

gspowley:
Modify upload_wheel to use tiledb://...s3://... URIs to create, register, and update the wheel filestore. This avoids the need for s3 credentials.

Thanks to @spencerseale for the suggestion and prototyping.

### DIFF
--- a/src/tiledb/cloud/utilities/wheel.py
+++ b/src/tiledb/cloud/utilities/wheel.py
@@ -1,11 +1,19 @@
+"""Install and access Python packages via wheel path or from PyPI."""
+
 import os
 import subprocess
 import sys
 import tempfile
 from site import USER_SITE
 from typing import Any, Mapping, Optional
+from dataclasses import dataclass
 
 import tiledb
+
+from tiledb.cloud.utilities import get_logger
+
+logger = get_logger()
+logger.propagate = False
 
 
 def upload_wheel(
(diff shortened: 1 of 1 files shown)


src/tiledb/cloud/utilities/wheel.py
```diff
@@ -37,62 +35,27 @@ def upload_wheel(
...
 
     with tiledb.scope_ctx(config):
         # Create the array if it doesn't exist
-        object_type = tiledb.object_type(array_uri)
+        object_type = tiledb.object_type(dest_uri)
         if object_type is None:
-            print(f"Creating wheel filestore - '{array_uri}'")
-            tiledb.Array.create(array_uri, tiledb.ArraySchema.from_file(wheel_path))
+            tiledb.Array.create(dest_uri, tiledb.ArraySchema.from_file(wheel_path))
+            print(f"Created filestore at '{dest_uri}'")
```
spencerseale, line 52: Do we want to use a logger instead of print? For task graph logger inheritance purposes

src/tiledb/cloud/utilities/wheel.py
```diff
@@ -37,62 +35,27 @@ def upload_wheel(
...
         elif object_type == "group":
```
spencerseale, line 53: Won't `object_type` be an `array` if the wheel is already registered since the wheel is registered as an array?

spencerseale (approved):

spencerseale:
Thanks George for adding this in!

gspowley (commented):
LGTM! I can't approve, since I opened the PR.

PR #579, opened 2024-06-06
Automatically update API docs on release

JohnMoutafis:
- Uses [Quarto](https://quarto.org/)/[quartodoc](https://quarto.org/docs/get-started/) to build documentation instead of Sphinx.
- Adds a Workflow to automatically build and upload the API docs on every `push` to [main](https://github.com/TileDB-Inc/TileDB-Cloud-Py/tree/main).


ihnorton (approved):
Great work! Based on the github pages rendering, this looks very good.

(I was not able to render completely locally yet, but I haven't spent much time debugging - lots of warnings)

JohnMoutafis:
@ihnorton those warnings are for method arguments that are not referred in the method's docstring.

spencerseale:
I like the updated theme, thanks for doing this!

PR #581, opened 2024-06-06
Add `region` kw argument to `login`.

NullHypothesis:
As mentioned in [this story](https://app.shortcut.com/tiledb-inc/story/48037/design-improve-support-for-running-task-graphs-in-specific-regions), there are several ways to allow users to specify the region:

1. Modify `tiledb.cloud.client.config.config.host` directly, which is the status quo.
2. Add a `region` keyword argument to classes like `DAG`, `Delayed`, etc.
3. Specify the region as part of logging in (this PR implements a PoC).

The first option is confusing to customers, so we need a better approach. I've been playing with the second approach and find it confusing too: It's easy to forget to pass the `region` argument. Besides, if a user wants to run a task graph in region *A*, odds are that they want all their code to run in region *A*. The third approach strikes me as the safest and least confusing but it feels odd having to specify the region as part of the login function, like so:
```python
tiledb.cloud.login(..., region=tiledb.cloud.AwsRegion.AP_SOUTHEAST_1)
```
Or perhaps we should instead expose a new function, like `tiledb.cloud.set_region()`? Any thoughts on this?

### DIFF
--- a/src/tiledb/cloud/client.py
+++ b/src/tiledb/cloud/client.py
@@ -78,7 +78,8 @@ def login(
     :param token: api token for login
     :param username: username for login
     :param password: password for login
-    :param host: host to login to
+    :param host: host to login to. the tiledb.cloud.regions module contains
+        region-specific host constants.
     :param verify_ssl: Enable strict SSL verification
     :param no_session: don't create a session token on login,
         store instead username/password
--- a/src/tiledb/cloud/region/__init__.py
+++ b/src/tiledb/cloud/region/__init__.py
@@ -1,0 +1,19 @@
+class AWS:
+    """
+    The AWS cloud regions that are currently supported.
+
+    US_EAST_1 = North America, United States (Virginia)
+    US_WEST_1 = North America, United States (California)
+    EU_WEST_1 = Europe, Ireland
+    EU_WEST_2 = Europe, London
+    AP_SOUTHEAST_1 = Asia, Singapore
+    """
+
+    US_EAST_1 = "https://us-east-1.aws.api.tiledb.com"
+    US_WEST_1 = "https://us-west-1.aws.api.tiledb.com"
+    EU_WEST_1 = "https://eu-west-1.aws.api.tiledb.com"
+    EU_WEST_2 = "https://eu-west-2.aws.api.tiledb.com"
+    AP_SOUTHEAST_1 = "https://ap-southeast-1.aws.api.tiledb.com"
+
+
+__all__ = ("AWS",)


src/tiledb/cloud/client.py
```diff
@@ -83,10 +84,20 @@ def login(
...
     """
     if host is None:
         host = config.default_host
+    if region is not None:
+        # If a region is provided, modify the host.
+        orig_url = urllib3.util.parse_url(host)
+        region_url = urllib3.util.Url(
+            scheme=orig_url.scheme,
+            # TODO: Validate region.
+            host=region + ".aws." + orig_url.host,
```
Shelnutt2, line 98: Should we make the region be `<region>.<cloud_provider>` so its future proofed?

NullHypothesis: Yes, good point. I turned the region into an enum, which also helps with validating the given region.

thetorpedodog: We don’t want to make this AWS-specific; since this is just going to replace the `host` parameter in the endpoint, would it make sense to have a set of constants that the user can pass in?

```python

AWS_US_EAST_1 = "https://us-east-1.aws.api.tiledb.com"
# etc.

# Or namespace it with classes:
class AWS:
  US_EAST_1 = "https://us-east-1.aws.api.tiledb.com"

class OtherCloud:
  SOME_REGION: "https://some-region.other.api.tiledb.com"
```

NullHypothesis: My thinking with the enum was that we simply add another one as we support new cloud providers. That said, if y'all prefer constants, then that's fine with me.

Also, in de2950f, I revised the code to expose a separate function instead, `tiledb.cloud.login_to_region()`. This feels more natural to use.

src/tiledb/cloud/client.py
```diff
@@ -63,6 +63,36 @@ def Ctx(config=None):
...
+
+
+def login_to_region(region: AWS, **kwargs):
+    """
+    Login to cloud service, using the given region
+
+    :param region: the cloud region to run code in
+    :return:
+    """
+    if not region in AWS:
```
gsarantid, line 91: I guess this should be `if region not in AWS` instead?

src/tiledb/cloud/client.py
```diff
@@ -63,6 +63,36 @@ def Ctx(config=None):
...
+    The AWS cloud regions that are currently supported.
+
+    US_EAST_1 = North America, United States (Virginia)
+    US_WEST_1 = North America, United States (California)
+    EU_WEST_1 = Europe, Ireland
+    EU_WEST_2 = Europe, London
+    AP_SOUTHEAST_1 = Asia, Singapore
+    """
+
+    US_EAST_1 = "https://us-east-1.aws.api.tiledb.com"
```
vlasisPit, line 77: Maybe we can expose a method to the clients to just print all the available regions?

vlasisPit (approved):

src/tiledb/cloud/client.py
NullHypothesis: They're semantically identical but your suggestions is more readable, so I addressed this in 5af2601.

src/tiledb/cloud/client.py
```diff
@@ -63,6 +63,36 @@ def Ctx(config=None):
     return tiledb.Ctx(Config(config))
 
 
+class AWS(enum.Enum):
```
thetorpedodog, line 66: I would call this AWSRegion or namespace it inside a Region class (so it shows up as `Region.AWS.whatever`)

src/tiledb/cloud/client.py
```diff
@@ -63,6 +63,36 @@ def Ctx(config=None):
...
+
+def login_to_region(region: AWS, **kwargs):
+    """
+    Login to cloud service, using the given region
+
+    :param region: the cloud region to run code in
+    :return:
+    """
+    if region not in AWS:
+        raise Exception("Cloud region is invalid")
```
thetorpedodog, line 92:
```suggestion
        raise ValueError(f"region {region} is invalid")
```

thetorpedodog (changes requested):

src/tiledb/cloud/client.py
NullHypothesis: Fixed in 25bccf9.

src/tiledb/cloud/client.py
NullHypothesis: Fixed in 445c41d.

src/tiledb/cloud/client.py
NullHypothesis: My expectation was that users turn to the docs to see what regions are supported. That said, if you see value in such a function then I'm happy to add it.

thetorpedodog (commented):
I’m still skeptical of this being its own function which takes enums rather than just a set of string constants that you can pass to `host`. Among other things, as we add new regions, existing installs would have a divide where existing regions can use the region type and the `login_to_region` function, but new regions would use `login` with `host=`.

In short, to me,

```python
tiledb.cloud.login(host=tiledb.cloud.Region.AWS.US_EAST_1)
```

(or maybe make `region` into a module, `tiledb.cloud.regions`)

seems like a perfectly reasonable thing to write, and doesn’t require any new functions.

NullHypothesis:
> In short, to me,
> 
> ```python
> tiledb.cloud.login(host=tiledb.cloud.Region.AWS.US_EAST_1)
> ```
> 
> (or maybe make `region` into a module, `tiledb.cloud.regions`)
> 
> seems like a perfectly reasonable thing to write, and doesn’t require any new functions.

See e5401a6.

src/tiledb/cloud/region/region.py
thetorpedodog, line 1: maybe put this directly into `region/__init__.py`?

thetorpedodog (approved):
one suggestion but looks good overall.

PR #582, opened 2024-06-07
File ingestion and indexing chained graph.

JohnMoutafis:
Implements a method that chains file ingestion with file indexing graphs.

- Spawns an ingestion graph that is awaited by the indexing graph.
- Utilizes the existing ingestion and indexing methods.
- Has the same defaults that both the respective methods provide.

### DIFF
--- a/src/tiledb/cloud/files/ingestion.py
+++ b/src/tiledb/cloud/files/ingestion.py
@@ -1,5 +1,5 @@
 import os
-from typing import Any, List, Mapping, Optional, Sequence, Union
+from typing import Any, Dict, List, Mapping, Optional, Sequence, Union
 
 import tiledb
 import tiledb.cloud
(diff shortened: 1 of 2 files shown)


Shelnutt2 (approved):

src/tiledb/cloud/files/ingestion.py
```diff
@@ -281,8 +284,249 @@ def ingest_files(
...
+        partial_index_resources=partial_index_resources,
+    )
+    indexing.depends_on(ingestion)
+
+    # Start the ingestion and indexing process
+    run_dag(graph, wait=wait, debug=verbose)
+    return str(graph.server_graph_uuid)
+
+
+ingest_and_index = as_batch(ingest_and_index)
```
Shelnutt2, line 532: Do we need to use different names to expose both the raw function and the `as_batch` version?

JohnMoutafis: I am doing this to follow the VCF's implementation according to ingestion.
But at some point we need to refactor the ingestion methods.
I am thinking something along the lines of Factory for ingestors and Strategy for ingestor selection

NikolaosPapailiou (approved):
LGTM

PR #584, opened 2024-06-07
Fix soma collection mapper viz

aaronwolen:
Currently, `tiledb.cloud.soma.build_collection_mapper_workflow_graph()` [uses experiment URIs to name task graph nodes](https://github.com/TileDB-Inc/TileDB-Cloud-Py/blob/main/src/tiledb/cloud/soma/mapper.py#L278). This was a reasonable choice but it prevents visualizing the graph because [pydot nodes can't contain colons](https://github.com/pydot/pydot/issues/258).

This PR uses experiment names to name the nodes so we can generate the task graph visualization.

Screenshot 2024-06-07 at 12 57 04 PM

### DIFF
--- a/src/tiledb/cloud/soma/mapper.py
+++ b/src/tiledb/cloud/soma/mapper.py
@@ -256,7 +256,7 @@ def build_collection_mapper_workflow_graph(
 
     node_outputs = {}
 
-    for _, soma_experiment_uri in soma_experiment_uris.items():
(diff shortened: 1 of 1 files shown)


johnkerl (approved):

thetorpedodog (approved):
Looks good, though this is something we really should be fixing in the visualization code in the long term. (Even more really, the pydot library itself shouldn’t allow GraphViz injection, but that’s beyond the scope of TileDB.)

PR #591, opened 2024-06-19
UDF: Add deregister method.

JohnMoutafis:
- Adds deregister method.
-  Also refactors delete method to work with URI.

### DIFF
--- a/src/tiledb/cloud/udf.py
+++ b/src/tiledb/cloud/udf.py
@@ -551,39 +551,31 @@ Delete a registered udf
 """
 
 
-def delete(name: str, namespace: str, async_req: bool = False) -> None:
-    """
-    DEPRECATION WARNING: This method will be replaced by delete_v2
-                         as delete from version 0.12.17 onwards
-    """
-    warning(
-        DeprecationWarning(
-            "This method will be replaced by the `delete_v2` method,",
-            "which will be renamed to `delete` from version 0.12.17.",
-        ),
-    )
-    delete_v2(name=name, namespace=namespace, async_req=async_req)
-
-
-def delete_v2(
-    uri: str = None, *, namespace: str = None, name: str = None, async_req: bool = False
+def delete(
+    uri: str, namespace: Optional[str] = None, *, async_req: bool = False
 ) -> None:
     """
     Deletes a registered udf
 
     :param uri: TileDB URI of the udf, defaults to None.
-    :param name: name of udf, defaults to None.
     :param namespace: namespace the udf belongs to, defaults to None.
+        DEPRECATION WARNING: Will be deprecate from version 0.12.17
     :param async_req: Return future instead of results for async support
     :return: deleted udf details
     """
-    if uri is None and (namespace is None and name is None):
-        raise RuntimeError(
-            "uri or namespace and name must be specified to delete a UDF"
-        )
-
-    if uri is not None:
-        (namespace, name) = utils.split_uri(uri)
+    try:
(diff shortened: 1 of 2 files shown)


src/tiledb/cloud/udf.py
```diff
@@ -550,14 +550,26 @@ def unshare(name=None, namespace=None, async_req=False):
 """
 
 
-def delete(name, namespace, async_req=False):
+def delete(
```
Shelnutt2, line 553: I believe this order change breaks the compatibility of the API? Users that are passing by position today would have breakage?

JohnMoutafis: I made the change to start keeping a more uniform style on those methods going forward.

What I can do is rename this method to `delete_new` and reinstate the previous `delete` method to work as expected with a deprecation warning.

src/tiledb/cloud/udf.py
```diff
@@ -550,14 +551,40 @@ def unshare(name=None, namespace=None, async_req=False):
...
+    DEPRECATION WARNING: This method will be replaced by delete_v2
+                         as delete from version 0.12.17 onwards
+    """
+    warning(
+        DeprecationWarning(
+            "This method will be replaced by the `delete_v2` method,",
+            "which will be renamed to `delete` from version 0.12.17.",
+        ),
+    )
+    delete_v2(name=name, namespace=namespace, async_req=async_req)
```
JohnMoutafis, line 565: @Shelnutt2 do you agree with that approach?

src/tiledb/cloud/udf.py
Shelnutt2: How about we add the uri parameter, to this function, deprecate the existing parameters and warn users. Then we can a avoid the `v2` function name, support compatibility for now and give us a clean future API?

JohnMoutafis: @ihnorton both deprecation approaches have their pros and cons.
We will highly value your input on which method to follow going forward (many methods will eventually be refactored)

ihnorton: We could make the change but fall back to the existing (interpret as `name`) behavior if the argument doesn't start with `tiledb://` -- and print a deprecation warning.

JohnMoutafis: According to @ihnorton suggestion, I modified the method as shown here.
The `namespace` argument will be available for backwards compatibility until version `0.12.17` and the method will throw a deprecation warning if the `uri` is not valid.
FYI @Shelnutt2

src/tiledb/cloud/udf.py
```diff
@@ -550,14 +551,32 @@ def unshare(name=None, namespace=None, async_req=False):
...
-    :param async_req: return future instead of results for async support
+
+    :param uri: TileDB URI of the udf, defaults to None.
+    :param namespace: namespace the udf belongs to, defaults to None.
+        DEPRECATION WARNING: Will be deprecate from version 0.12.17
+    :param async_req: Return future instead of results for async support
     :return: deleted udf details
     """
+    try:
+        namespace, name = utils.split_uri(uri)
```
ihnorton, line 567: Does this work correctly for `tiledb://<namespace>/s3://...` URIs?

ihnorton (approved):
One question but otherwise LGTM.

src/tiledb/cloud/udf.py
JohnMoutafis: Sort of.
- `namespace` will be correct
- `name` will be the whole `s3://...` path

PR #593, opened 2024-07-02
[WIP] CI VFS access

JohnMoutafis:
[WIP] Not ready for review


Shelnutt2 (changes requested):
We must only use gitlab ci for access S3 for testing. Please follow the same strategy we used for the core library testing. We do not want github accessing s3. Please revoke permissions from the roles.

cc: @antalakas

antalakas:
> We must only use gitlab ci for access S3 for testing. Please follow the same strategy we used for the core library testing. We do not want github accessing s3. Please revoke permissions from the roles.
> 
> cc: @antalakas

We will follow the same strategy we used for core testing

PR #596, opened 2024-07-10
Performance improvement for SOMA collection-mapper UDF

johnkerl:
Per @aaronwolen 

See also https://github.com/single-cell-data/TileDB-SOMA/issues/2787


johnkerl:
Closing as dup of #597 by @aaronwolen

PR #598, opened 2024-07-10
Add verbose argument to collection mapper

aaronwolen:
This adds a `verbose` argument to the SOMA experiment collection mapper.

Like other functions in the package, verbose=False limits logs to INFO level. To decrease the amount of logging shown by default I changed the node type logs to DEBUG.

I also addressed [sc-4490] to key results by soma experiment names, instead of URIs, which will be more user friendly.


johnkerl (approved):
🚢

PR #601, opened 2024-07-11
Enhance tiledb.cloud.dag module typing and documentation

spencerseale:
Add class doc str to `Node` and `DAG` so they're built into API ref: https://tiledb-inc.github.io/TileDB-Cloud-Py/reference/dag.dag.html

Original slack: https://tiledb.slack.com/archives/CB0NNLS6B/p1720713156703479

### DIFF
--- a/src/tiledb/cloud/dag/dag.py
+++ b/src/tiledb/cloud/dag/dag.py
@@ -95,13 +95,25 @@ class ParentFailedError(futures.CancelledError):
 class Node(futures.FutureLike[_T]):
     """Representation of a function to run in a DAG.
 
-    :param id: UUID for instance.
-    :param dag: Dag this node is associated with.
-    :param mode: Mode the node is to run in.
-    :param parents: Parent Node instances.
-    :param children: Children Node instances.
-    :param args: position args for Node.
-    :param kwargs: key-word args for Node.
+    :param func: Function to run as UDF task.
+    :param *args: Positional arguments to pass to UDF.
+    :param name: Human-readable name of Node task.
+    :param dag: DAG this node is associated with.
+    :param mode: Mode the Node is to run in.
+    :param expand_node_output: Node to expand processes upon.
+    :param _download_results: An optional boolean to override default
+        result-downloading behavior. If True, will always download the
+        results of the function immediately upon completion.
+        If False, will not download the results of the function immediately,
+        but will be downloaded when ``.result()`` is called.
+    :param _internal_prewrapped_func: For internal use only. A function that returns.
+        something that is already a Result, which does not require wrapping.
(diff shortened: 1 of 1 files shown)


JohnMoutafis:
@spencerseale Have you tested with quartodoc locally how they will appear in the docs?

spencerseale:
`DAG.submit` is not getting included in the doc build because it is defined via `submit = submit_udf`. 

For users, this may be confusing as many of our tutorials reference `submit`, not `submit_udf`. 

Can we remove this redef and just rename the original `submit_udf` to `submit`? @thetorpedodog @JohnMoutafis @ihnorton

src/tiledb/cloud/dag/dag.py
```diff
@@ -103,32 +127,12 @@ def __init__(
...
-        :param _accepts_stored_params: For internal use only.
-            Applies only when ``_prewrapped_func`` is used.
-            ``True`` if ``_prewrapped_func`` can accept stored parameters.
-            ``False`` if it cannot, and all parameters must be serialized.
-        :param kwargs: dictionary for keyword arguments
-        """
+        **kwargs: Any,
+    ) -> None:
         self.id = uuid.uuid4()
+        """UUID for Node instance."""
```
sgillies, line 133: @spencerseale what would you think about using Python comments instead of unassigned strings for this PR?

```suggestion
        # UUID for Node instance.
```

I actually think most of the comments you've added are pretty clear from the code and can be omitted. They won't surface in documentation.

sgillies (changes requested):
@spencerseale thanks for taking this on! Good docstrings are great for our users.

I've made a suggestion for general comment style. I think that many of these comments are clear from the code and can be omitted.

Many type hints have been added. It would be good to update the title of the PR to reflect that so that reviewers understand the intent as they begin.

src/tiledb/cloud/dag/dag.py
spencerseale: These strings are just meant to materialize the attribute definitions in the docs rather than for someone looking through the code. I definitely agree on not overdoing the comments! My main intent was to get users all information about these classes into the api ref, where currently there is nothing: https://tiledb-inc.github.io/TileDB-Cloud-Py/reference/dag.dag.html

sgillies: @spencerseale Great! I had forgotten that attribute docstrings are a thing. In https://discuss.python.org/t/revisiting-attribute-docstrings/36413 I see that they're pretty widely used and appreciated. I'll amend my review.

sgillies (approved):
My preview review was made from a place of ignorance regarding attribute docstrings. I'm in favor of them.

JohnMoutafis (approved):

PR #605, opened 2024-07-16
Convert boolean recursive params to "true" or "false"

sgillies:
The root of the problem is that our API server ignores `recursive="True"` and only expects `recursive="true"`.

I'm converting values of `recursive` before the lowest level, autogenerated client code is called.


sgillies:
Ready for review @JohnMoutafis @Shelnutt2 .

Shelnutt2 (approved):

PR #606, opened 2024-07-17
Refactor `groups.deregister` to pass `recursive` flag to REST.

JohnMoutafis:
Refactors `group.deregister` method to pass the `recursive` flag to the REST server instead of handling recursive deregistering from the cloud-py side


sgillies (approved):
Brilliant @JohnMoutafis.

vlasisPit (approved):

PR #607, opened 2024-07-17
Support double configuration and vfs pushdown in bioimg ingestor

ktsitsi:
This PR:

- Modifies the current ingestion UDF to remove any VFS calls and contexts for passing configs.
- Introduces a new keyword argument for passing a different configuration for the destination's access path

Important:

- This PR does not break the API and is backwards - compatible
- Requires the release of `tiledb-bioimg==0.2.14`, which in its turn requires the release of `tiledb-py==0.30.3`. 

After its merge/release it will allow a user to define separate `tiledb.configurations/ctx` between the source's and destination's access path. However the `ACN` passed in the ingestion batch job should provide access to both of these.


JohnMoutafis (approved):
LGTM. Let's wait for the other components to be in place before we merge?

PR #614, opened 2024-07-24
update wheel to rm cached libs

spencerseale:
https://app.shortcut.com/tiledb-inc/story/51618/update-install-wheel-to-override-cache

### DIFF
--- a/src/tiledb/cloud/utilities/wheel.py
+++ b/src/tiledb/cloud/utilities/wheel.py
@@ -1,12 +1,16 @@
-"""Install and access Python packages via wheel path or from PyPI."""
+"""Install and upload Python packages via a wheel path or from PyPI.
+
+Please be aware, this module is experimental as we explore installing
+private Python wheels at runtime.
+"""
 
 import os
 import subprocess
 import sys
 import tempfile
 from dataclasses import dataclass
-from site import USER_SITE
-from typing import Any, Mapping, Optional
+from dataclasses import field
+from typing import Any, List, Mapping, Optional, Sequence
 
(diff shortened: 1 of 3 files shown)


src/tiledb/cloud/utilities/wheel.py
```diff
@@ -180,5 +234,8 @@ def install_wheel(
         print(res.stderr)
 
     # Modify sys.path if the wheel was installed with --user
-    if not installer.in_venv and USER_SITE not in sys.path:
+    from site import USER_SITE
+
+    if not installer.in_venv:
```
spencerseale, line 239: @gspowley I made this last if statement, less stringent since we want to ensure even if the user site is in the sys.path that it sits at index 0

JohnMoutafis (approved):
lgtm

src/tiledb/cloud/utilities/wheel.py
gspowley: That makes sense. The goal was to prevent the path from growing too large when running `install_wheel` multiple times. We could use something like this:

```python
    if not installer.in_venv and sys.path[0] != USER_SITE:
```

It's a rare edge case though. What you have is good.

gspowley (approved):
LGTM!

src/tiledb/cloud/utilities/wheel.py
spencerseale: going to add

spencerseale:
@JohnMoutafis I added one small change, it's ready to merge now

PR #615, opened 2024-07-25
Validate "memory" string in given resources.

NullHypothesis:
It's easy to forget that the memory string requires a unit suffix to specify something other than bytes.  That results in users passing "2", thinking that it's Gigabytes but it's really bytes.

This commit validates the given memory string. If it's a unitless number less than 10,000,000 (i.e., 10 Mb), the code raises an exception.

### DIFF
--- a/src/tiledb/cloud/dag/dag.py
+++ b/src/tiledb/cloud/dag/dag.py
@@ -224,8 +224,7 @@ class Node(futures.FutureLike[_T]):
             # suffix and request "2" instead of "2Gi". Anything less than 10 Mb
             # (written in bytes) was probably a user error.
             if resources_set and "memory" in self._resources:
-                pattern = re.compile("^[0-9]{1,7}$")
-                if pattern.match(self._resources["memory"]) is not None:
+                if re.match(r"^[0-9]{1,7}$", self._resources["memory"]):
                     raise tce.TileDBCloudError(
                         "The `memory` key in `resources` is missing a"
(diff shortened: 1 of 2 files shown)


src/tiledb/cloud/dag/dag.py
```diff
@@ -219,6 +220,16 @@ def _check_resources_and_mode(self) -> None:
                     )
             elif not resources_set:
                 self._resource_class = "standard"
+            # Validate the memory string because it's easy to forget the unit
+            # suffix and request "2" instead of "2Gi". Anything less than 10 Mb
+            # (written in bytes) was probably a user error.
+            if resources_set and "memory" in self._resources:
+                pattern = re.compile("^[0-9]{1,7}$")
+                if pattern.match(self._resources["memory"]) is not None:
```
thetorpedodog, line 228: this can be

```python
if re.match(r"...", self._resources["memory"]):
  raise ...
```

just using the `re.match` function rather than compiling an intermediate expression, and using the fact that any regular expression match is truthy.

thetorpedodog (approved):
minor suggestion, but a good change otherwise

src/tiledb/cloud/dag/dag.py
NullHypothesis: Much nicer, thanks!

NullHypothesis:
Somebody else is going to have to merge this. GitHub tells me that I'm not authorized to merge.

PR #616, opened 2024-07-25
Add feature: Split VCF

spencerseale:
Adding a split feature for distributed split of a multi-sample VCF. A common obstacle occurring prior to VCF ingestion. 

https://app.shortcut.com/tiledb-inc/story/48501/get-the-multisample-vcf-splitter-released-as-a-public-registered-task-graph


src/tiledb/cloud/vcf/split.py
```diff
@@ -0,0 +1,343 @@
...
+
+    graph = split_vcf(
+        vcf_uri="s3://tiledb-spencer/projects/horses/raw/thesis_intersect_pub.decomposed.vcf.gz",
+        output_uri="s3://tiledb-spencer/projects/vcf-isolate-dev/isolated-singles",
+        config={"vfs.s3.region": "us-west-2"},
+        samples=["A1543"],
+        namespace="TileDB-Inc",
+        acn="tiledb-cloud-sandbox-role",
+        compute=True,
+    )
```
gspowley, line 343: Please remove the testing code. It could be added to the manually run pytests, see https://github.com/TileDB-Inc/TileDB-Cloud-Py/blob/main/tests/test_vcf.py.

src/tiledb/cloud/vcf/split.py
```diff
@@ -0,0 +1,343 @@
...
+
+    Args:
+        vcf_uri:
+            S3 path to aggregate VCF.
+        config:
+            TileDB config params.
+
+    Returns:
+        Samples included in VCF.
+    """
```
gspowley, line 22: Does this comment style work with the doc generator? I know we use sphinx style in other places.

src/tiledb/cloud/vcf/split.py
```diff
@@ -0,0 +1,343 @@
...
+    vcf_uri: str,
+    output_uri: str,
+    namespace: str,
+    acn: str,
+    resources: Mapping[str, str] = {"cpu": "2", "memory": "30Gi"},
+    compute: bool = True,
+    verbose: bool = False,
+    samples: Optional[Sequence[str]] = None,
+    retry_count: int = 1,
+    max_workers: int = 100,
```
gspowley, line 207: These params need docstrings.

gspowley (approved):
Looks good, a few nits. Adding a manually run pytest is a good idea.

src/tiledb/cloud/vcf/split.py
spencerseale: Yeah I will! This PR is still draft before ready for approval. i'll add unit tests as well

src/tiledb/cloud/vcf/split.py
spencerseale: Sphinx picks it up nicely, but you're right I should assimilate with other doc style and was just a force of habit

spencerseale:
@gspowley @leipzig reformatted doc str to align with rest of base and confirmed local doc build looks good. Created `tests/vcf` and added unit tests for this module and moved other tests to same module. Ready to for review

src/tiledb/cloud/vcf/split.py
spencerseale: Quarto did not like that formatting, thank you for noticing this, should be fixed now

JohnMoutafis (approved):
lgtm.
Please add a description on the PR and merge it.

leipzig (changes requested):
I think this is great. Let's just make sure you have tests for the edge cases
- splitting a VCF with 1 sample
- splitting a VCF with 0 samples
- splitting a VCF with 100 samples but no variants
These are all "do nothing, but notify" edge cases

spencerseale:
@leipzig been thinking about the edge cases:

splitting a VCF with 1 sample: 
- If the user passes in the single sample, then bcftools will perform split without error and logging that. Doing exactly what was requested. I don't believe we should try to engineer something to ensure they're not confused, but rather ensure we do explicitly whats requested to caller.
- If user doesn't specify samples, the first stage of task graph will log that just one sample exists in VCF and the second stage will split that one sample. It will be clear from first stage the number of samples in VCF.

splitting a VCF with 0 samples:
- If they don't specify samples, will perform first stage where samples are listed and dynamically spawn subsequent stage based on number of values in samples. So no splitting will occur and the first stage would list that no samples were found.
- If they do specify a sample that doesn't exist in the vcf aggregate, I made an edit to promote the bcftools error and that will be returned to logger and look like: `Error: subset called for sample that does not exist in header`.

splitting a VCF with 100 samples but no variants:
- Would just split each sample as it's own node and return headers without variant data. Since we would still have to start a node per sample to get this check, I don't see any ramp in efficiency/cost savings if we first checked for variants. To check for variant data, I would need to then open each of the split VCF files once written to s3, incurring additional billing (although small).

Let me know your comments on these. My opinion is because the logger is verbose enough and i've promoted bcftools to return stderr when it returns a non-zero return code that the tests are sufficient.

spencerseale:
I know you're busy so no rush on this PR, just when you'd like it to be pushed. But do I need to do anything academy-wise for this prior to its release?

leipzig:
> I know you're busy so no rush on this PR, just when you'd like it to be pushed. But do I need to do anything academy-wise for this prior to its release?

should be documented at the end of  life sciences > popgen > tutorials > advanced > Scalable Ingestion

leipzig (approved):

PR #618, opened 2024-07-26
Upload folder implementation.

JohnMoutafis:
Upload a local folder, utilizing the `upload_file` method.

- Respects the uploading folder structure by default and recreates it in the destination.
- Ingests the files into a given Group, or creates a new one form the folder name


src/tiledb/cloud/files/utils.py
```diff
@@ -264,3 +267,134 @@ def _auth_headers(cfg: configuration.Configuration) -> Mapping[str, object]:
...
     return {}
     # No authentication has been provided. Do nothing.
+
+
+def upload_folder(
+    input_uri: str,
+    output_uri: str,
+    *,
+    group_uri: Optional[str] = None,
+    exclude_files: Optional[List[str]] = None,
```
thetorpedodog, line 277: `Optional[Sequence[str]] = ()`

src/tiledb/cloud/files/utils.py
```diff
@@ -264,3 +267,134 @@ def _auth_headers(cfg: configuration.Configuration) -> Mapping[str, object]:
...
+    exclude_files = exclude_files or []
+
+    input_uri = input_uri if input_uri.endswith(os.sep) else input_uri + os.sep
+    base_dir = os.path.dirname(input_uri)
+    base_dir = os.path.basename(base_dir)
+
+    namespace, name = utils.split_uri(output_uri)
+    _, sp, acn = groups._default_ns_path_cred(namespace=namespace)
+
+    storage_path = name if name.startswith("s3://") else sp
```
thetorpedodog, line 316: Why are we treating Amazon URIs specially here?

src/tiledb/cloud/files/utils.py
```diff
@@ -264,3 +267,134 @@ def _auth_headers(cfg: configuration.Configuration) -> Mapping[str, object]:
...
+    storage_path = f"{storage_path.strip('/')}/{base_dir}"
+    logger.debug("Output storage path: %s" % storage_path)
```
thetorpedodog, line 318:
```suggestion
    logger.debug("Output storage path: %s", storage_path)
```

src/tiledb/cloud/files/utils.py
```diff
@@ -264,3 +267,134 @@ def _auth_headers(cfg: configuration.Configuration) -> Mapping[str, object]:
...
+    if not tiledb.object_type(group_uri, ctx=tiledb.cloud.Ctx()) == "group":
+        group_namespace, group_name = utils.split_uri(group_uri)
+        groups.create(
+            name=group_name,
+            namespace=group_namespace,
+            storage_uri=storage_path,
+            credentials_name=access_credentials_name,
+        )
+        group_created = True
+        logger.debug("Group URI: '%s' created" % group_uri)
```
thetorpedodog, line 340: use comma and %r here

src/tiledb/cloud/files/utils.py
```diff
@@ -264,3 +267,134 @@ def _auth_headers(cfg: configuration.Configuration) -> Mapping[str, object]:
...
+        - Input URI: %s
+        - Output URI: %s
+        - Group URI: %s
+            - Created: %s
+        - Excluded Files: %s
+        - Flatten: %s
+        ----------------------------------------------------
+        """
+        % (input_uri, output_uri, group_uri, group_created, exclude_files, flatten)
+    )
```
thetorpedodog, line 355: I would recommend making these individual log lines.

src/tiledb/cloud/files/utils.py
```diff
@@ -264,3 +267,134 @@ def _auth_headers(cfg: configuration.Configuration) -> Mapping[str, object]:
...
+                    access_credentials_name=access_credentials_name,
+                )
+                uploaded += 1
+            except Exception as exc:
+                upload_errors[fname] = str(exc)
+
+    return {
+        "msg": f"Uploaded {uploaded}/{len(input_ls) - dir_count} files",
+        "errors": upload_errors,
+    }
```
thetorpedodog, line 400: maybe make this an attrs class?

src/tiledb/cloud/files/utils.py
```diff
@@ -264,3 +267,134 @@ def _auth_headers(cfg: configuration.Configuration) -> Mapping[str, object]:
...
+            try:
+                upload_file(
+                    input_uri=fname,
+                    output_uri=f"{storage_path}/{out_path}",
+                    filename=filename,
+                    access_credentials_name=access_credentials_name,
+                )
+                uploaded += 1
+            except Exception as exc:
+                upload_errors[fname] = str(exc)
```
thetorpedodog, line 395: why are we turning the exception here into a string? it strips away any useful information that might have been included.

src/tiledb/cloud/files/utils.py
```diff
@@ -264,3 +267,134 @@ def _auth_headers(cfg: configuration.Configuration) -> Mapping[str, object]:
...
+        - Excluded Files: %s
+        - Flatten: %s
+        ----------------------------------------------------
+        """
+        % (input_uri, output_uri, group_uri, group_created, exclude_files, flatten)
+    )
+
+    vfs = tiledb.VFS(config=config)
+    # Create the base dir in the destination, if it does not exist.
+    if not vfs.is_dir(storage_path):
```
Shelnutt2, line 359: We should not call VFS operations here. We should just be creating groups and/or uploading files.

src/tiledb/cloud/files/utils.py
```diff
@@ -264,3 +267,134 @@ def _auth_headers(cfg: configuration.Configuration) -> Mapping[str, object]:
...
+        logger.debug("No group_uri provided. Choosing one...")
+        if not name.startswith("s3://"):
+            group_uri = f"tiledb://{namespace}/{base_dir}"
+        else:
+            group_uri = output_uri
+
+    group_created = False
+    if not tiledb.object_type(group_uri, ctx=tiledb.cloud.Ctx()) == "group":
+        group_namespace, group_name = utils.split_uri(group_uri)
+        groups.create(
```
Shelnutt2, line 333: Let's use TileDB-Py library group functionality here. We also need to open the group and add all the members.

src/tiledb/cloud/files/utils.py
JohnMoutafis: I prefer the result to be serializable.

src/tiledb/cloud/files/utils.py
JohnMoutafis: I prefer the simplicity of dictionaries, especially as return argument.
Plus it is a serializable result.

src/tiledb/cloud/files/utils.py
JohnMoutafis: Although I am not against writing these in individual lines, I believe that in that case the "packed" information format is more readable.

src/tiledb/cloud/files/utils.py
JohnMoutafis: I am now checking if the name contains `://` will that suffice?

src/tiledb/cloud/files/utils.py
JohnMoutafis: Part of the task description is to recreate the folder structure in the destination.

Shelnutt2: This is correct, but it should not be done by creating VFS commands, it should be done by creating a group for each folder location and a file upload at the same place as the original files.

src/tiledb/cloud/files/utils.py
thetorpedodog: In both cases, I think the right solution is to implement a simple wrapper which transforms the result into something that is JSON-serializable, rather than preemptively serializing the information in the return. Doing so just means that anybody who wants to make use of it has to parse it back again rather than using a defined structure.

src/tiledb/cloud/files/utils.py
```diff
@@ -264,3 +270,170 @@ def _auth_headers(cfg: configuration.Configuration) -> Mapping[str, object]:
...
+    uploaded = 0
+    dir_count = 0
+    upload_errors: Dict[str, str] = {}
+    results = UploadFoldersResults(
+        directory=base_dir, report="", errors={}, sub_folders=[]
+    )
+
+    vfs = tiledb.VFS(config=config)
+    # List local folder
+    input_ls: List[str] = vfs.ls(input_uri)
```
Shelnutt2, line 390: There is also a `recursive=True` flag available, I think this might be needed to list the whole hierarchy below?

Shelnutt2: @JohnMoutafis can you check the performance of the current implementation?

JohnMoutafis: Using the following snippet:

```python
def vfs_ls_recursive(uri):
    vfs = tiledb.VFS()
    files = []
    for fname in vfs.ls(uri):
        files.append(fname)
        if vfs.is_dir(fname):
            files += vfs_ls_recursive(fname)
    return files
```

to time the implementation's manual recursion against `vfs.ls(<folder URI>, recursive=True)` and `vfs.ls_recursive` on a local folder containing sub-folders, got the following results:
- `vfs_ls_recursive`: 1.18 ms ± 24.4 µs per loop (mean ± std. dev. of 100 runs, 1,000 loops each)
- `vfs.ls(<folder URI>, recursive=True)`: 1.95 ms ± 121 µs per loop (mean ± std. dev. of 100 runs, 1,000 loops each)
- `vfs.ls_recursive`: 1.96 ms ± 35.4 µs per loop (mean ± std. dev. of 100 runs, 1,000 loops each)

Timings were collected using [IPython's `%timeit`](https://ipython.readthedocs.io/en/stable/interactive/magics.html#magic-timeit) for 100 runs of 1000 loops per run.
The solutions seem comparable.

PR #620, opened 2024-07-28
Speed up recursive find for VCF URIs

gspowley:
Use `VFS.ls_recursive` when recursively searching for VCF files.


Shelnutt2 (approved):

gspowley:
@normanb we should look into using `ls_recursive` for `tiledb.cloud.utilities.find`. It wasn't immediately obvious how to return an iterator without letting the `ls_recursive` call finish.

JohnMoutafis (approved):

PR #621, opened 2024-07-29
add import to init

spencerseale:
This is to correct the code so the docs: https://docs.tiledb.com/cloud/api-reference/utilities#listing-tasks are correct

https://app.shortcut.com/tiledb-inc/story/51947/correctly-import-tasks-function

### DIFF
--- a/src/tiledb/cloud/__init__.py
+++ b/src/tiledb/cloud/__init__.py
@@ -34,9 +34,12 @@ from .notebook import rename_notebook
 from .notebook import upload_notebook_contents
 from .notebook import upload_notebook_from_file
 from .rest_api import models
+from .tasks import fetch_results
+from .tasks import fetch_results_pandas
 from .tasks import last_sql_task
 from .tasks import last_udf_task
 from .tasks import task
+from .tasks import tasks
 from .tiledb_cloud_error import TileDBCloudError
 
 _pickle_compat.patch_cloudpickle()
@@ -86,5 +89,8 @@ __all__ = (
     "last_sql_task",
     "last_udf_task",
     "task",
+    "tasks",
+    "fetch_results",
+    "fetch_results_pandas",
     "TileDBCloudError",
 )
--- a/src/tiledb/cloud/__init__.py
+++ b/src/tiledb/cloud/__init__.py
@@ -36,10 +36,10 @@ from .notebook import upload_notebook_from_file
 from .rest_api import models
 from .tasks import fetch_results
 from .tasks import fetch_results_pandas
+from .tasks import fetch_tasks
 from .tasks import last_sql_task
 from .tasks import last_udf_task
 from .tasks import task
-from .tasks import tasks
 from .tiledb_cloud_error import TileDBCloudError
 
 _pickle_compat.patch_cloudpickle()
@@ -89,7 +89,7 @@ __all__ = (
     "last_sql_task",
     "last_udf_task",
     "task",
-    "tasks",
+    "fetch_tasks",
     "fetch_results",
     "fetch_results_pandas",
     "TileDBCloudError",
--- a/src/tiledb/cloud/tasks.py
+++ b/src/tiledb/cloud/tasks.py
@@ -37,7 +37,7 @@ def task(id, async_req=False):
         raise tiledb_cloud_error.maybe_wrap(exc) from None
 
 
-def tasks(
+def fetch_tasks(
     namespace=None,
     array=None,
     start=None,
@@ -48,7 +48,7 @@ def tasks(
     async_req=False,
 ):
     """
-    Fetch all tasks a user has access too
+    Fetch all tasks a user has access too.
     :param str namespace: optional filter by namespace
     :param str array: optional limit tasks to specific array
     :param datetime start: optional start time for listing of tasks,
--- a/src/tiledb/cloud/tasks.py
+++ b/src/tiledb/cloud/tasks.py
@@ -1,6 +1,6 @@
 import datetime
 import uuid
-from typing import TYPE_CHECKING, Any, Optional
+from typing import TYPE_CHECKING, Any, Dict, Optional, Sequence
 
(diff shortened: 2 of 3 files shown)


spencerseale:
The alternative is to change the docs and not the code, not sure what we'd want to do here. @kounelisagis @nickvigilante

nickvigilante:
@spencerseale I'm generally in favor of changing docs before changing code, unless this particular code change makes the product more usable and less buggy for our end users, in which case we should do it. Docs are easy to change, but unless it's a customer request or an emergency, I'm really only updating GitHub as opposed to GitBook.

kounelisagis:
> @nickvigilante Makes sense, I just see that we import other functions into that same level from that method such as `task`, so assumed we would also want to import `tasks`. Would like to also see what Agis would like to do and can proceed that way.

Spencer's argument seems logical to me. The way it works, as shown in the documentation, feels better for the user. Let's make that change.

kounelisagis (approved):

nickvigilante (approved):
I'll approve then as well. Thanks for this change!

spencerseale:
@kounelisagis the unit tests are alerting of some unfortunate collisions in a function named `tasks` being located in a module named `tiledb.cloud.tasks`. 

We need to either rename the function or the module to make this cleaner.

Renaming the module would probably be harder to track down all the breaks this would cause. If you are still down to allow the tasks function to be imported into `tiledb.cloud`, what should I rename it as and can you think of any other issues this would cause?

JohnMoutafis:
@spencerseale Indeed the test issue is coming from an import collision between the `tasks` module and the `tasks` function that is now explicitly imported as `tiledb.cloud.tasks`.

- Since the `tasks` function just returns a sequence of tasks it would make sense to rename it to `fetch_tasks`, `task_list` or `all_tasks` and leave the module name as is.
- Another option that will solve the issue for now (but it has the potential to reemerge)  would be to:

      from .tasks import tasks as fetch_tasks
      
      ...
      __all__ = (
         ...
         fetch_tasks
         ...
      )
  and update the docs to match (`all_tasks = tiledb.cloud.fetch_tasks()`)

spencerseale:
@JohnMoutafis since we use 'fetch' elsewhere i'll change the function to `fetch_tasks`. Can you think of elsewhere I should look for any breaks? Thank you for your input and I apologize for not including you initially.

spencerseale:
@nickvigilante can we update: https://docs.tiledb.com/cloud/api-reference/utilities#listing-tasks to `tiledb.cloud.fetch_tasks`? idk if we have any other docs for this, that's all I can find..

Shelnutt2:
We need to be careful about backwards API compatibility. This is a breaking change and as we saw with the `files` rename it is likely to directly impact some users. Are there alternatives we can consider to maintain compatibility?

spencerseale:
@Shelnutt2 see latest commit and let me know if that's sufficient. Otherwise we can scrap this PR and update docs, it just feels funny with some functions imported into `tiledb.cloud` from that module and others not.

This passes:

```python
import tiledb.cloud
import datetime

ago = datetime.datetime.utcnow() - datetime.timedelta(days=35)

old = tiledb.cloud.tasks.tasks(namespace="spencerseale", start=ago)
new = tiledb.cloud.fetch_tasks(namespace="spencerseale", start=ago)

assert old == new
```

My proposition is to include this to support backwards compat and change all docs to specify `tiledb.cloud.fetch_results`

spencerseale:
Another thing, one of the examples listed in the [docs](https://docs.tiledb.com/cloud/api-reference/utilities#listing-tasks:~:text=array_tasks%20%3D%20tiledb.cloud.tasks(array%3D%22tiledb%3A//TileDB%2DInc/quickstart_sparse%22)) doesn't work even when I call it with correct module tree: 

`array_tasks = tiledb.cloud.tasks.tasks(array="tiledb://TileDB-Inc/quickstart_sparse")`

```
TileDBCloudError                          Traceback (most recent call last)
Input In [1], in <cell line: 2>()
      1 import tiledb.cloud
----> 2 array_tasks = tiledb.cloud.tasks.tasks(array="tiledb://TileDB-Inc/quickstart_sparse")

File /opt/conda/lib/python3.9/site-packages/tiledb/cloud/tasks.py:108, in tasks(namespace, array, start, end, status, page, per_page, async_req)
    105     return api_instance.tasks_get(**args)
    107 except GenApiException as exc:
--> 108     raise tiledb_cloud_error.maybe_wrap(exc) from None

TileDBCloudError: Namespace not found - Code: 25 - Extra details: {'request_id': '3fc7fc5c-dc27-4f86-8a08-66d20a35ec50'}
```

JohnMoutafis (approved):

Shelnutt2:
> @Shelnutt2 see latest commit and let me know if that's sufficient. Otherwise we can scrap this PR and update docs, it just feels funny with some functions imported into `tiledb.cloud` from that module and others not.
> 
> This passes:
> 
> ```python
> import tiledb.cloud
> import datetime
> 
> ago = datetime.datetime.utcnow() - datetime.timedelta(days=35)
> 
> old = tiledb.cloud.tasks.tasks(namespace="spencerseale", start=ago)
> new = tiledb.cloud.fetch_tasks(namespace="spencerseale", start=ago)
> 
> assert old == new
> ```
> 
> Including this change will ensure backwards compatibility while we change all docs to specify `tiledb.cloud.fetch_results`

Thanks, I think this is good!

spencerseale:
@JohnMoutafis i'll open a shortcut for that and merge this. It's mainly just the `arrays` arg that's not working.

JohnMoutafis:
Thank you @spencerseale

spencerseale:
@nickvigilante I'm merging and we have backwards compatibility maintained. Can you update the originally linked docs so any current reference `tiledb.cloud.tasks` is changed to `tiledb.cloud.fetch_tasks`. Thank you!

PR #624, opened 2024-07-31
More updates for VCF ingestion with tiledb URIs

gspowley:
Follow up to #623.

Skip `variant_stats` and `allele_count` consolidation when ingesting to a `tiledb://` URI.

### DIFF
--- a/src/tiledb/cloud/vcf/ingestion.py
+++ b/src/tiledb/cloud/vcf/ingestion.py
@@ -900,6 +900,10 @@ def consolidate_dataset_udf(
             group = tiledb.Group(dataset_uri)
 
             for member in group:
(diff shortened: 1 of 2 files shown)


src/tiledb/cloud/vcf/ingestion.py
```diff
@@ -900,6 +900,10 @@ def consolidate_dataset_udf(
             group = tiledb.Group(dataset_uri)
 
             for member in group:
+                # Skip non-array members
+                if member.type != tiledb.Array:
```
Shelnutt2, line 904: I think `member.type` is a string? I don't think we can compare to the class?

Shelnutt2 (approved):

PR #626, opened 2024-07-31
Reference Registered UDF to as_batch

spencerseale:
I want to be able to run registered UDFs using `as_batch`

Repro:

```python
batched_ls = as_batch("TileDB-Inc/ls_uri")

batched_ls(
    "s3://tiledb-spencer/projects",
    config={"vfs.s3.region": "us-west-2"},
    namespace="TileDB-Inc",
    acn="tiledb-cloud-sandbox-role",
)
```

results: https://cloud.tiledb.com/taskgraphlogs/TileDB-Inc/c643bc0c-5dfb-479f-ac9a-a1485ba47638

Can you double check my changes closely here? I _think_ we can actually just pop those args we pass into `Node` rather than getting them and then ignoring them downstream.

### ALSO

We have code in `tiledb.cloud.utilities._common`, how do we want to make these available to the api ref: https://tiledb-inc.github.io/TileDB-Cloud-Py/reference/

If we import from module `_common`, I wish we could get those to show up just under the utilities library in the sidebar.

### DIFF
--- a/src/tiledb/cloud/utilities/_common.py
+++ b/src/tiledb/cloud/utilities/_common.py
@@ -372,7 +372,7 @@ def _filter_kwargs(function: Callable, kwargs: Mapping[str, Any]) -> Mapping[str
 _CT = TypeVar("_CT", bound=Callable)
 
 
-def as_batch(func: _CT) -> _CT:
+def as_batch(func: Union[_CT, str]) -> _CT:
     """
     Decorator to run a function as a batch UDF on TileDB Cloud.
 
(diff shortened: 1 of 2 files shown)


src/tiledb/cloud/utilities/_common.py
```diff
@@ -382,9 +382,20 @@ def as_batch(func: _CT) -> _CT:
...
 
-    :param func: function to run
+    :param func: function to run or name of registered UDF
     """
 
+    if isinstance(func, str):
+        remote_udf = True
+        registered_udf = func
+
+        def func():
```
spencerseale, line 392: Is there a better way? Need to get a function that we can wrap, so I just make a dummy one and then overwrite its name attr...

PR #627, opened 2024-08-02
Allow passing URIs for asset-related functions.

thetorpedodog:
We had a mix of functions that accepted a (namespace, name) pair, and functions which accepted a URI. This is annoying. In this change, we update all these functions to also accept URIs.

In most cases, the namespace and name are positional arguments, and often have other positional arguments after them. This means we usually can't just accept a URI in the same positional argument, because that would end up with everything else offset.

Instead, in these cases, we add a new param, `dest_uri`, which a user can pass instead of the namespace/name pair. Support for namespace/name is maintained, and we ensure that both are never passed.

---

It took a little time to figure out what I think is a good API for this. If you have any ideas, I am open to suggestions.

### DIFF
--- a/src/tiledb/cloud/_common/utils.py
+++ b/src/tiledb/cloud/_common/utils.py
@@ -4,9 +4,8 @@ import functools
 import logging
 import sys
 import threading
-import urllib.parse
 from enum import Enum
-from typing import Any, Callable, Optional, Type, TypeVar, Union
+from typing import Any, Callable, Optional, Tuple, Type, TypeVar, Union
 
 import cloudpickle
 import urllib3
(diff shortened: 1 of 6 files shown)


src/tiledb/cloud/_common/utils.py
```diff
@@ -21,18 +20,85 @@
 logger = logging.getLogger("tiledb.cloud")
 
 
-def split_uri(uri):
+def canonicalize_ns_name_uri(
+    **ns_name_uri: Optional[str],
```
JohnMoutafis, line 24: Although this is clever, should we do this more strict and just have the `namespace`, `name` and `uri` there?

JohnMoutafis (commented):
Wasn't the ultimate goal to deprecate the usage of both `namespace & name` or `tiledb://` style URIs and use only the later everywhere?
Is this an intermediate step in that process?

thetorpedodog:
> Wasn't the ultimate goal to deprecate the usage of both `namespace & name` or `tiledb://` style URIs and use only the later everywhere? Is this an intermediate step in that process?

If the goal is to fully remove separated ns/name parameters, this could be an intermediate step in that process.

The problem I am trying to avoid is that of breaking existing code. For a long time, that has been the prime directive, and there’s no real way to do that _without_ breaking existing code. As I said, because in many cases the namespace/name parameters are positional, it will inevitably result in a hard cutover that can’t really be worked around since you have to just remove a positional argument.

JohnMoutafis (approved):
As an intermediate step lgtm!

src/tiledb/cloud/_common/utils.py
thetorpedodog: Whoops! Forgot to respond to this:

I’m not hugely worried about this because it’s internal-only, and while this is a little clever, it makes the callsites extremely straightforward, since the call is always going to look like `param_name=param_name`.

PR #628, opened 2024-08-02
revert scope

spencerseale:
Ensure the scope covers the install call to rectify install from tiledb://


gspowley (approved):

Shelnutt2 (approved):

spencerseale:
I'm guessing there may be something weird with the tests because I have it uploading wheels to my namespace and my bucket. IDK how we usually deal with this for testing. I try to mock normally, but for here I thought it be important to test the wheel is actually uploaded to tiledb cloud.

spencerseale:
I forced manual pass of the `in_venv` I was finding with deeper testing that was install into --user when a conda env was active and would actually want to just install in my conda runtime rather than .local. 

Also the fails are permissions issues but tests passed locally.

I'm trying to upload a wheel to `s3://tiledb-spencer/...` and then register to TileDB-Inc namespace. But looks like actions isn't authorized to do that?

spencerseale:
Removed tests, will add as separate pr when more time to familiarize with injecting secrets into tests. @gspowley if you could look at one more time before merge, I bumped `in_venv` as an arg to just have finer control of where wheel/lib is installed. Also found a case where if a wheel is registered by the original URI was deleted then `tiledb.object_type` will throw a `TileDBError`. 

I'm going out of cell service today, I'm not familiar with timeline of the project this impacts so please feel free to merge or make updates as you wish. i'll check phone when back in service later today. Thank you.

spencerseale:
```python
import sys
from tiledb.cloud.utilities import install_wheel

try:
    import vcf_annotation
except ModuleNotFoundError:
    print("Does not exist")
    install_wheel("tiledb://spencerseale/s3://tiledb-spencer/dev/wheels/vcf_annotation-1.1.0-py3-none-any.whl", no_deps=True, verbose=True)
    import vcf_annotation
    assert "vcf_annotation" in sys.modules
    print("passed")
```

stdout/stderr:
```
Does not exist
Processing /var/folders/r6/3tz8wzn561z444rkv7fv92340000gn/T/tmp9w1rmnwi/vcf_annotation-1.1.0-py3-none-any.whl
Installing collected packages: vcf-annotation
Successfully installed vcf-annotation-1.1.0


passed
```

PR #629, opened 2024-08-07
Unit tests for install_wheel

spencerseale:
Follow-up from recent PR to include unit tests.

I added the `tiledb.cloud.Config` to hit if no config specified as requested by @Shelnutt2


gspowley (commented):
Looks good, thanks!

One thing to look at, this test [failed](https://github.com/TileDB-Inc/TileDB-Cloud-Py/actions/runs/10293198379/job/28489033523#step:11:91) the first time and [passed](https://github.com/TileDB-Inc/TileDB-Cloud-Py/actions/runs/10293198379/job/28489033523#step:11:165) when rerun.

spencerseale:
> Looks good, thanks!
> 
> One thing to look at, this test [failed](https://github.com/TileDB-Inc/TileDB-Cloud-Py/actions/runs/10293198379/job/28489033523#step:11:91) the first time and [passed](https://github.com/TileDB-Inc/TileDB-Cloud-Py/actions/runs/10293198379/job/28489033523#step:11:165) when rerun.

Thanks @gspowley, `ruff` as a githook deleted the `import django` on my side which needs to occur after the install for it to show up in `sys.modules`. I added an ignore tag to that line and updated.

gspowley (approved):

JohnMoutafis:
I merged this PR because we will push a new CLoud-Py version to facilitate our UI.

tests/utilities/test_wheel.py
```diff
@@ -0,0 +1,162 @@
...
+            dest_uri=_FULL_URI,
+            overwrite=False,
+            config=_CONFIG,
+        )
+
+    upload_wheel(
+        wheel_path=_LOCAL_WHEEL,
+        dest_uri=_FULL_URI,
+        overwrite=True,
+        config=_CONFIG,
```
sgillies, line 76: @spencerseale @gspowley @ryan-williams we're observing regular failures of this test in CI. Do we have a race condition in these uploads or something? Can we restructure the test to eliminate the race or collision?

gspowley: Yes, I believe there is a collision when more than one CI runner is running these tests at the same time. We can restructure the test to create unique asset names for each test.

Semi-related, there are 6 `run-tests` [runners](https://github.com/TileDB-Inc/TileDB-Cloud-Py/actions/runs/10709331229/job/29693713400) in GHA, which are now all running the all tests? In the past, I believe each runner ran a subset of the tests to reduce test time. All runners running the same tests would make the collision more likely.

We still need to restructure the test though.

gspowley: #636 modifies `test_wheel.py` to avoid the collision. Tests are currently in flight.

gspowley: Tests are passing in #636 now and the pytest splitting issue is resolved.

spencerseale: @gspowley @sgillies sorry guys, I didn't check my notifications this week. George thank you for taking care of it.

PR #632, opened 2024-08-22
Add an 'official' caller for batch UDFs

spencerseale:
This is more approachable from a docs point of view compared to `as_batch` and accepts both in-memory and references to registered UDFs. 

Because of the interests of users for calling UDFs with custom resource specifications and access credential names, this function acts to satisfy that interest.

### DIFF
--- a/src/tiledb/cloud/dag/dag.py
+++ b/src/tiledb/cloud/dag/dag.py
@@ -17,6 +17,7 @@ from typing import (
     FrozenSet,
     Hashable,
     List,
+    Mapping,
     Optional,
     Sequence,
     Set,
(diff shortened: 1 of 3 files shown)


src/tiledb/cloud/dag/dag.py
```diff
@@ -46,6 +49,8 @@
 from . import visualization as viz
 from .mode import Mode
 
+logger = logging.getLogger(__name__)
```
JohnMoutafis, line 52: For uniformity reasons, I prefer that we use `tiledb.cloud.utilities.get_logger_wrapper` for logging everywhere.

The method also allows for a verbosity level (`verbose=True/False`)  that should be set from the UDF's arguments (as is the case with the `as_batch` method)

spencerseale: @JohnMoutafis I agree, but we have a circular import problem because `run_dag` is in the utilities module.

sgillies: @spencerseale @JohnMoutafis are we still stuck here? Do I understand correctly that we don't have a circular import yet, but *will* when `run_dag()` calls this new function?

If we refactored and moved `get_logger_wrapper()` to, for example, `tiledb.cloud.logging`, that would eliminate the potential circular import, yes? I'm willing to do that work.

spencerseale:
@gspowley @sgillies got kinda buried, but still would like this function for calling batch UDFs in a more "official" manner. Currently we just have the `as_batch` decorator and that won't work for registered UDFs that are not in memory.

spencerseale:
Up for any ideas, including `@taskgraph` and `@udf` decorators that could also just do `udf("registered_udf")`

PR #633, opened 2024-08-23
Optimize VCF ingestion.

teo-tsirpanis:
[SC-53394](https://app.shortcut.com/tiledb-inc/story/53394/vcf-ingestion-optimization-opportunities)

* The `find_index` and `get_record_count` utility functions were updated to accept a TileDB VFS object which enables caching previously obtained temporary credentials.
* Instead of creating two thread pools one after the other, while we could have reused them for both map stages, and go even further by merging them, allowing to reuse the TileDB contexts created per thread.

### DIFF
--- a/src/tiledb/cloud/vcf/ingestion.py
+++ b/src/tiledb/cloud/vcf/ingestion.py
@@ -822,15 +822,14 @@ def ingest_samples_udf(
                             return None
                 return uri
 
-            # Run sort and bgzip on uncompressed VCF files
             with ThreadPool(threads) as pool:
+                # Run sort and bgzip on uncompressed VCF files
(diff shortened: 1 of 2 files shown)


sgillies:
@teo-tsirpanis do we still need this after the merge of #636 ?

teo-tsirpanis:
It remains important to reuse VFS objects (see also first bullet point of linked story), but I realized after opening the PR that we can't do that because TileDB API objects are not thread-safe due to the last error of API calls getting stored in the same context object. Fixing that would require major changes to the C API (not very hard to migrate a single component but challenging to manage migration and compatibility of multiple components).

Feel free to close this for now.

sgillies:
Thanks for the explanation @teo-tsirpanis !

PR #636, opened 2024-09-03
Improve handling of unsorted VCF files

gspowley:
* Improve handling unsorted VCF files, by not relying on the filename extension. 
* Simplify VCF `sort/bgzip/index` by using one `ThreadPool`.

Non-related CI changes:
* Add GHA step to check for valid `.test_durations` and avoid error splitting pytests
* Modify `test_wheel.py` to avoid failures due to collisions between concurrent tests.

### DIFF
--- a/src/tiledb/cloud/vcf/ingestion.py
+++ b/src/tiledb/cloud/vcf/ingestion.py
@@ -787,46 +787,29 @@ def ingest_samples_udf(
                 tmp_space = "."
             tmp_uris = []
 
-            def bgzip_worker(uri: str) -> Optional[str]:
-                """
-                Sort and bgzip a VCF file, storing the result in the tmp space.
-
-                :param uri: URI of the VCF file
-                :return: URI of the bgzipped VCF
-                """
-                if uri.endswith(".vcf"):
-                    with tiledb.scope_ctx(config):
-                        logger.debug("sort and bgzip %r", uri)
-                        try:
-                            uri = sort_and_bgzip(uri, tmp_space=tmp_space)
-                            tmp_uris.append(uri)
-                        except Exception as e:
-                            logger.error("Error bgzipping %r: %s", uri, e)
-                            return None
-                return uri
-
             def create_index_file_worker(uri: str) -> Optional[str]:
                 """
-                Create a VCF index file in the current working directory.
+                Create a VCF index file in the current working directory. If the VCF
+                file cannot be indexed, try to sort and bgzip the VCF file first.
 
                 :param uri: URI of the VCF file
                 """
                 with tiledb.scope_ctx(config):
                     if create_index or not find_index(uri):
-                        logger.debug("indexing %r", uri)
+                        logger.info("indexing %r", uri)
                         try:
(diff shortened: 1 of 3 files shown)


src/tiledb/cloud/vcf/ingestion.py
```diff
@@ -787,47 +787,30 @@ def ingest_samples_udf(
...
                         except Exception as e:
-                            logger.error("Error creating index for %r: %s", uri, e)
-                            return None
+                            logger.warning("%r: %s", uri, e)
+                            logger.info("sort, bgzip, and index %r", uri)
+                            try:
+                                uri = sort_and_bgzip(uri, tmp_space=tmp_space)
+                                tmp_uris.append(uri)
+                                create_index_file(uri)
+                            except Exception as e:
```
sgillies, line 809: @gspowley is it possible to catch only the one (I assume? Could be wrong) specific exception that occurs in `create_index_file()`?

src/tiledb/cloud/vcf/ingestion.py
```diff
@@ -787,47 +787,30 @@ def ingest_samples_udf(
...
+                                logger.error("%r: %s", uri, e)
+                                return None
```
sgillies, line 811: Should we be re-raising here instead of returning?

sgillies (commented):
@gspowley this is related to #633, yes? It looks like it supersedes some of that PR.

I don't understand the change to `test_wheel.py`.

I added two questions inline.

src/tiledb/cloud/vcf/ingestion.py
gspowley: Most likely, I'll have to check.

src/tiledb/cloud/vcf/ingestion.py
gspowley: In this case, we want to log the error and proceed. This allows the ingestion to proceed with the good VCF files.

gspowley:
The change to `test_wheel.py` is intended to avoid issues that happen when 2 different CI jobs are running the tests in `test_wheel.py` at the same time. The current change does not fix the issue, @spencerseale may have some ideas.

gspowley:
> this is related to https://github.com/TileDB-Inc/TileDB-Cloud-Py/pull/633, yes?

I wasn't aware of that PR, but yes the changes do overlap. I recommend we merge this PR first, since it's making function changes to the logic.

tests/utilities/test_wheel.py
```diff
@@ -32,9 +32,14 @@
...
 
 @pytest.fixture(scope="module", autouse=True)
-def tag_wheel():
+def tag_wheel(tmp_path_factory):
     """Make a tagged copy of the wheel before running tests."""
 
+    global _LOCAL_WHEEL
+
+    tmp_path = tmp_path_factory.mktemp("wheel")
+    _LOCAL_WHEEL = os.path.join(tmp_path, os.path.basename(_LOCAL_WHEEL))
```
sgillies, line 41: Great!

PR #637, opened 2024-09-04
`invite_to_organization`: fix `InvitationOrganizationJoinEmail` ctor call

ryan-williams:
[sc-54346]

### DIFF
--- a/src/tiledb/cloud/invites.py
+++ b/src/tiledb/cloud/invites.py
@@ -67,6 +67,7 @@ def invite_to_organization(
     :param organization: name or UUID of organization.
     :param recipients: list of recipient emails/usernames.
     :param role: role assigned to the recipient.
+    :param actions: an optional sequence of actions.
     :return: None
     """
     invitation_api = client.build(rest_api.InvitationApi)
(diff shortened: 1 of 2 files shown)


ryan-williams:
Looks like the `autobuild` GHA has been failing off and on lately, including on `main` currently. Anyone have context?

sgillies (changes requested):
Looks correct to me @ryan-williams ! Can you add a line about `actions` to the docstring and a simple test? I don't think we need much more than a new `test_organization.py` and a test that calls `invite_to_organization()`, mocking tiledb.cloud.client, so that we can be sure that we've got it wired properly and that it stays that way.

For the docstring, it's okay to be vague, like we are for the other parameters. Eventually it could be nice to include a enumeration of acceptable values (here 
https://github.com/TileDB-Inc/TileDB-Cloud-API-Spec/blob/3783c171eb780c47c7c5890a3616870c2bb9e061/openapi-v1.yaml#L330, I think), but we don't need to figure out the best way to do that now.

thetorpedodog:
> Looks like the `autobuild` GHA has been failing off and on lately, including on `main` currently. Anyone have context?

This looks like we’re trying to write things to the same backend path in multiple tests and it’s stepping on each other. We (the main maintainers of this package, not necessarily you) should make sure we set up per-run unique paths for our tests.

sgillies (approved):

PR #639, opened 2024-09-09
Update find to use VFS.ls_recursive

gspowley:
Update `tiledb.cloud.utilities.find` to use the more performant `vfs.ls_recursive`.

`tiledb.cloud.utilities.find` is currently used in `files` and `geospatial` ingestion. `vcf` ingestion will update to use the same version (instead of a custom `vcf` version) after this PR is merged.

### DIFF
--- a/src/tiledb/cloud/utilities/_common.py
+++ b/src/tiledb/cloud/utilities/_common.py
@@ -445,7 +445,7 @@ def find(
     include: Optional[Union[str, Callable]] = None,
     exclude: Optional[Union[str, Callable]] = None,
     max_count: Optional[int] = None,
-) -> Iterator[str]:
+) -> Sequence[str]:
     """Searches a path for files matching the include/exclude pattern using VFS.
 
     :param uri: Input path to search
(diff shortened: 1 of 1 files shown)


src/tiledb/cloud/utilities/_common.py
```diff
@@ -455,44 +455,42 @@ def find(
...
-                            continue
-                    yield f
-
-        for f in list_files(listing):
-            yield f
-
-            current_count += 1
-            if max_count and current_count == max_count:
-                return
+        vfs = tiledb.VFS()
```
sgillies, line 458: @gspowley do we need the config and ctx here as in the original code?

sgillies (approved):
Makes sense to me @gspowley 👍

src/tiledb/cloud/utilities/_common.py
gspowley: I believe `with tiledb.scope_ctx(config):` should cover it, but we can add config and ctx back to minimize changes.

PR #640, opened 2024-09-10
Add `registered_name` to SOMA ingestion.

JohnMoutafis:
Currently SOMA ingestion does one of the following, depending on the `output_uri`:
- If the uri is just a bucket path, it ingests the files into the location but doesn't register a Group (the registration needs to happen manually afterwards)
- If the uri is of the form `tiledb://<namespace>/<bucket-path>` it will ingest the files into the bucket location and will register a Group with the name of the ingested folder/file.

VCF, GeoSpatial and BioImaging ingestors allow the user to set a `register_name` for the ingested files. SOMA ingestor should do too for uniformity.

### DIFF
--- a/src/tiledb/cloud/soma/ingest.py
+++ b/src/tiledb/cloud/soma/ingest.py
@@ -2,18 +2,73 @@ import logging
 import os
 import re
 import warnings
-from typing import ContextManager, Dict, Optional
+from typing import Any, ContextManager, Dict, Mapping, Optional
 from unittest import mock
 
 import tiledb
 from tiledb.cloud import dag
 from tiledb.cloud._common import functions
+from tiledb.cloud._common import utils
 from tiledb.cloud.utilities import as_batch
 from tiledb.cloud.utilities import get_logger_wrapper
 from tiledb.cloud.utilities import run_dag
 
 _DEFAULT_RESOURCES = {"cpu": "8", "memory": "8Gi"}
 """Default resource size; equivalent to a "large" UDF container."""
+
+
+def register_dataset_udf(
+    dataset_uri: str,
+    *,
+    register_name: str,
+    acn: str,
+    namespace: Optional[str] = None,
+    config: Optional[Mapping[str, Any]] = None,
+    verbose: bool = False,
+) -> None:
+    """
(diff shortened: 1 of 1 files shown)


sgillies (commented):
@JohnMoutafis I made a small suggestion inline.

Is the new code covered by existing tests? If not, can you create a unit test or two to assert that the branches are followed as expected? At least for the happy path.

JohnMoutafis:
@sgillies SOMA ingestion is very tightly coupled with VFS at the moment and those tests will fail when run from CI/CD.
We should for sure add and improve existing tests as a follow-up to #595

src/tiledb/cloud/soma/ingest.py
```diff
@@ -303,6 +360,19 @@ def run_ingest_workflow(
             )
         )
 
+    try:
+        ns, dst = utils.split_uri(output_uri)
+        namespace = namespace or ns
+        # Ensure compatibility with "tiledb://<namespace>/<bucket-path>"
+        # style URIs
+        if "://" in dst:
```
sgillies, line 368: @JohnMoutafis could we check the scheme of the output_uri instead? Like `if dst.startswith("tiledb")`? I think that would be more meaningful.

sgillies (approved):
@JohnMoutafis I left a comment previously about an inline suggestion but apparently I failed to finalize it. I have left a new suggestion at line 368.

src/tiledb/cloud/soma/ingest.py
JohnMoutafis: The [`tiledb.cloud._common.utils.split_uri`](https://github.com/TileDB-Inc/TileDB-Cloud-Py/blob/5d4cd8207c34bc91b96253bd2736146088034a1d/src/tiledb/cloud/_common/utils.py#L90) already checks for that and throws a `ValueError` if the URI doesn't start with `tiledb://`.
That case is handled under the assumption that if the exception was thrown, a bucket URI was passed instead.

sgillies: Thanks for the explanation, @JohnMoutafis !

PR #641, opened 2024-09-10
Add submit_taskgraph for a common UX

gspowley:
This PR adds `tiledb.cloud.utilities.submit_taskgraph` as a proposed common function to launch ingestion (and other) taskgraphs.

If `tqdm` is available, a progress timer is displayed.

Output while running:
```
Taskgraph submitted to TileDB - https://cloud.tiledb.com/activity/taskgraphs/...
Running:  5:30 min:sec 
```

Output when complete:
```
Taskgraph submitted to TileDB - https://cloud.tiledb.com/activity/taskgraphs/...
Completed:  10:32 min:sec 
```

### DIFF
--- a/src/tiledb/cloud/utilities/_common.py
+++ b/src/tiledb/cloud/utilities/_common.py
@@ -8,6 +8,7 @@ import subprocess
 import sys
 from concurrent.futures import ThreadPoolExecutor
 from concurrent.futures import wait
+from concurrent.futures._base import TimeoutError
 from fnmatch import fnmatch
 from typing import (
     Any,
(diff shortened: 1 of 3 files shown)


JohnMoutafis (approved):
lgtm

gspowley:
> This looks like a better version of `run_dag` ([source](https://github.com/TileDB-Inc/TileDB-Cloud-Py/blob/main/src/tiledb/cloud/utilities/_common.py#L151)).
> 
> Can we make a note to deprecate that eventually?

Yes, it's a good idea to converge on one convenience function for running taskgraphs. The `debug`/`_print_logs` functionality is useful and can be added to `submit_taskgraph`. I believe the retry mechanism in `run_dag` is legacy and adding a `RetryStrategy` to the `DAG` is the preferred approach.

src/tiledb/cloud/utilities/_common.py
```diff
@@ -195,6 +196,74 @@ def run_dag(
...
+                print("Please install tqdm to display a progress timer.")
+
+            def set_description_str(self, *args, **kwargs):
+                pass
+
+    graph.compute()
+
+    # Display a link to the log in TileDB, if available
+    if graph.server_graph_uuid:
+        print(
```
sgillies, line 230: Let's opt in to printing. I don't think a library function should print by default. I can't think of anything in Python's standard lib that prints except `print()`.

Like `if verbose and graph.server_graph_uuid:`

src/tiledb/cloud/utilities/_common.py
```diff
@@ -195,6 +196,74 @@ def run_dag(
...
+            f"Taskgraph submitted to TileDB - https://cloud.tiledb.com/activity/taskgraphs/{graph.namespace}/{graph.server_graph_uuid}",
+        )
+
+    if wait:
+        # Display the progress bar in stdout, so the output is not red in a notebook
+        pbar = tqdm(
```
sgillies, line 236: Similarly, I think a progress bar should be explicitly opt-in and not conditioned on waiting. I can easily imagine wanting to waiting silently. How about `if verbose: pbar = ...`?

src/tiledb/cloud/utilities/_common.py
```diff
@@ -195,6 +196,74 @@ def run_dag(
...
+        )
+        cancelled = False
+
+        while True:
+            try:
+                graph.wait(update_sec)
+            except TimeoutError:
+                # Still running
+                pbar.set_description_str(str(graph.status))
+                continue
```
sgillies, line 249: I'm confused about what `graph.wait()` does. Is the status set only when it returns?

src/tiledb/cloud/utilities/_common.py
```diff
@@ -195,6 +196,74 @@ def run_dag(
...
+
+            if cancelled:
+                pbar.set_description("Cancelled")
+                return {"status": "Cancelled", "graph_id": str(graph.server_graph_uuid)}
+
+            # The taskgraph is done, break the loop
+            pbar.set_description_str(str(graph.status))
+            break
+
+    return {"status": str(graph.status), "graph_id": str(graph.server_graph_uuid)}
```
sgillies, line 264: It's more natural for a Python function to return a tuple, no? More like standard lib functions is what i mean.

sgillies (commented):
@gspowley I left a few suggestions for how we could make this more Pythonic (for lack of a better term). Run silently by default and return a tuple, basically.

I like tqdm in command line programs but not in library functions so much, but that's my personal taste. I can see how this function and its visual feedback would be good for Notebook users.

PR #642, opened 2024-09-11
Add functions to access new asset listing endpoints.

thetorpedodog:
Also includes a separate change regenerating our API code.


sgillies:
@thetorpedodog do the new functions require regeneration of the API? Can you tease these apart?

thetorpedodog:
> Of course, new endpoints, new YAML. I'm surprised by the number of changed files. Is this with the same version of the API generator as in the committed bash script? Are all of these changes due to updates and improvements in the spec?

Yes; this is _strictly_ the result of running `./update_generated.sh` (you can try it yourself). It’s a pretty typical diff in my experience.

sgillies (approved):
@thetorpedodog 👍

src/tiledb/cloud/asset.py
```diff
@@ -1,19 +1,109 @@
...
+
+
+def list(
+    *,
+    namespace: Optional[str] = None,
+    search: Optional[str] = None,
+    type: Optional[_AssetType] = None,
+    ownership_level: Optional[_OwnershipLevel] = None,
+    depth: Optional[_Depth] = None,
+    expand: Optional[str] = None,
```
Shelnutt2, line 26: Should this be a union of a list or string and we handle the conversion to the csv parameter?

src/tiledb/cloud/asset.py
```diff
@@ -1,19 +1,111 @@
...
+        ownership_level=ownership_level,
+        depth=depth,
+        expand=_canonicalize_csv(expand),
+        page=page,
+        per_page=per_page,
+        order_by=order_by,
+    )
+
+
+def list_public(
```
Shelnutt2, line 68: Does `list_public` not have the `expand` option @snagles ?

snagles: correct

Shelnutt2 (approved):

PR #643, opened 2024-09-13
Fix SOMA ingestion's registration race condition.

JohnMoutafis:
# Issue

The SOMA `ingest` creates a taskgraph called `launcher` (`run_ingest_workflow_udf`) that creates the h5ad ingestor task and calls `compute` on that graph.
That creates a race condition with the ingested SOMA Group registration (initiated from the `ingest` graph) that results in an empty SOMA although every node of the graph resolves without failures, or in another case in a registration failure (manual registration of that Group showed that the ingestion process was resolved correctly).

# Solution

Move the registration inside the `launcher` process.


johnkerl (approved):

sgillies (approved):
Thanks for the analysis and description @JohnMoutafis !

PR #644, opened 2024-09-13
Re-generate generated API to add new r/w/d organization role.

NullHypothesis:
This PR re-generates the generated API to add the new read/write/delete organization role that was introduced in https://github.com/TileDB-Inc/TileDB-Cloud-API-Spec/pull/483. In addition, this PR makes a minor change to make update_generated.sh support macOS.

---

Note that I re-generated the full API. As far as I can tell, the only change we need is [this one](https://github.com/TileDB-Inc/TileDB-Cloud-Py/commit/0e47a0bf5ddb71a4df6f34346052da074c5c5ee2#diff-b1fcd24ce4a39975ea63026f0d188e5f1cc511c7e863e17f591a3989e090bd0dR33-R43).


NullHypothesis:
> @NullHypothesis can you re-regenerate and push to see if the conflicts are resolved?

Thanks for taking a look! I'll pick this back up once the corresponding [REST PR](https://github.com/TileDB-Inc/TileDB-Cloud-REST/pull/4847) is merged.

NullHypothesis:
@sgillies This is ready for review but we should hold off on merging until [the corresponding REST PR is deployed](https://github.com/TileDB-Inc/TileDB-Cloud-REST/pull/4847). I'll un-draft this PR once that's the case.

Regarding the code itself: I noticed that re-generating the API on macOS results in a large number of changes while running it on Linux results in exactly what we'd expect. I therefore taught the script to abort if it's run on a non-Linux platform.

sgillies (approved):
@NullHypothesis 👍

Do you understand why we see extra changes on macOS even with the same generator version?

NullHypothesis:
> Do you understand why we see extra changes on macOS even with the same generator version?

I took another look and realized that this was (mostly) my own fault. I think I didn't properly run the pre-commit hook. That said, one still needs to use GNU sed instead of the default sed on macOS for this to work, which should be fixed in 8631bacd138d35d4a6160310a94cd3d150082251. After that, the output is nearly identical to what I'm getting on Linux, except these VERSION changes we don't really matter:

```
Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	deleted:    src/tiledb/cloud/_common/api_v2/.openapi-generator/VERSION
	deleted:    src/tiledb/cloud/rest_api/.openapi-generator/VERSION

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	src/tiledb/cloud/_common/api_v2/VERSION
	src/tiledb/cloud/rest_api/VERSION
```

NullHypothesis:
@sgillies We should be ready to merge this PR. We deployed the new read/write/delete role in REST earlier today.

sgillies (approved):

PR #645, opened 2024-09-13
Perform HEAD requests to find the correct location to upload.

thetorpedodog:
The service at https://api.tiledb.com which redirects requests to the appropriate TileDB Cloud backend doesn't accept large uploads (and even if it did, we don't want to make unnecessarily large requests). However, it *will* respond with redirect information to a HTTP HEAD request.

This change starts uploads by probing the correct destination URL by making HEAD requests before it tries actually sending the data.

---

Tests for this live behind the `--run-bigfiles` test flag. I have manually confirmed that `test_big_file` works; `test_huge_file` is pending a cluster configuration change.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,8 +47,9 @@ repository = "https://github.com/TileDB-Inc/TileDB-Cloud-Py"
 requires = ["setuptools>=42", "wheel", "setuptools_scm>=6"]
 
 [tool.pytest.ini_options]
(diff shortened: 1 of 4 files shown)


sgillies (approved):
Makes sense @thetorpedodog 👍 

How many redirects do we perform? Is it worth bailing out after some number to avoid getting stuck in an unexpected situation?

JohnMoutafis (approved):

PR #648, opened 2024-09-16
xfail 5 tests because of 2 upstream issues

sgillies:
1. Our retry script isn't working on these particular failures.
2. Instead of increasing the number of retries over time, I want to mark that we have upstream issues (or invalid usage) and then use passing of the marked tests as acceptance criteria for fixes.

### DIFF
--- a/.github/workflows/tiledb-cloud-py.yaml
+++ b/.github/workflows/tiledb-cloud-py.yaml
@@ -151,13 +151,13 @@ jobs:
             --store-durations \
             --splitting-algorithm least_duration \
             --clean-durations \
-            -r xX \
+            -r fExX \
           || .github/scripts/retry 2 \
             pytest -sv \
             --last-failed \
             --tb short \
             --color yes \
-            -r fxX
+            -r fExX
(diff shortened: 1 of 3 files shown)


.github/workflows/tiledb-cloud-py.yaml
```diff
@@ -151,11 +151,13 @@ jobs:
...
             --splitting-algorithm least_duration \
             --clean-durations \
+            -r xX \
           || .github/scripts/retry 2 \
             pytest -sv \
             --last-failed \
             --tb short \
-            --color yes
+            --color yes \
+            -r fxX
```
sgillies, line 160: To gather xpass/xfail info into a report section.

thetorpedodog: I think we also want `E` in this collection, since the default value for `-r` is `fE`. Other than that this looks good!

thetorpedodog (approved):

JohnMoutafis (approved):
lgtm

PR #649, opened 2024-09-16
Default to venv autodetect in install_wheel

spencerseale:
Reverting to original implementation while enabling user to spec for configurability.


gspowley (approved):
Thanks!

sgillies (approved):

PR #650, opened 2024-09-16
Move our custom logger from _common.py to logging.py

sgillies:
This unblocks some work on DAG runners. It also deprecates bioimg's own custom logger, replacing it with a shim that warns.


src/tiledb/cloud/bioimg/helpers.py
```diff
@@ -1,31 +1,32 @@
...
 import tiledb
-from tiledb.cloud.utilities import get_logger
 
+# from tiledb.cloud.logging import get_logger
 
-def get_logger_wrapper(
-    verbose: bool = False,
-) -> logging.Logger:
-    """
-    Get a logger instance and log version information.
```
sgillies, line 12: @ktsitsi 👋 what do you think about replacing bioimg's custom logger with the new standard logger in tiledb.cloud.logging?

ktsitsi (approved):
LGTM nothing really changes here. It removes code duplication which was on our radar for a while.

ihnorton (approved):

sgillies:
Thanks @ihnorton @ktsitsi. The wheel upload test failures are disconcerting but unconnected. I'll wrap this up tomorrow.

PR #651, opened 2024-09-17
Only follow a limited number of redirects.

thetorpedodog:
Based on a comment from the previous review, and a realization about how we can use `urllib3`’s `response.url`.

### DIFF
--- a/src/tiledb/cloud/files/utils.py
+++ b/src/tiledb/cloud/files/utils.py
@@ -227,43 +227,44 @@ def upload_file(
         namespace = urllib.parse.quote(namespace)
         name = urllib.parse.quote(name, safe="")
 
-        request_url = f"{config.config.host}/v2/files/{namespace}/{name}/upload?{query}"
+        orig_url = f"{config.config.host}/v2/files/{namespace}/{name}/upload?{query}"
 
         # First, we make HEAD requests to find out if we should be redirected.
-        while True:
-            resp = pool.request(
-                method="HEAD",
-                url=request_url,
-                headers=headers,
-                redirect=False,
-            )
-            if redirect_to := resp.get_redirect_location():
-                # Follow the redirect and try again.
-                request_url = redirect_to
-                continue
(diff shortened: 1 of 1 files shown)


src/tiledb/cloud/files/utils.py
```diff
@@ -227,43 +227,44 @@ def upload_file(
...
-        # we might still get redirected.
+        # We'll let urllib3 follow redirects for this one.
+        probe = pool.request(
+            method="HEAD",
+            url=orig_url,
+            headers=headers,
+        )
+        if not (200 <= probe.status < 300 or probe.status == 405):
+            # Once we reach the end, we should get either 405 Method Not Allowed
+            # or some form of OK (2xx). If not, that's an error.
```
sgillies, line 241: 👍

src/tiledb/cloud/files/utils.py
```diff
@@ -227,43 +227,44 @@ def upload_file(
...
+            raise tce.TileDBCloudError.from_response(probe)
+        final_url = probe.url or orig_url  # Where we were redirected to.
+
+        # While we're probably at the place we want to end up posting to,
+        # we might still get redirected. This time, we need to handle redirects
+        # manually, since we have to rewind the request body each time.
+        redirects_left = 3
         while True:
```
sgillies, line 249: @thetorpedodog your use of the walrus operation suggested to me that we could use more built-in Python features. Maybe this?

```suggestion
        while redirects_left > 0:
            ...
        else:
            raise tce.TileDBCloudError.from_response(resp)
```

sgillies (approved):
I approve, left a suggestion that you may like or not.

src/tiledb/cloud/files/utils.py
thetorpedodog: I thought about doing something like this, but ended up with the structure we have now, because all the dealing with the response needs to happen in a try/finally for _each individual request_, so this was the best structure.

HAVING SAID THAT this made me realize that I think we need to release the connection for the `HEAD` request as well!

PR #653, opened 2024-09-17
Add datetime parsing for executing UDFs at timestamps.

thetorpedodog:
This change adds the initial step for executing UDFs at timestamps. To allow users to specify a timestamp of a registered UDF to use, we add `@` followed by a timestamp to the name of the UDF.

For example, to execute the version of `my/udf` as it was as of 2023-06-17 at 07:16 UTC, they would execute:

    "my/udf@2023-06-17 07:16"

This is unambiguous and avoids the need to add a new parameter, which would be difficult to name and could shadow parameters in user code.

---

Setting this up now separately so that discussion of how we specify old versions isn't caught up in the rest of the implementation details.

### DIFF
--- a/src/tiledb/cloud/udf.py
+++ b/src/tiledb/cloud/udf.py
@@ -185,6 +185,8 @@ def _parse_udf_name_timestamp(
     full_name: str,
 ) -> Tuple[str, Optional[datetime.datetime]]:
     name, at, ts_str = full_name.partition("@")
+    name = name.strip()
+    ts_str = ts_str.strip()
     if not at:
         # This means that "@" was not found in the string,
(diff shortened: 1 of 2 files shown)


sgillies:
@thetorpedodog I think `@` as a delimeter works (it's like what `pip` uses for tags and commits). What would you think about eliminating the blank between date and time (whitespace being a source of bugs) and replacing it with a `T` ala https://datatracker.ietf.org/doc/html/rfc3339#section-5.8?

sgillies (approved):
👍 My only concern about the whitespace is UX: identifiers without whitespace are easy to select, copy, and share.

sgillies (approved):

PR #654, opened 2024-09-23
Adding PNG support and enabling experimental reader in kwargs

ktsitsi:
This PR:

- Enables the `experimental_reader` in the Cloud-py ingestor which was introduced in tiledb-bioimg 0.3.0 and onwards
- Incorporates the new PNG converter for use in cloud ingestion

### DIFF
--- a/src/tiledb/cloud/bioimg/ingestion.py
+++ b/src/tiledb/cloud/bioimg/ingestion.py
@@ -1,5 +1,6 @@
 import logging
 import warnings
+from enum import Enum
 from typing import Any, Iterator, Mapping, Optional, Sequence, Tuple, Union
 
 import tiledb
@@ -17,6 +18,11 @@ DEFAULT_IMG_NAME = "3.9-imaging-dev"
 DEFAULT_DAG_NAME = "bioimg-ingestion"
 _SUPPORTED_EXTENSIONS = (".tiff", ".tif", ".svs", ".ndpi", ".png")
 _SUPPORTED_CONVERTERS = ("tiff", "zarr", "osd", "png")
+
+
+class ReaderType(Enum):
+    PRODUCTION = "production"
+    EXPERIMENTAL = "experimental"
 
 
 def ingest(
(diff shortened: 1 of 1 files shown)


sgillies (commented):
@ktsitsi "experimental_reader" sounds like a feature that might go away in the future or become the new default, at which we'd want to remove the option.

Is there a way that we can support more than one reader without using a flag that will eventually become deprecated? What about `reader=experimental` and `reader=production` (the default)?

ktsitsi:
> @ktsitsi "experimental_reader" sounds like a feature that might go away in the future or become the new default, at which we'd want to remove the option.
> 
> 
> 
> Is there a way that we can support more than one reader without using a flag that will eventually become deprecated? What about `reader=experimental` and `reader=production` (the default)?

Indeed it is the latter the experimental reader will be the new default. I will change it as you recommend and I will assign the selection inside the function so we avoid the depreciation.

sgillies (approved):
Sounds good @ktsitsi !

spencerseale (approved):

src/tiledb/cloud/bioimg/ingestion.py
```diff
@@ -350,6 +356,11 @@ def register_dataset_udf(
 
     # serialize udf arguments
     compressor = kwargs.pop("compressor", None)
+
+    # Get either the new experimental or default reader
+    reader = kwargs.pop("reader", "production")
+    experimental_reader = True if reader == "experimental" else False
```
ihnorton, line 362: Should this be an enum?

ktsitsi: You are right @ihnorton I changed it in my last commit. I just thought that since the old reader is about to be deprecated once the experimental becomes the default one, that I did not want to over-engineer it. But it is cleaner this way thanks!

src/tiledb/cloud/bioimg/ingestion.py
```diff
@@ -358,8 +364,8 @@ def register_dataset_udf(
     compressor = kwargs.pop("compressor", None)
 
     # Get either the new experimental or default reader
-    reader = kwargs.pop("reader", "production")
-    experimental_reader = True if reader == "experimental" else False
+    reader = kwargs.pop("reader", ReaderType.PRODUCTION.value)
```
sgillies, line 367: @ktsitsi how about making `reader` a name keyword argument of `ingest()` instead?

I'm doing a deep dive into our ingestion code right now and the way we have been moving things in and out of `kwargs` obfuscates the code.

sgillies (changes requested):
@ktsitsi I made a suggestion/question inline.

src/tiledb/cloud/bioimg/ingestion.py
ktsitsi: It indeed does however, the reason I did not do it in the first place is to avoid deprecating it later and breaking the API in the next releases.

sgillies (approved):

PR #655, opened 2024-09-24
Improve missing object detection in upload_wheel

gspowley:
`upload_wheel` uses `tiledb.object_type(...)` to check for existing objects. 

* `tiledb-py<=0.31.1` returns `None` (NoneType) if the object doesn't exist. 
* `tiledb-py>=0.32.0` returns `"None"` (a string) if the object doesn't exist. 

This PR updates the `upload_wheel` to handle both return values.

### DIFF
--- a/src/tiledb/cloud/utilities/wheel.py
+++ b/src/tiledb/cloud/utilities/wheel.py
@@ -71,7 +71,8 @@ def upload_wheel(
             )
 
         # Create the array if it doesn't exist
-# `object_type == "None"` is work-around for regression, see https://app.shortcut.com/tiledb-inc/story/55930 for fix
+        # `object_type == "None"` is work-around for regression,
+        # see https://app.shortcut.com/tiledb-inc/story/55930 for fix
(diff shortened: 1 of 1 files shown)


src/tiledb/cloud/utilities/wheel.py
```diff
@@ -71,7 +71,7 @@ def upload_wheel(
             )
 
         # Create the array if it doesn't exist
-        if object_type is None:
+        if object_type is None or object_type == "None":
```
ihnorton, line 74:
```suggestion
# `object_type == "None"` is work-around for regression, see https://app.shortcut.com/tiledb-inc/story/55930 for cix
        if object_type is None or object_type == "None":
```

ihnorton (approved):

spencerseale:
Guessing we'll follow-up with tiledb-py to revert that to NoneType? Thanks @gspowley!

spencerseale (approved):

sgillies (approved):

PR #656, opened 2024-09-25
Workaround for object_type in py>=0.32.0

ktsitsi:
This PR:

- Fixes all the occurrences of if-condition statements in ingestors that use the `tiledb.object_type` to work around a regression in tiledb-py==0.32.0 and upwards.


sgillies (approved):
@ktsitsi thank you!

gspowley (approved):

ktsitsi:
@Thomas-tiledb @sgillies we need to revert this one.

PR #657, opened 2024-09-25
Add tests of with_metadata option

sgillies:
The low-level autogenerated API changes have already been merged. Wiring up the high-level API is straight forward. I've run the functions against my own groups and arrays and see what I expect. Note well: some of our arrays have very large metadata and listing them using `with_metadata=True` will appear to hang your computer. It's really just transferring a lot of bytes across the network.

I've added some new tests. They don't check the results, but do check that we make a request to the API endpoints with the right query params.

### DIFF
--- a/src/tiledb/cloud/client.py
+++ b/src/tiledb/cloud/client.py
@@ -177,6 +177,7 @@ def list_public_arrays(
     page=None,
     per_page=None,
     async_req=False,
+    with_metadata=None,
 ):
     """
     List public arrays
@@ -191,6 +192,7 @@ def list_public_arrays(
     :param int page: optional page for pagination
     :param int per_page: optional per_page for pagination
     :param async_req: return future instead of results for async support
+    :param bool with_metadata: include the metadata of the array
     :return: list of all array metadata you have access to that meet the filter applied
     """
 
@@ -217,6 +219,8 @@ def list_public_arrays(
             kwargs["page"] = page
         if per_page is not None:
             kwargs["per_page"] = per_page
+        if with_metadata is not None:
+            kwargs["with_metadata"] = with_metadata
         res = api_instance.arrays_browser_public_get(**kwargs)
 
         # if the user didn't ask for pagination just return raw array list
@@ -237,6 +241,7 @@ def list_shared_arrays(
     page=None,
     per_page=None,
     async_req=False,
+    with_metadata=None,
 ):
     """
     List shared arrays
@@ -251,6 +256,7 @@ def list_shared_arrays(
     :param int page: optional page for pagination
     :param int per_page: optional per_page for pagination
     :param async_req: return future instead of results for async support
+    :param bool with_metadata: include the metadata of the array
     :return: list of all array metadata you have access to that meet the filter applied
     """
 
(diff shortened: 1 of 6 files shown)


tests/test_basic.py
```diff
@@ -437,3 +440,37 @@ def needsUnittestUser(self):
         """Skips the test unless it is run as the ``unittest`` user."""
         if not testonly.is_unittest_user():
             self.skipTest("May fail with non-unittest users.")
+
+
+@pytest.fixture(scope="function")
+def with_metadata_checker(monkeypatch):
```
sgillies, line 446: This fixture monkeypatches the urllib3 pool manager to intercept outgoing requests. It's a lightweight alternative to setting up a proxy or some other more complicated software.

tests/test_basic.py
```diff
@@ -437,3 +440,37 @@ def needsUnittestUser(self):
...
+    """Intercept requests to arrays and groups list endpoints and checks the URLs."""
+
+    def fake_and_check_request(self, method, url, *args, **kwargs):
+        assert "with_metadata=True" in url
```
sgillies, line 450: Here's the actual test assertion, inside the fixture. If it fails, pytest does the right thing and reports an error that's quite understandable.

sgillies: @ktsitsi @spencerseale @gspowley what do you think of this approach for testing our SDK - API interface? Is it one that you'd be comfortable using and maintaining with me?

sgillies, line 450: I honestly don't know which side, client or server, has the responsibility for getting the values of these boolean options right.

src/tiledb/cloud/client.py
```diff
@@ -177,6 +177,7 @@ def list_public_arrays(
     page=None,
     per_page=None,
     async_req=False,
+    with_metadata=None,
```
spencerseale, line 180: Should this default to `False`? We don't have any pydantic like validator so in theory a str or some other dtype would make it into `kwargs` and the api request

tests/test_basic.py
spencerseale: So basically, we mock the actual request and just ensure that whatever selection we're testing is passed into the request?

sgillies: @spencerseale correct. Actual integration tests are good, too, but these new tests run really fast and could help eliminate a lot of lurking generic bugs living between the high level Python API and the low level auto-generated API.

src/tiledb/cloud/client.py
sgillies: @spencerseale I asked myself the same question. I'm defaulting to `None` so that we don't put a `with_metadata=False` in our request URL and so that we default to exactly (I hope!) same response as before.

`async_req` is a different case, it isn't passed to our API servers, but leads to a new branch in our low-level client code.

tests/test_basic.py
Shelnutt2: I love the mocking! Great way to test and interate in the development and allow more comprehensive testing in the future without needing a full stack.

Shelnutt2: I usually opt for the server to be more flexible but we should also standardize it in the openapi definition

Shelnutt2 (approved):

src/tiledb/cloud/client.py
spencerseale: I see, thank you for explaining that!

tests/test_basic.py
spencerseale: Thanks @sgillies, I will be sure to follow this strategy.

spencerseale:
I have a project that could use this new feature, will happily use it if it can go into next cloud-py release

PR #658, opened 2024-09-27
Add http(s) scheme if missing when logging in

sgillies:
Usually, a hostname doesn't include a protocol scheme, but our SDK strictly requires the http(s) scheme. Thus, we add it if it is missing when logging in.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -4,7 +4,24 @@ import tiledb.cloud
 import tiledb.cloud.config
 
 
-def test_login_bare_host():
+def test_login_bare_host(monkeypatch, tmp_path):
     """Accept a bare host, store it with https scheme."""
+    monkeypatch.setattr(
+        tiledb.cloud.client.config,
+        "default_config_file",
+        tmp_path.joinpath("cloud.json"),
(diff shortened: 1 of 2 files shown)


ihnorton (approved):

Thomas-tiledb (approved):

sgillies:
The worker that ran `tests/test_dag.py::DAGClassTest::test_kwargs` is stuck. I'm going to note this and merge regardless.

tests/test_config.py
```diff
@@ -4,7 +4,24 @@
...
+        "default_config_file",
+        tmp_path.joinpath("cloud.json"),
+    )
+    monkeypatch.setattr(
+        tiledb.cloud.client.config,
+        "config",
+        tiledb.cloud.config.configuration.Configuration(),
+    )
+    monkeypatch.setattr(tiledb.cloud.client.config, "logged_in", False)
+    monkeypatch.setattr(tiledb.cloud.client, "client", tiledb.cloud.client.Client())
```
sgillies, line 20: Here's the fix. A reversible login is tricky to do.

PR #659, opened 2024-10-02
Remove xfail on 3 graph connection tests

sgillies:
The upstream issue has been resolved and these marks can be removed.


ihnorton (approved):
Thanks for pushing this through!

PR #661, opened 2024-10-03
as_batch gets parameters from kwargs, never popping

sgillies:


src/tiledb/cloud/utilities/_common.py
```diff
@@ -343,11 +343,16 @@ def wrapper(*args, **kwargs) -> Dict[str, object]:
         """
 
         name = kwargs.get("name", func.__name__)
+
+        # The wrapper's batch mode graph uses several of the parameters
+        # that will be passed to the wrapped function for use in its
+        # own graph.
         namespace = kwargs.get("namespace", None)
+        resources = kwargs.get("resources", None)
```
sgillies, line 351: @ktsitsi I'm pinging you for review because this PR reverts the change you made in f10aa95. If we pop the resources parameter off, then it isn't propagated to the SOMA ingestion workflow.

ktsitsi: Either popping or getting it the `resources` argument is being assigned here https://github.com/TileDB-Inc/TileDB-Cloud-Py/blob/689352404f6561db1f7dc73782a2cecd639aa2bc/src/tiledb/cloud/utilities/_common.py#L365 for all ingestors that have this positional argument. (which all do). Linking this https://tiledb.slack.com/archives/C06NN0F195X/p1711025149390339 for reference. The popping came up since we were facing issues with double arguments registration, when the positional ones given to the submit function were also left inside the `kwargs`.

ktsitsi (changes requested):

PR #664, opened 2024-10-08
Preserve SOMA ingest keyword arguments

sgillies:
(Edited because I came up with a better idea)

The root cause is that `Node()` consumes its `resources` keyword argument and does not propagate it to the cloud function. In ingestion code, we've overlooked this or worked around it.

I propose adding a `_propagate_resources` flag to `Node()`. If `True` it does not consume the parameters and allows them to be passed to the function. The default should be `False` because it's likely that cloud functions aren't expecting any new keyword arguments.

For ingestion code, whenever you pass `resources` to `DAG.submit()`, consider also passing `_propagate_resources=True` if your cloud function needs to specify resources. Otherwise, they'll be lost.

### DIFF
--- a/src/tiledb/cloud/soma/ingest.py
+++ b/src/tiledb/cloud/soma/ingest.py
@@ -25,6 +25,7 @@ def register_dataset_udf(
     namespace: Optional[str] = None,
     config: Optional[Mapping[str, Any]] = None,
     verbose: bool = False,
+    **kwargs,
 ) -> None:
     """
     Register the dataset on TileDB Cloud.
@@ -233,6 +234,7 @@ def ingest_h5ad(
     ingest_mode: str = "write",
     logging_level: int = logging.INFO,
     dry_run: bool = False,
+    **kwargs,
 ) -> None:
     """Performs the actual work of ingesting H5AD data into TileDB.
 
@@ -402,12 +404,17 @@ def run_ingest_workflow(
         mode=dag.Mode.BATCH,
     )
 
-    # Step 1: Ingest workflow UDF
-    carry_along: Dict[str, str] = {
-        "resources": _DEFAULT_RESOURCES if resources is None else resources,
-        "namespace": namespace,
-        "access_credentials_name": acn,
-    }
+    carry_along: Dict[str, str] = kwargs.pop(
+        "carry_along",
+        {
+            "resources": _DEFAULT_RESOURCES if resources is None else resources,
+            # I don't see why this is needed. I'm not finding any code that
+            # pops "namespace" from kwargs.
+            "namespace": namespace,
+            # Because workflows might be using older user-defined functions?
+            "access_credentials_name": acn,
+        },
+    )
 
     grf.submit(
         _run_ingest_workflow_udf_byval,
--- a/src/tiledb/cloud/soma/ingest.py
+++ b/src/tiledb/cloud/soma/ingest.py
@@ -87,7 +87,7 @@ def run_ingest_workflow_udf(
     acn: Optional[str] = None,
     logging_level: int = logging.INFO,
     dry_run: bool = False,
-    _resources: Optional[Dict[str, Any]] = None,
+    _propagate_resources: bool = False,
     **kwargs,
 ) -> Dict[str, str]:
     """
@@ -95,11 +95,6 @@ def run_ingest_workflow_udf(
     can we do VFS with access_credentials_name -- that does not work correctly
     on the client.
     """
-
-    # "_resources" is the data to be propagated to the next cloud function.
-    _resources = _resources or {}
-    if resources:
-        _resources["resources"] = resources
 
     # Some kwargs are eaten by the tiledb.cloud package, and won't reach
     # our child. In order to propagate these to a _grandchild_ we need to
@@ -139,11 +134,11 @@ def run_ingest_workflow_udf(
             extra_tiledb_config=extra_tiledb_config,
             ingest_mode=ingest_mode,
             platform_config=platform_config,
-            resources=resources or _resources.get("resources", None),
+            resources=resources or _DEFAULT_RESOURCES,
             access_credentials_name=carry_along.get("access_credentials_name", acn),
             logging_level=logging_level,
             dry_run=dry_run,
-            _resources=_resources,
+            _propagate_resources=_propagate_resources,
         )
 
     elif vfs.is_dir(input_uri):
@@ -188,11 +183,11 @@ def run_ingest_workflow_udf(
                 extra_tiledb_config=extra_tiledb_config,
                 ingest_mode=ingest_mode,
                 platform_config=platform_config,
-                resources=resources or _resources.get("resources", None),
+                resources=resources or _DEFAULT_RESOURCES,
                 access_credentials_name=carry_along.get("access_credentials_name", acn),
                 logging_level=logging_level,
                 dry_run=dry_run,
-                _resources=_resources,
+                _propagate_resources=_propagate_resources,
             )
             collector.depends_on(node)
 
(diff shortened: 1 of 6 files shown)


src/tiledb/cloud/soma/ingest.py
```diff
@@ -25,6 +25,7 @@ def register_dataset_udf(
     namespace: Optional[str] = None,
     config: Optional[Mapping[str, Any]] = None,
     verbose: bool = False,
+    **kwargs,
```
sgillies, line 28: Adding a catch-all `**kwargs` to all of our distributed functions allows them to ignore unexpected keyword arguments. I think we might make this standard?

src/tiledb/cloud/utilities/_common.py
```diff
@@ -308,8 +308,12 @@ def _filter_kwargs(function: Callable, kwargs: Mapping[str, Any]) -> Mapping[str
     :return: filtered kwargs
     """
     valid_args = inspect.signature(function).parameters
-    filtered_kwargs = {k: v for k, v in kwargs.items() if k in valid_args}
-    return filtered_kwargs
+
+    # Detect if the function has a catch-all **kwargs parameter.
+    # If so, we'll preserve all keyword arguments.
+    has_kwargs = any(param.kind == param.VAR_KEYWORD for param in valid_args)
```
sgillies, line 314: This allows a distributed function's `**kwargs`` to work under `as_batch()`.

sgillies, line 314: This allows a distributed function's `**kwargs`` to work under `as_batch()`.

src/tiledb/cloud/utilities/_common.py
```diff
@@ -356,7 +363,16 @@ def wrapper(*args, **kwargs) -> Dict[str, object]:
...
 
-        # Submit the function as a batch UDF
+        # Guard against Node constructor peeling off "resources" or
+        # other functions peeling off "namespace" and "credentials".
+        # The "carry along" pattern was introduced in gh-630.
+        kwargs["carry_along"] = {
+            "resources": resources,
+            "namespace": namespace,
+            "access_credentials_name": acn,
+        }
```
sgillies, line 373: It looks to me that we need to extend the "carry along" pattern all the way into `as_batch()` so that `graph.submit()` below doesn't lose the resources parameter.

src/tiledb/cloud/utilities/_common.py
sgillies: My new test caught a bug in this that has been fixed.

sgillies:
The wheel install test failure is unrelated.

Some changes since the first iteration: I propose we propagate "resources" and "resource_class" differently from "namespace" and "access_credentials_name".

The first two are directly impacted by `graph.submit()` and this impact will not go away unless we rewrite `dag.Node`. It's a specific problem and I want to treat it with a specific solution. The root problem with "access_credentials_name" is different. I'm not convinced there is a problem with "namespace" at all, I can't find any code that consumes it.

Solution: If you want to run a cloud function with `graph.submit()`, and you want that function to receive "resources" and "resource_class", you copy these parameters into a dict named "_resources" and then pass that dict to `graph.submit()`. On the other end, cloud functions should expect to receive a keyword argument named "_resources", and should unpack those parameters from it.

I've modified `as_batch()` and other SOMA ingestion functions to do the above.

I've added a test to demonstrate and now we have 100% test coverage of `as_batch()`. This PR fixes our main problem with specifying resources for SOMA 1-click.

ktsitsi:
Two comments on this one:
- Will we need changes in the current API of `bioimg.ingest` and `vcf.ingest` functions?
- `bioimg.ingest` is being tested in our rest CI repo just before deployment https://github.com/TileDB-Inc/TileDB-Cloud-REST/blob/1a05ac341fb6a00c67a93b5b6351fc52da3474be/testing/smoke/categories/bioimg.py . TMK it is the only ingestor being tested eventually before deployment. But still I don't think that architecturally speaking we should move testing code there and make these tests heavier than they already are, which delays deployment and catches any possible error in the last step before deployment.
- Thinking about a general approach that might be useful in the future and would standarize the process is to decouple the ingestors from cloud-py repo. This is an architecture decision that we need to discuss. I was thinking that we could/should implement a Protocol/Plugin pattern and exposing it with entrypoints so the ingest code lives inside the standalone VCF, BIOIMG and SOMA following this Protocol standard. I have implemented that for different Converters in TileDB-Bioimaging.

tests/utilities/test_as_batch.py
```diff
@@ -0,0 +1,27 @@
...
+import unittest
+import unittest.mock
+
+import tiledb.cloud.utilities._common
+from tiledb.cloud.utilities._common import as_batch
+
+
+@unittest.mock.patch.object(tiledb.cloud.utilities._common.dag, "DAG")
+def test_propagate_resources(mock_dag):
+    """Assert that as_batch propagates resource parameters."""
```
sgillies, line 12: This test shows that `as_batch()` is doing it's job: passing a key parameter that Cloud-UI needs, but doesn't know about, on to cloud functions.

tests/test_dag.py
```diff
@@ -1334,6 +1334,53 @@ def test_replace_stored_params_pandas(self):
...
+        resources={"cpu": "8", "memory": "32Gi"},
+        _propagate_resources=True,
+    )
+    assert node._name == "test"
+    assert node._resources == {"cpu": "8", "memory": "32Gi"}
+    assert "arg1" in node.args
+    assert "arg2" in node.args
+    assert node.kwargs["kwarg1"] == "foo"
+    assert node.kwargs["resources"] == {"cpu": "8", "memory": "32Gi"}
+
```
sgillies, line 1382: Testing that `Node()` behaves properly when receiving our new flag.

src/tiledb/cloud/soma/ingest.py
```diff
@@ -451,4 +463,4 @@ def run_ingest_workflow(
 _register_dataset_udf_byval = functions.to_register_by_value(register_dataset_udf)
 _hack_patch_anndata_byval = functions.to_register_by_value(_hack_patch_anndata)
 
-ingest = as_batch(_run_ingest_workflow_byval)
+ingest = as_batch(_run_ingest_workflow_byval, _propagate_resources=True)
```
sgillies, line 466: This means: propagate resources specified by Cloud-UI.

src/tiledb/cloud/soma/ingest.py
```diff
@@ -402,12 +408,17 @@ def run_ingest_workflow(
...
-        "namespace": namespace,
-        "access_credentials_name": acn,
-    }
+    carry_along: Dict[str, str] = kwargs.pop(
+        "carry_along",
+        {
+            "resources": _DEFAULT_RESOURCES if resources is None else resources,
+            # I don't see why this is needed. I'm not finding any code that
+            # pops "namespace" from kwargs.
+            "namespace": namespace,
```
sgillies, line 417: Actually, I do see functions accidentally consuming `namespace`. But let's treat that differently in a different piece of work.

sgillies:
@ktsitsi I've updated bioimg ingestion, which was simple.

@gspowley I haven't done anything about VCF ingestion. If you haven't hit resource limits yet, maybe no action is needed.

We need review of this pretty seriously next week 🙏

src/tiledb/cloud/soma/ingest.py
```diff
@@ -233,6 +237,7 @@ def ingest_h5ad(
     ingest_mode: str = "write",
     logging_level: int = logging.INFO,
     dry_run: bool = False,
+    **kwargs,
```
sgillies, line 240: Addition of `**kwargs` for forward compatibility.

gspowley:
VCF ingestion has code to auto-scale resources based on the batch size and supports overriding the auto-scale value with an `ingest_resources` arg, which avoids the parameter dropping issue.

I believe no action is required. I'll review and run some tests though.

sgillies:
@gspowley @johnkerl @ktsitsi would you all be more approving of a PR that did *not* modify `Node` and instead used the same approach as VCF?

sgillies:
Closing this one. It feels like too much change, too fast.

PR #669, opened 2024-10-14
Propagate resources to ingest_h5ad node

sgillies:
Take 2: solve the resources propagation problem like VCF does. Instead of #664.

Sorry about the churn @johnkerl, @gspowley, @ktsitsi. How about this less risky approach proven in VCF? I'll make the corresponding PR for our 1-click UI.

### DIFF
--- a/src/tiledb/cloud/soma/ingest.py
+++ b/src/tiledb/cloud/soma/ingest.py
@@ -124,6 +124,9 @@ def run_ingest_workflow_udf(
             namespace=carry_along.get("namespace", namespace),
         )
 
+        # We've propagated ingest_resources to this function so that
+        # we can use it as the resources argument of this method
+        # call.
         h5ad_ingest = grf.submit(
             _ingest_h5ad_byval,
             output_uri=output_uri,
@@ -132,7 +135,7 @@ def run_ingest_workflow_udf(
             extra_tiledb_config=extra_tiledb_config,
             ingest_mode=ingest_mode,
             platform_config=platform_config,
-            resources=ingest_resources,
+            resources=ingest_resources,  # Apply propagated resources here.
             access_credentials_name=carry_along.get("access_credentials_name", acn),
             logging_level=logging_level,
             dry_run=dry_run,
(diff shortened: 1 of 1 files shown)


src/tiledb/cloud/soma/ingest.py
```diff
@@ -149,7 +149,7 @@ def run_ingest_workflow_udf(
         grf = dag.DAG(
             name=name,
             mode=dag.Mode.BATCH,
-            namespace=namespace,
+            namespace=carry_along.get("namespace", namespace),
```
sgillies, line 152: Not related, but clearly missed in the original work.

src/tiledb/cloud/soma/ingest.py
```diff
@@ -132,7 +132,7 @@ def run_ingest_workflow_udf(
             extra_tiledb_config=extra_tiledb_config,
             ingest_mode=ingest_mode,
             platform_config=platform_config,
-            resources=carry_along.get("resources", resources),
+            resources=ingest_resources,
```
sgillies, line 135: Here is where we want the UI-selected resources to land and stick.

src/tiledb/cloud/soma/ingest.py
```diff
@@ -418,7 +419,7 @@ def run_ingest_workflow(
         extra_tiledb_config=extra_tiledb_config,
         platform_config=platform_config,
         ingest_mode=ingest_mode,
-        resources=resources,
+        ingest_resources=ingest_resources,
```
sgillies, line 422: propagates the UI-selected resources.

ktsitsi (approved):

ktsitsi:
Will we need a `carry_along` for bioimg too?

src/tiledb/cloud/soma/ingest.py
johnkerl: Please add a comment in the code, right here, saying just that. :)

gspowley (approved):

johnkerl (approved):

sgillies:
@ktsitsi I honestly don't 100% understand why we're doing `carry_along` here. If you haven't noticed you need it for bio imaging ingestion, then I don't think you need it.

PR #670, opened 2024-10-15
Skip sub-folders and make pattern matching more useful

sgillies:
These changes help unlock 1-click ingestion of folders of SOMA sources.

Unneeded log messages have been removed.

### DIFF
--- a/src/tiledb/cloud/soma/ingest.py
+++ b/src/tiledb/cloud/soma/ingest.py
@@ -178,6 +178,9 @@ def run_ingest_workflow_udf(
             namespace=carry_along.get("namespace", namespace),
         )
 
+        # We've propagated ingest_resources to this function so that
+        # we can use it as the resources argument of this method
+        # call.
         h5ad_ingest = grf.submit(
             _ingest_h5ad_byval,
             output_uri=output_uri,
@@ -186,7 +189,7 @@ def run_ingest_workflow_udf(
             extra_tiledb_config=extra_tiledb_config,
             ingest_mode=ingest_mode,
             platform_config=platform_config,
-            resources=carry_along.get("resources", resources),
+            resources=ingest_resources,  # Apply propagated resources here.
             access_credentials_name=carry_along.get("access_credentials_name", acn),
             logging_level=logging_level,
             dry_run=dry_run,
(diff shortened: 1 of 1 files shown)


src/tiledb/cloud/soma/ingest.py
```diff
@@ -168,8 +175,12 @@ def run_ingest_workflow_udf(
             entry_output_uri += base
             logger.debug("ENUMERATOR ENTRY_OUTPUT_URI=%r", entry_output_uri)
 
-            if pattern is not None and not re.match(pattern, entry_input_uri):
-                logger.debug("ENUMERATOR SKIP NO MATCH ON <<%r>>", pattern)
+            if pattern is not None and not re.search(pattern, entry_input_uri):
```
sgillies, line 178: `re.search()` matches anywhere in the string.

src/tiledb/cloud/soma/ingest.py
```diff
@@ -168,8 +175,12 @@ def run_ingest_workflow_udf(
...
+                logger.info(
+                    "Skipping non-matching input: pattern=%r, entry_input_uri=%r",
```
sgillies, line 180: I'm going to start nudging all log entry into this form. `Event happened: var1=foo, var2=bar`.

Tile-Kyle:
can I confirm the changes here will skip any sub folders from the `source` input path? which I believe was included in the original [PRD](https://www.notion.so/SOMA-1-Click-Ingestion-95962a7e579144fcbd55af7180b87b93?pvs=4#e48b1eb08256464dbbc5e6b94e611265) - but just checking. Since we will need to communicate this to the user on the UI. Thanks!

src/tiledb/cloud/soma/ingest.py
```diff
@@ -103,44 +103,13 @@ def run_ingest_workflow_udf(
...
-            resources=carry_along.get("resources", resources),
-            access_credentials_name=carry_along.get("access_credentials_name", acn),
-            logging_level=logging_level,
-            dry_run=dry_run,
-        )
-
-    elif vfs.is_dir(input_uri):
-        logger.debug("ENUMERATOR VFS.IS_DIR")
-
+    if vfs.is_dir(input_uri):
```
sgillies, line 112: Here's the fix! `vfs.is_file(input_uri)` returns `True`, erroneously. We need to always test with `is_file()` first.

sgillies:
@Tile-Kyle yes, that's right. Only top-level contents of a folder are to be ingested.

sgillies:
@ktsitsi tagging you for review because I want you to see this issue with `VFS`.

johnkerl (approved):
🚢

PR #671, opened 2024-10-17
Add ingest_resources for use with UI

sgillies:
Also remove a few params that look unnecessary.
[sc-57856]

### DIFF
--- a/src/tiledb/cloud/bioimg/ingestion.py
+++ b/src/tiledb/cloud/bioimg/ingestion.py
@@ -219,7 +219,8 @@ def ingest(
             "png": Converters.PNG,
         }.get(converter, Converters.OMETIFF)
 
-        experimental_reader = kwargs.get("experimental_reader", False)
+        experimental_reader = kwargs.pop("experimental_reader", False)
+
         compressor = kwargs.get("compressor", None)
         if compressor:
             compressor_args = dict(compressor)
--- a/src/tiledb/cloud/bioimg/ingestion.py
+++ b/src/tiledb/cloud/bioimg/ingestion.py
@@ -35,6 +35,7 @@ def ingest(
     num_batches: Optional[int] = None,
     threads: Optional[int] = 0,
     resources: Optional[Mapping[str, Any]] = None,
+    ingest_resources: Optional[Mapping[str, Any]] = None,
     compute: bool = True,
     register: bool = True,
     mode: Optional[Mode] = Mode.BATCH,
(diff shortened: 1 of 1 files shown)


src/tiledb/cloud/bioimg/ingestion.py
```diff
@@ -381,7 +386,7 @@ def register_dataset_udf(
         *args,
         name=f"{dag_name} ingestor ",
         expand_node_output=input_list_node,
-        resources=DEFAULT_RESOURCES if resources is None else resources,
+        resources=resources or ingest_resources or DEFAULT_RESOURCES,
```
sgillies, line 389: @ktsitsi Users of `ingest(resources={...})` without the `as_batch()` wrapper will get the behavior they expect. At the same time, our UI can specify `ingest_resources` and `as_batch()` will pass it through. How does that sound? I'll make the corresponding change upstream in the UI repo.

src/tiledb/cloud/bioimg/ingestion.py
```diff
@@ -400,8 +405,6 @@ def register_dataset_udf(
             namespace=namespace,
             name=f"{dag_name} registrator ",
             expand_node_output=ingest_list_node,
-            resources=DEFAULT_RESOURCES if resources is None else resources,
```
sgillies, line 403: Unless I'm missing something, registration doesn't need special resources. Does this look okay @ktsitsi ?

src/tiledb/cloud/bioimg/ingestion.py
```diff
@@ -400,8 +405,6 @@ def register_dataset_udf(
...
-            image_name=DEFAULT_IMG_NAME,
```
sgillies, line 404: Same for the image, I think. The basic image should work, yes?

src/tiledb/cloud/bioimg/ingestion.py
ktsitsi: For this one the architecture we are using is dynamic batch taskgraphs. I don't know if changes the resources between the different stages will lead to downscales of running pods or sth. Mainly I had it this way to be sure that the different stages correspond to the same node resources. This is a nice exercise though to test.

ktsitsi: Indeed though you are right the registration stage is supposed to be lightweight so no special resources are generally needed.

src/tiledb/cloud/bioimg/ingestion.py
```diff
@@ -219,7 +219,8 @@ def ingest_tiff_udf(
             "png": Converters.PNG,
         }.get(converter, Converters.OMETIFF)
 
-        experimental_reader = kwargs.get("experimental_reader", False)
+        experimental_reader = kwargs.pop("experimental_reader", False)
```
sgillies, line 222: @ktsitsi I found that if I didn't pop off this keyword argument, `from_bioimg()` could receive it twice.

sgillies:
This works. The extra resources are only needed for the task that touches the imagery. Prep and registration tasks work fine with minimal resources.

Untitled

PR #672, opened 2024-10-21
Remove value registration wrappers

sgillies:
George's cloudpickle approach makes them obsolete.


johnkerl (changes requested):
> George's cloudpickle makes them obsolete

Let's get that procedure written out explicitly in this repo's wiki first.

johnkerl (approved):
🚢 

Thanks @sgillies !!!!

PR #674, opened 2024-10-22
Move test requirements, adjust and streamline CI YAML

sgillies:
This moves the requirements files to a new /ci folder, updates the YAML, and removes some unneeded installation when building wheels. The goal: get all of us developers to use exactly the same dependencies defined in pyproject.toml.


.github/workflows/tiledb-cloud-py.yaml
```diff
@@ -233,8 +233,7 @@ jobs:
 
       - name: Build wheel
         run: |
-          pip install --upgrade pip build wheel setuptools setuptools-scm
-          pip install -r requirements-py3.9.txt
```
sgillies, line 237: Build dependencies are defined in pyproject.toml and will be fetched by our build backend. The test dependencies aren't needed at build time.

spencerseale:
Can we just install from `pyproject.toml` within ci and remove requirements.txt related files altogether?

sgillies:
@spencerseale From my perspective, changing our CI environment is a different matter. Switching from a requirements file, with versioning pinning, to the largely unpinned specs in pyproject.toml means that our CI builds will be less reproducible and will start to fail occasionally due to changes that we ourselves have not made to cloud-py. Let's have a discussion about that, but not as a part of this PR, okay?

spencerseale:
@sgillies yes that sounds good, but I was definitely pushing for pinning within a pyproject.toml PEP631

spencerseale (approved):

PR #676, opened 2024-10-29
fix obs and var defaults to return all values if none spec'd

spencerseale:
Setting as empty list if `None` specified for `obs_attrs` and/or `var_attrs` results in an `AnnData` not having all attributes in obs/var dataframes.


Shelnutt2 (approved):

sgillies (approved):
@spencerseale Just a bug fix, yes? No extra documentation needed?

spencerseale:
@sgillies yes just bug fix. I want to make sure @aaronwolen reviews prior to merge. There may be some historical context on why we do the `x or []` logic.

PR #677, opened 2024-10-31
Add region property to DAG

spencerseale:
Add `DAG.region` for better detection of DAG host region


spencerseale:
The only issue I foresee is if someone overrides region after DAG already started. `DAG.region` as it's implemented here, gets current region. The alternative would be to return a non-`NoneType` only after `DAG.compute` is fired. Since a `DAG` is only usable for one task graph execution. However the original inspiration for this was to validate region prior to executing DAG.

sgillies:
@spencerseale if I read the code correctly, the new property doesn't tell us for sure what region the DAG is running in, it tells us the current client configuration. Which is pretty much what you said above, right?

I think this goes against the OO principle of encapsulation. DAG methods should be based on DAG data, not data in another class or module.

spencerseale:
@sgillies That's correct and you're right. I can take another stab at it so it's pinned to the DAG object.

PR #679, opened 2024-11-11
Register ingested groups using the correct URI

sgillies:
In the directory ingestion case, the deployed code creates groups at `OUTPUT_URI/NAME/NAME/` (note the doubled `NAME`) where `NAME` is the basename/stem of an H5ad file. But then it tries to register a group at `OUTPUT_URI`. There's no group there, so this fails.

As a part of solving this problem, I'm generalizing directory ingest to cover the single file case. We no longer have two ingestion branches. There's much less code and I think the intent of the remaining code is more clear. This change helped me fix the bug and helps us make sure it stays fixed.

There's an output change, though: every H5ad file becomes a group at `OUTPUT_URI/NAME/` where `NAME` is the basename/stem of the H5ad file no matter what. Ingested as a single file, or as an item in a folder, the same output. In theory (and practice, often) this kind of symmetry and de-specialization is a win. I'm going to need some review from actual SOMA users and developers, for sure.

The test failures are unrelated. Trouble with task graph run times again.

### DIFF
--- a/src/tiledb/cloud/soma/ingest.py
+++ b/src/tiledb/cloud/soma/ingest.py
@@ -1,4 +1,5 @@
 import logging
+import os.path
 import pathlib
 import re
 import warnings
@@ -138,7 +139,8 @@ def run_ingest_workflow_udf(
     )
 
     for input_file in input_files:
-        output_group_uri = pathlib.Path(output_uri.joinpath(input_file.stem))
+        stem = pathlib.Path(input_file).stem
+        output_group_uri = os.path.join(output_uri, stem)
         logger.info(
             "Building task for h5ad file: input_file=%r, output_group_uri=%r",
             input_file,
@@ -171,7 +173,7 @@ def run_ingest_workflow_udf(
             )
             register_soma.depends_on(collector)
 
-    logger.info("Computing DAG: grf=%r, grf")
+    logger.info("Computing DAG: grf=%r", grf)
     grf.compute()
     return grf.server_graph_uuid
 
--- a/src/tiledb/cloud/soma/ingest.py
+++ b/src/tiledb/cloud/soma/ingest.py
@@ -1,5 +1,5 @@
 import logging
-import os
+import pathlib
 import re
 import warnings
 from typing import Any, ContextManager, Dict, Mapping, Optional
@@ -34,9 +34,7 @@ def register_dataset_udf(
     :param config: config dictionary, defaults to None
     :param verbose: verbose logging, defaults to False
     """
-
     logger = get_logger_wrapper(verbose)
-
     namespace = namespace or tiledb.cloud.user_profile().default_namespace_charged
     tiledb_uri = f"tiledb://{namespace}/{register_name}"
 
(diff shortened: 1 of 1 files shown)


sgillies:
Directory ingestion works: https://beta.cloud.tiledb.com/taskgraphlogs/sean-gillies/dd293232-6537-44b5-aa20-9b7befc75787.

Single file: https://beta.cloud.tiledb.com/taskgraphlogs/sean-gillies/8f05b23d-e53b-41b7-a964-0ffe4159a69f.

sgillies:
@johnkerl @aaronwolen can you check the soundness of my decision to normalize the ingestion outputs as I have done?

src/tiledb/cloud/soma/ingest.py
```diff
@@ -103,87 +102,54 @@ def run_ingest_workflow_udf(
 
     logger = get_logger_wrapper(level=logging_level)
 
-    h5ad_ingest = None
-    collector = None
-
+    input_files = []
```
sgillies, line 105: We build up a list of H5ad files to ingest as our first step.

src/tiledb/cloud/soma/ingest.py
```diff
@@ -103,87 +102,54 @@ def run_ingest_workflow_udf(
...
+        input_files.append(input_item)
+    else:
+        raise ValueError("input_uri %r is neither a file nor a directory", input_uri)
 
-        grf = dag.DAG(
-            name=name,
-            mode=dag.Mode.BATCH,
-            namespace=carry_along.get("namespace", namespace),
-        )
+    logger.info("Building DAG for SOMA ingestion: input_files=%r", input_files)
```
sgillies, line 130: After we've made the input files list, we only need one branch of code to set up and execute this UDF's task graph.

src/tiledb/cloud/soma/ingest.py
```diff
@@ -103,87 +102,54 @@ def run_ingest_workflow_udf(
 
     logger = get_logger_wrapper(level=logging_level)
 
-    h5ad_ingest = None
-    collector = None
```
sgillies, line 107: Not needed.

src/tiledb/cloud/soma/ingest.py
```diff
@@ -103,87 +102,54 @@ def run_ingest_workflow_udf(
...
+        output_uri=output_uri,
+    )
 
-        # We've propagated ingest_resources to this function so that
-        # we can use it as the resources argument of this method
-        # call.
-        h5ad_ingest = grf.submit(
+    for input_file in input_files:
+        stem = pathlib.Path(input_file).stem
+        output_group_uri = os.path.join(output_uri, stem)
```
sgillies, line 143: I originally tried to use pathlib on this line, too, but it doesn't respect the double slashes in our input paths.

src/tiledb/cloud/soma/ingest.py
```diff
@@ -103,87 +102,54 @@ def run_ingest_workflow_udf(
...
-                )
-                continue
-
-            base = os.path.basename(entry_input_uri)
-            base, _ = os.path.splitext(base)
-
-            entry_output_uri = output_uri + "/" + base
-            if not output_uri.endswith("/"):
-                entry_output_uri += "/"
-            entry_output_uri += base
```
sgillies, line 144: I don't understand why we appended `base` to the output path twice. Now we only do it once.

src/tiledb/cloud/soma/ingest.py
sgillies: Note that every group is created at `{output_uri}/{stem}` now. There's always exactly one extra path part.

src/tiledb/cloud/soma/ingest.py
johnkerl: I think it's not a requirement

johnkerl (approved):

PR #680, opened 2024-11-12
typ{o,e} fixes

ryan-williams:
- First arg to `TileDBCloudError` is supposed to be `http_status: Optional[int]`, but these callsites were passing a `str.
  - I guessed that those should be `body` (and `body` should be allowed to be `str | bytes`, not just `bytes`)
- Couple docstr/comment typos

### DIFF
--- a/src/tiledb/cloud/dag/dag.py
+++ b/src/tiledb/cloud/dag/dag.py
@@ -1655,7 +1655,7 @@ class DAG:
                 if name not in _SKIP_BATCH_UDF_KWARGS
             }
 
-            all_args = types.Arguments(node_args, filtered_node_kwargs)
+            all_args = types.Arguments(tuple(node_args), filtered_node_kwargs)
             encoder = _BatchArgEncoder(input_is_expanded=bool(node._expand_node_output))
             kwargs["arguments"] = encoder.encode_arguments(all_args)
 
--- a/src/tiledb/cloud/dag/dag.py
+++ b/src/tiledb/cloud/dag/dag.py
@@ -28,7 +28,6 @@ from typing import (
 from .. import array
 from .. import client
 from .. import rest_api
-from .. import tiledb_cloud_error as tce
 from .. import udf
 from .._common import functions
 from .._common import futures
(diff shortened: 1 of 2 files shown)


sgillies (changes requested):
@ryan-williams Unless I've overlooked one, all these instances of `raise tce.TileDBCloudError` happen when validating inputs before taking the step of creating a graph or node. They are not associated with making a REST server request, right?

Thus, I believe we shouldn't be raising this particular exception in these cases. Let's raise ValueError instead, like at https://github.com/TileDB-Inc/TileDB-Cloud-Py/blob/8c2da9916f1bc5957ea880c04cc31428d7008695/src/tiledb/cloud/dag/dag.py#L2016. ValueError takes just a string.

ryan-williams:
Good point, converted them to `ValueError`s

src/tiledb/cloud/dag/dag.py
```diff
@@ -1083,12 +1078,12 @@ def submit_udf_stage(
             **kwargs,
         )
 
-    def submit_sql(self, *args: Any, **kwargs: Any) -> Node:
+    def submit_sql(self, sql: str, *args: Any, **kwargs: Any) -> Node:
```
sgillies, line 1081: @ryan-williams If we pick `sql` out of `*args`, then I think we have to pass it explicitly on line 1092, yes?

sgillies (approved):
@ryan-williams Great! I believe I see one required change.

src/tiledb/cloud/dag/dag.py
ryan-williams: yikes, yes, good catch. Looks like it failed some tests as well, thankfully.

sgillies: I'm glad we have a test that covers this! https://github.com/TileDB-Inc/TileDB-Cloud-Py/actions/runs/11842972791/job/33002846291?pr=680#step:12:411

ryan-williams:
Looks like `ConnectToExistingTest::test_running_graph` is failing on `main`, at least some of the time. [Example](https://github.com/TileDB-Inc/TileDB-Cloud-Py/actions/runs/11829336769/job/32960978779#step:12:263):
```
FAILED tests/taskgraphs/test_server_side.py::ConnectToExistingTest::test_running_graph - concurrent.futures._base.TimeoutError: timed out after 120 sec
```

sgillies:
@ryan-williams I'm confident that the failures are unrelated and will merge.

PR #683, opened 2024-11-18
Add a batch_mode flag to VCF scalable queries

gspowley:
Add a `batch_mode` arg to allow VCF scalable queries to run in batch mode instead of realtime mode.


leipzig (approved):

spencerseale:
Would `batch_mode=True` break the `resource_class` arg?

gspowley:
Good question. A quick test shows `resource_class` works for both `batch` and `realtime` UDFs.

PR #685, opened 2024-12-04
Fix credentials discrepancy between Bioimaging/VCF and SOMA registration UDFs

sgillies:


tests/vcf/test_vcf.py
```diff
@@ -1,26 +1,17 @@
...
 import tiledb.cloud.vcf as vcf
 import tiledb.cloud.vcf.vcf_toolbox as vtb
 
-# Pickle the vcf module by value, so tests run on the latest code.
-cloudpickle.register_pickle_by_value(vcf)
-
-
-# Run VCF tests with:
-#   pytest -m vcf --run-vcf -n 8
-
```
ktsitsi, line 14: Are we sure that we want to remove this one?  I was actually thinking the opposite of introducing it to all ingestors. It is a nice hack by @gspowley. I will add him as a reviewer.

ktsitsi (commented):
General: From the PR desc I understand that ideally we want the source's access configuration to be different from the destination access config. This would allow registering an asset that was accessed and ingested from a user's namespace to another namespace given we have access to both of them. I had implemented sth similar (if I get the problem right) here https://github.com/TileDB-Inc/TileDB-Cloud-Py/pull/607 feel free to check it out and if not relevant let's sync if I got sth wrong.

tests/vcf/test_vcf.py
gspowley: It looks like the changes in #683 are reverted in this PR. Maybe those changes need to be merged into this branch?

gspowley (changes requested):
Merge changes from #683.

gspowley (approved):
Thanks for merging those changes!

PR #691, opened 2024-12-17
Task graph decorators

spencerseale:
Make task graphs fun.

Manifesto: https://www.notion.so/Improve-Python-Task-Graph-Experience-1543265011758039b535e8bca4192c07 

* Starting with UDF
* Marking as draft to keep public


spencerseale:
@gspowley has something that's further along than this. Will discontinue progress on this branch. Closing

PR #692, opened 2024-12-30
Register DAG

spencerseale:
Adds functionality to register a `DAG` instance:

```python
from tiledb.cloud.dag import DAG

graph = DAG(name="dag-testing-tg-deps-2", namespace="spencerseale")

stagea = graph.submit(func=lambda x: x)
stageb = graph.submit(func=lambda x: x, args=(stagea,))
stagec = graph.submit(func=lambda x: x, args=(stagea,))
staged = graph.submit(func=lambda x, y: (x, y), args=(stageb, stagec,))

graph.register()

# OR

from tiledb.cloud.taskgraphs.registration import register

register(
    graph=graph,
    namespace="spencerseale"
)
```

The `DAG.register` method is to be prioritized, but also usable with `registration.register` as that's what's used in the `DAG.register` method.

### Open question about type hinting

I need to update type hint in `tiledb.cloud.taskgraphs.registration.register` to `graph: Union[tgbuilder, DAG]`. So I then add the import to that module `from tiledb.cloud.dag.dag import DAG`. But it's interesting, when I go to do that using my editable install and then run:

`python -m src.tiledb.cloud.dag.dag`

 I get: 
```
(tiledb-cloud-py-py39) spencerseale 12:27 PM $ TileDB-Cloud-Py python -m src.tiledb.cloud.dag.dag

  File "/Users/spencerseale/miniconda3/envs/tiledb-cloud-py-py39/lib/python3.9/runpy.py", line 188, in _run_module_as_main
    mod_name, mod_spec, code = _get_module_details(mod_name, _Error)
  File "/Users/spencerseale/miniconda3/envs/tiledb-cloud-py-py39/lib/python3.9/runpy.py", line 111, in _get_module_details
    __import__(pkg_name)
  File "/Users/spencerseale/tiledb_apis/TileDB-Cloud-Py/src/tiledb/cloud/__init__.py", line 3, in <module>
    from . import array
  File "/Users/spencerseale/tiledb_apis/TileDB-Cloud-Py/src/tiledb/cloud/array.py", line 7, in <module>
    from . import client
  File "/Users/spencerseale/tiledb_apis/TileDB-Cloud-Py/src/tiledb/cloud/client.py", line 13, in <module>
    import tiledb.cloud._common.api_v2.models as models_v2
ImportError: cannot import name 'cloud' from 'tiledb' (/Users/spencerseale/miniconda3/envs/tiledb-cloud-py-py39/lib/python3.9/site-packages/tiledb/__init__.py)
```

I get the same stack trace when I update the module `tiledb.cloud.taskgraphs.registration` to import anything. I think this is the way we have cloud-py wired up and how the Py system path is getting created. Switching to `poetry` or something similar would fix the dev experience a lot so the current dev library is installed into the system path as a Python library to simulate how cloud-py will execute as an installed library.

@sgillies any ideas on this?

### DIFF
--- a/src/tiledb/cloud/dag/dag.py
+++ b/src/tiledb/cloud/dag/dag.py
@@ -1,3 +1,5 @@
+"""Directed acyclic graphs as TileDB task graphs."""
+
 import collections
 import datetime
 import itertools
@@ -666,6 +668,7 @@ class Node(futures.FutureLike[_T]):
 
         :param existing_names: The set of names that have already been used,
             so that we don't generate a duplicate node name.
+        :return: Mapping of Node for registration.
         """
 
         return {
@@ -1859,42 +1862,25 @@ class DAG:
                 raise  # Bail out and fail loudly
 
     def _tdb_to_json(self, override_name: Optional[str] = None) -> Dict[str, Any]:
-        """Converts this DAG to a registerable/executable format.
+        """Converts this DAG to a registerable format.
 
         :param override_name: Name to override DAG conversion.
         :return: Mapping of DAG tree to json for submission to REST.
         """
 
-        # mapping of node name to node obj
-        nodes = self.nodes_by_name
+        nodes = _topo_sort_nodes(self.nodes)
 
-        existing_names = set(nodes.keys())
+        existing_names = set(self.nodes_by_name.keys())
 
-        node_jsons = [nodes[n].to_registration_json(existing_names) for n in nodes]
+        node_jsons = [n.to_registration_json(existing_names) for n in nodes]
 
-        for n, n_json in zip(nodes.values(), node_jsons):
-            # TODO, need to relace _deps with how deps are stored in DAG
-            n_json["depends_on"] = [
-                str(parent.id) for parent in self._deps.parents_of(n)
-            ]
+        for n, n_json in zip(nodes, node_jsons):
+            n_json["depends_on"] = [str(parent.id) for _, parent in n.parents.items()]
 
-        return dict(
-            name=override_name or self.name,
-            nodes=node_jsons,
-        )
-
-        # nodes = self._deps.topo_sorted
-        # # We need to guarantee that the existing node names are maintained.
-        # existing_names = set(self._by_name)
-        # node_jsons = [n.to_registration_json(existing_names) for n in nodes]
-        # for n, n_json in zip(nodes, node_jsons):
-        #     n_json["depends_on"] = [
-        #         str(parent.id) for parent in self._deps.parents_of(n)
-        #     ]
-        # return dict(
-        #     name=override_name or self.name,
-        #     nodes=node_jsons,
-        # )
+        return {
+            "name": override_name or self.name,
+            "nodes": node_jsons,
+        }
 
 
 def list_logs(
@@ -2169,16 +2155,3 @@ def array_task_status_to_status(status: models.ArrayTaskStatus) -> Status:
 
 def task_graph_log_status_to_status(status: models.TaskGraphLogStatus) -> Status:
     return _TASK_GRAPH_LOG_STATUS_TO_STATUS_MAP.get(status, Status.NOT_STARTED)
-
-
-if __name__ == "__main__":
-    # node = Node(func=lambda x: x)
-
-    # n = node._registration_name(existing=set(["a", "b"]))
-    # print(n)
-
-    graph = DAG()
-
-    graph.submit(func=lambda x: x)
-
-    print(graph.__dict__)
--- a/src/tiledb/cloud/dag/dag.py
+++ b/src/tiledb/cloud/dag/dag.py
@@ -27,6 +27,8 @@ from typing import (
     Union,
 )
 
+import tiledb
+
 from .. import array
 from .. import client
 from .. import rest_api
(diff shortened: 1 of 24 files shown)


src/tiledb/cloud/dag/dag.py
```diff
@@ -1801,7 +1862,56 @@ def _update_status(self) -> None:
...
+            n_json["depends_on"] = [str(parent.id) for _, parent in n.parents.items()]
+
+        return {
+            "name": override_name or self.name,
+            "nodes": node_jsons,
+        }
+
+    def register(
+        self,
+        override_name: Optional[str] = None,
```
sgillies, line 1890: Let's use just `name` here like we do in other methods.

src/tiledb/cloud/dag/dag.py
```diff
@@ -1801,7 +1862,56 @@ def _update_status(self) -> None:
...
+            )
+
+        registration.register(
+            graph=self,
+            name=tg_name,
+            namespace=self.namespace,
+        )
+
+        with tiledb.open(f"tiledb://{self.namespace}/{tg_name}", "w") as A:
+            A.meta["dataset_type"] = "registered_task_graph"
```
sgillies, line 1912: It would be nice if the registration endpoint could take metadata, then this would be more atomic.

src/tiledb/cloud/dag/dag.py
```diff
@@ -1801,7 +1862,56 @@ def _update_status(self) -> None:
...
+
+        if not tg_name:
+            raise ValueError(
+                "Must specify registration name to DAG.name or override_name."
+            )
+
+        registration.register(
+            graph=self,
+            name=tg_name,
+            namespace=self.namespace,
```
sgillies, line 1908: At first, I was wondering where all the new private methods were coming from, and then I realized you had copied them from the taskgraph module, and that they are called by the JSON serializer. I think this is an okay way to go. Could you look into another approach, to make sure that we're not missing out on a chance to do this with less code?

It occurs to me that `dag.compute()` uses our task graph builder to produce JSON that is sent to the server to be executed. Isn't this the same JSON that we would use when registering? Could your new `register()` method run `_build_batch_taskgraph()`, for example, to produce the JSON?

If this is a dumb or not workable idea, I will gladly drop it.

sgillies (commented):
@spencerseale I've got a suggestion and a question inline.

sgillies:
@spencerseale if registration.py imports dag.py and dag.py imports registration.py we have a circular import, no? Let's avoid doing that, especially if only to have perfect type hints. I think it's okay to relax the typing for `registration.register()` and use `graph: Any` for now.

But if we do want better type hints without a circular import... `registration.register()` can actually accept any argument that has a `_tdb_to_json()` method. So, we could create an abstract base class with that method and register `DAG` with the ABC, or we could use Python structural typing.

sgillies:
@spencerseale about test failures: tiledb versions 0.33.1 and 0.33.2 broke the `CloudArray` interface. If you see errors like `AttributeError: 'SparseArrayImpl' object has no attribute 'apply'` (for example) in our autobuild jobs, they aren't caused by your changes.

src/tiledb/cloud/dag/dag.py
sgillies: @spencerseale I have a better idea, one that will make new modules and new interfaces unnecessary for now. Instead of importing `tiledb.cloud.taskgraphs.registration` (which tangles our modules together) and calling `registration.register()`, copy the body of `registration.register()` into `Dag.register()`.

A little bit of code duplication, it's only two copies, doesn't bother me at all.

tests/test_dag.py
```diff
@@ -1350,6 +1353,38 @@ def _b64(x: bytes) -> str:
     return str(base64.b64encode(x), encoding="ascii")
 
 
+@pytest.fixture
+def dag_fixture():
+    """DAG fixture for pytests."""
+
+    graph = dag.DAG(name="dag-test-fixture", namespace=default_user().username)
```
sgillies, line 1360: @spencerseale I believe we will need a globally unique name for the DAG since the new test might be run at the same time for two different Python versions.

tests/test_dag.py
```diff
@@ -1350,6 +1353,38 @@ def _b64(x: bytes) -> str:
...
+    dag_fixture: dag.DAG,
+) -> None:
+    """Test DAG.register"""
+
+    # verify DAG.name used to register
+    registered_name1 = dag_fixture.register()
+    assert registered_name1 == f"{dag_fixture.namespace}/{dag_fixture.name}"
+
+    # verify override name
+    registered_name2 = dag_fixture.register(name="override-name")
```
sgillies, line 1379: We'll need a globally unique name for registration here, too.

tests/test_dag.py
```diff
@@ -1350,6 +1353,38 @@ def _b64(x: bytes) -> str:
...
+    assert registered_name2 == f"{dag_fixture.namespace}/override-name"
+
+    # verify catch if no name set to DAG.name or override_name
+    dag_fixture.name = None
+    with pytest.raises(ValueError):
+        dag_fixture.register()
```
sgillies, line 1385: Finally, should we ensure that the registered DAGs are deleted?

sgillies (changes requested):
@spencerseale looking good! I have a couple recommendations.

sgillies:
@spencerseale I think this will need to be rebased onto main as well (the wheel test, for example).

tests/test_dag.py
spencerseale: Good point and i'll try and be more aware of that in the future!

tests/test_dag.py
spencerseale: The registration is mocked, so nothing is actually being registered. The api model is assumed tested elsewhere, so this unittest just tests the wrapper around the `rest_api.RegisteredTaskGraphsApi`

sgillies (approved):
Great @spencerseale !

PR #695, opened 2025-01-17
Add a udf mark, to skip tests that can't pass without a 3.12 image

sgillies:
"udf" is short for "requires server-side code execution". I'm marking a few modules and select tests within a few other modules.

The code in this project appears to be 99.9% fine with a new Python version. I only had to replace a unittest method that was removed in Python 3.11.


.github/workflows/tiledb-cloud-py.yaml
```diff
@@ -160,6 +160,7 @@ jobs:
             --splitting-algorithm least_duration \
             --clean-durations \
             -r fExX \
+            ${{ matrix.python-version == '3.12' && '-m "not udf"' || '' }} \
```
sgillies, line 163: We have no Python 3.12 images in production. The udf mark allows us to skip the many tests that rely on server side code execution.

ci/requirements.txt
```diff
@@ -10,7 +10,7 @@ python-dateutil==2.8.2
 pytz==2024.1
 six==1.16.0
 tblib==1.7.0
-tiledb==0.25.0
+tiledb==0.32.5
```
sgillies, line 13: 0.32 is the current production version.

pyproject.toml
```diff
@@ -50,6 +51,7 @@ explicit-only = ["bigfiles", "geospatial", "vcf"]
 markers = [
     "bigfiles: tests that create and upload really big files",
     "geospatial: tests that require the geospatial libraries",
+    "udf: tests that require a server UDF container",
```
sgillies, line 54: Here's the new mark.

tests/common/test_pickle_compat.py
```diff
@@ -58,6 +58,10 @@ def test_pandas_compat(pd_ver: str, name_want: Tuple[str, Callable[[], Any]]) ->
     assert want.equals(got)
 
 
+@pytest.mark.skipif(
+    sys.version_info >= (3, 12, 0),
+    reason="serialization is now controlled by vendored cloudpickle frozen at 2.1.0.",
+)
```
sgillies, line 64: Now that we completely control the versions on both ends, we don't need to run these tests for every Python version.

ktsitsi (approved):

PR #696, opened 2025-01-22
Exclude tiledb-py versions 0.33.0-0.33.2

sgillies:
I successfully ran a SOMA ingestion task graph using this branch on the client side + tiledb-py 0.33.3.rc0. The combination are compatible with our current production versions.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ dependencies = [
     # Not directly used on the client, but some server-side environments have
     # tblib transitively enabled, so this is needed for unpickling.
     "tblib~=1.7",
-    "tiledb>=0.33.3.rc0",
+    "tiledb~=0.25,!=0.33.0,!=0.33.1,!=0.33.2",
     "typing-extensions",
     "urllib3>=1.26",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ dependencies = [
     # Not directly used on the client, but some server-side environments have
     # tblib transitively enabled, so this is needed for unpickling.
     "tblib~=1.7",
-    "tiledb~=0.25,!=0.33.0,!=0.33.1,!=0.33.2",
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -18,7 +18,7 @@ dependencies = [
     # Not directly used on the client, but some server-side environments have
     # tblib transitively enabled, so this is needed for unpickling.
     "tblib~=1.7",
-    "tiledb>=0.15.2",
+    "tiledb~=0.25,!=0.33.0,!=0.33.1,!=0.33.2",
```
sgillies, line 21: I'm not entirely sure what the lower band should be 🤔 

0.15.2 is too low, it doesn't immediately install on a Mac. Our CI tests pass with 0.25 on Linux, so that seems like a good candidate?

sgillies: Nevermind, we require 0.30, the first version to provide `ls_recursive()`.

Shelnutt2:
Can we get some details on the compatibility issues here? Seeing several version of TileDB-Py being blocked here is likely to cause some headaches for packaging and users. What are the incompatible changes? Why are we having these issues?

cc: @ihnorton

kounelisagis:
> Can we get some details on the compatibility issues here? Seeing several version of TileDB-Py being blocked here is likely to cause some headaches for packaging and users. What are the incompatible changes? Why are we having these issues?
> 
> cc: @ihnorton

This was caused by: https://github.com/TileDB-Inc/TileDB-Py/pull/2123/commits/aa995974214a47d43864b0f18bcf83ab8b50d5f4
and fixed by: https://github.com/TileDB-Inc/TileDB-Py/pull/2144

sgillies:
@Shelnutt2 Just as @kounelisagis said. tiledb-py module attributes were removed in 0.30.0 and this broke our UDF serialization.

This PR starts to broaden our CI matrix so that we may see such issues sooner.

ihnorton (approved):

ktsitsi (approved):

PR #697, opened 2025-01-23
Add tiledb main branch to CI matrix

sgillies:
With this PR, we now test tiledb-cloud against 1) the minimum supported version of tiledb(-py), 2) the most recent supported release on PyPI, and 3) the main branch of the tiledb-py project.

In the first case, we `pip install tiledb==0.30` and, after, `pip install .[tests]` determines that 0.30.0 meets the requirements and uses it.

In the second case, we `pip install tiledb`, get 0.33.2, and then `pip install .[tests]` determines that 0.33.2 is excluded, and then replaces it with a new install of 0.33.0 from PyPI.

In the third case, we install tiledb-py's main branch from GitHub, and `pip install .[tests]` determines that it meets the requirements.

The current test failures are due to timeouts on the server side. I'll rerun and they'll pass.

### DIFF
--- a/.github/workflows/tiledb-cloud-py.yaml
+++ b/.github/workflows/tiledb-cloud-py.yaml
@@ -47,6 +47,9 @@ jobs:
           - ubuntu-20.04
           # - macos-12
         python-version: ["3.9", "3.12"]
+        is-scheduled:
+          - ${{ github.event.schedule == "34 1 * * *" }}
+          - ${{ github.event.schedule != "34 1 * * *" }}
         tiledb-version-spec:
           [
             "tiledb==0.30",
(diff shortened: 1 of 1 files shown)


.github/workflows/tiledb-cloud-py.yaml
```diff
@@ -37,6 +37,7 @@ jobs:
         run: pre-commit run -a -v
 
   run-tests:
+    name: "${{ matrix.tiledb-version-spec }}:${{ matrix.python-version }}:${{ matrix.dependencies }}:${{ matrix.os }}"
```
sgillies, line 40: So that the tiledb version appears first in the limited width left sidebar on, for example, https://github.com/TileDB-Inc/TileDB-Cloud-Py/actions/runs/12937321533.

ihnorton (approved):
g2g after moving to nightly as discussed. Thanks!

.github/workflows/tiledb-cloud-py.yaml
```diff
@@ -55,6 +63,15 @@ jobs:
             path: ~/.cache/pip
           # - os: macos-12
           # path: ~/Library/Caches/pip
+        exclude:
+          - tiledb-version-spec: "tiledb==0.30"
+            dependencies: fresh
+          - tiledb-version-spec: "tiledb==0.30"
+            is-pr: true
```
sgillies, line 70: Builds using the oldest supported tiledb are excluded for unscheduled PR builds.

.github/workflows/tiledb-cloud-py.yaml
```diff
@@ -55,6 +63,15 @@ jobs:
...
+          - tiledb-version-spec: "tiledb@git+https://github.com/TileDB-Inc/TileDB-Py.git@main"
+            dependencies: pinned
+          - tiledb-version-spec: "tiledb@git+https://github.com/TileDB-Inc/TileDB-Py.git@main"
+            is-pr: true
```
sgillies, line 74: Builds using the latest unpublished tiledb are also excluded for unscheduled PR builds.

PR #698, opened 2025-01-23
Workflow register, run, and resume

gspowley:
Add functions to register, run, and resume Nextflow workflows.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,6 +10,7 @@ dependencies = [
     "attrs>=21.4.0",
     "certifi",
     "importlib-metadata",
+    "jsonschema",
     "packaging",
     "pandas>=1.2.4",
     "pyarrow>=3.0.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,6 +37,7 @@ tests = [
     "pytest-split",
     "tiledbsoma",
     "xarray",
+    "nextflow",
 ]
 
 [project.urls]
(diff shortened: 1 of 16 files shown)


sgillies (commented):
@gspowley how long do the tests take, as a fraction of our current time? Do we want more tests? Is the coverage good?

The main thing I was looking for is that we don't import this into the top level of tiledb.cloud like we've (mistakenly, in my opinion) done for other modules like https://github.com/TileDB-Inc/TileDB-Cloud-Py/blob/main/src/tiledb/cloud/__init__.py#L3. So, I'm happy 👍

gspowley:
> how long do the tests take, as a fraction of our current time? Do we want more tests? Is the coverage good?

Running all tests sequentially takes ~12 minutes, which is ~4x a test partition. However, the run related tests will not work until 3.0 is released. I plan to add more tests and can check the coverage.

Thanks for taking a look. Any early feedback is appreciated.

leipzig (approved):
excellent. let's get this merged.

gspowley:
@sgillies any objections to merging this PR? Almost all changes are related to the new `tiledb.cloud.workflows` subpackage.

sgillies (approved):
👍 

@gspowley @leipzig when I ran mypy on the new code, I saw a few errors. These three stand out to me as being worth fixing. We might prevent str/bytes confusion in one case and should raise a specific error type in another.

```
src/tiledb/cloud/workflows/nextflow/history.py:129: error: Incompatible types in assignment (expression has type "bytes", target has type "str")  [assignment]
src/tiledb/cloud/workflows/nextflow/run.py:168: error: Exception must be derived from BaseException  [misc]
src/tiledb/cloud/workflows/nextflow/run.py:214: error: Function "builtins.callable" is not valid as a type  [valid-type]
src/tiledb/cloud/workflows/nextflow/run.py:214: note: Perhaps you meant "typing.Callable" instead of "callable"?
```

PR #699, opened 2025-01-27
Remove dest_config from kwargs

ktsitsi:
Desc:
    The bioimg ingestion function, receives a `multiple argument error` when `dest_config` is set.
Reason:
    `dest_config` is not being removed from `kwargs` and this way is being carried over in the upcoming function calls

This PR:

- Fixes the issue by `pop` instead of `get` the value of `dest_config` from the `kwargs`


sgillies (approved):
🚀

PR #701, opened 2025-01-28
Don't use https for non-SSL connections

ihnorton:
Defaulting to https when `verify_ssl=False` has caused confusion several times when working in a development environment.

(please feel free to adjust as needed)

### DIFF
--- a/src/tiledb/cloud/client.py
+++ b/src/tiledb/cloud/client.py
@@ -106,11 +106,11 @@ def login(
     # scheme, but our SDK strictly requires the http(s) scheme.
     elif not host.startswith(("http://", "https://")):
         if verify_ssl:
-          host = f"https://{host}"
+            host = f"https://{host}"
         else:
-          # Don't use https:// for non-SSL connections
-          # (defaulting to https:// when user sets verify_ssl=False can be very confusing!)
-          host = f"http://{host}"
+            # Don't use https:// for non-SSL connections
+            # (defaulting to https:// when verify_ssl=False is confusing)
+            host = f"http://{host}"
 
     config_args = {
         "username": username,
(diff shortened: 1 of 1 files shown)


src/tiledb/cloud/client.py
```diff
@@ -107,6 +100,18 @@ def login(
...
+    if host is None:
+        host = config.default_host
+    # See sc-56351. Usually, a hostname doesn't include a protocol
+    # scheme, but our SDK strictly requires the http(s) scheme.
+    elif not host.startswith(("http://", "https://")):
+        if verify_ssl:
+          host = f"https://{host}"
+        else:
+          # Don't use https:// for non-SSL connections
+          # (defaulting to https:// when user sets verify_ssl=False can be very confusing!)
```
Shelnutt2, line 112: `verify_ssl=False` is useful when you have certificates that are not recognized by the client, such as self signed certs. This variable does not mean "no ssl".

src/tiledb/cloud/client.py
```diff
@@ -107,6 +100,18 @@ def login(
...
+        host = config.default_host
+    # See sc-56351. Usually, a hostname doesn't include a protocol
+    # scheme, but our SDK strictly requires the http(s) scheme.
+    elif not host.startswith(("http://", "https://")):
+        if verify_ssl:
+            host = f"https://{host}"
+        else:
+            # Don't use https:// for non-SSL connections
+            # (defaulting to https:// when verify_ssl=False is confusing)
+            host = f"http://{host}"
```
sgillies, line 113: @ihnorton in light of @Shelnutt2's comment, perhaps we do this below? Assuming that our own review apps and staging redirect to HTTPS.

```suggestion
    elif not host.startswith(("http://", "https://")):
        host = f"http://{host}"
```

sgillies (commented):
Suggestion made inline.

PR #702, opened 2025-02-04
CI updates and noise reduction

sgillies:
This moves test job conditions to matrix exclusion rules to eliminate no-op jobs from GitHub actions reports and xfails several tests when they use unsupported numpy 2 in UDFs.

### DIFF
--- a/.github/workflows/tiledb-cloud-py.yaml
+++ b/.github/workflows/tiledb-cloud-py.yaml
@@ -72,6 +72,8 @@ jobs:
             dependencies: pinned
           - tiledb-version-spec: "tiledb@git+https://github.com/TileDB-Inc/TileDB-Py.git@main"
             is-pr: true
+          # - dependencies: fresh
+          #   is-pr: true
         pytest-split-group: [1, 2, 3, 4, 5, 6] # Range the parallel test groups defined by env.PYTEST_SPLIT_GROUPS
 
     steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/tiledb-cloud-py.yaml
```diff
@@ -72,6 +72,8 @@ jobs:
             dependencies: pinned
           - tiledb-version-spec: "tiledb@git+https://github.com/TileDB-Inc/TileDB-Py.git@main"
             is-pr: true
+          - dependencies: fresh
+            is-pr: true
```
sgillies, line 76: This has been the intent of our CI, and this is a cleaner way to do it.

.github/workflows/tiledb-cloud-py.yaml
```diff
@@ -159,16 +161,12 @@ jobs:
           fi
 
       - name: Run tests for vendored cloudpickle
-        # Test pinned dependencies on commit / tag and fresh installs on nightlies
-        if: (matrix.dependencies == 'fresh') == (github.event_name == 'schedule' || github.event_name == 'workflow_dispatch')
```
sgillies, line 163: This check is no longer needed with our new matrix exclusion rule.

tests/test_dag.py
```diff
@@ -918,6 +918,9 @@ def test_dag_sql_exec(self):
 
         self.assertEqual(node.result()["a"][0], numpy.sum(orig["a"]))
 
+    @pytest.mark.xfail(
+        np.__version__.startswith("2"), reason="Numpy 2 is not allowed in UDFs"
+    )
```
sgillies, line 923: We still have numpy 1.25 on the server side, so tests with numpy >= 2.0 on the client will fail.

spencerseale: This is a great safety feature

spencerseale (approved):
Thanks Sean!

tests/test_dag.py
sgillies: We *do* want to check most parts of this package against the latest numpy, but there are specific tests that can't pass right now.

PR #703, opened 2025-02-05
Exclude more unpredictable tests

sgillies:
This PR xfails a task graph test with an unpredictable finishing time, and skips two wheel upload tests that are too flaky.

### DIFF
--- a/tests/taskgraphs/test_server_side.py
+++ b/tests/taskgraphs/test_server_side.py
@@ -19,6 +19,7 @@ _WAIT_TIME_S = 120
 
 
 class ConnectToExistingTest(unittest.TestCase):
+    @pytest.mark.xfail(reason="Time to complete is not predictable.")
     def test_completed_graph(self) -> None:
         to_run = dag.DAG(mode=dag.Mode.BATCH)
         one = to_run.submit(lambda: 1, name="one")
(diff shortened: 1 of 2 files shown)


tests/taskgraphs/test_server_side.py
```diff
@@ -19,6 +19,7 @@
 
 
 class ConnectToExistingTest(unittest.TestCase):
+    @pytest.mark.xfail(reason="Time to complete is not predictable.")
```
sgillies, line 22: This test is once again not completing in time. I don't think that increasing the wait is a good strategy, it delays all our builds.

tests/utilities/test_wheel.py
```diff
@@ -69,6 +69,7 @@ def array_teardown():
     tiledb.Array.delete_array(_FULL_URI, ctx=ctx)
 
 
+@pytest.mark.skip(reason="Race condition in tests.")
```
sgillies, line 72: These wheel tests fail ~20% of the time. They can't be xfailed, they must be skipped because the bug is in the `array_teardown` fixture.

ktsitsi (approved):

sgillies:
@spencerseale @gspowley we can circle back to the wheel tests after other work settles down.

PR #704, opened 2025-02-06
Fix error when passing `output_array_name` to `tiledb.cloud.sql.exec`

nickvigilante:
Based on this discussion: https://tiledb.slack.com/archives/C5RLSKUF9/p1738880642359909


sgillies (approved):
Thanks for linking to the discussion with Spencer!

PR #705, opened 2025-02-13
Expose get_group_contents() via high level API

sgillies:
With tests.


spencerseale (approved):
Thanks Sean, this is a great addition!

PR #707, opened 2025-02-13
Provide a default name for arrays when registering

sgillies:
In exactly the same way we do for groups.

### DIFF
--- a/src/tiledb/cloud/array.py
+++ b/src/tiledb/cloud/array.py
@@ -1,3 +1,7 @@
+"""Register, search, and manage arrays with TileDB."""
+
+import posixpath
+import urllib.parse
 import uuid
 import warnings
 from typing import Any, Callable, Iterable, List, Optional, Sequence, Union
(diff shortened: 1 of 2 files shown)


src/tiledb/cloud/array.py
```diff
@@ -237,12 +241,16 @@ def register_array(
...
+    :param str namespace: The user or organization to register the array
+        under. If unset will default to the user
+    :param str array_name: name of array, optional. Defaults to stem of
+        the storage URI.
     :param str description: optional description
-    :param str access_credentials_name: optional name of access credentials to use,
-        if left blank default for namespace will be used
+    :param str access_credentials_name: optional name of access
+        credentials to use. If left blank default for namespace will be
+        used.
```
sgillies, line 253: A little documentation and grammar cleanup.

src/tiledb/cloud/array.py
```diff
@@ -1,3 +1,7 @@
+"""Register, search, and manage arrays with TileDB."""
```
sgillies, line 1: I'm adding module docstrings as I go. These will show up in our docs.

spencerseale (approved):
Thanks Sean! Sorry I didn't get to this...

sgillies:
The error is unrelated. I'm addressing it in a separate PR.

PR #708, opened 2025-02-13
Avoid udf name contention between different CI jobs

sgillies:
This crops up recently because we're running that particular test twice at the same time only after expanding the CI build matrix.

### DIFF
--- a/tests/test_asset.py
+++ b/tests/test_asset.py
@@ -221,7 +221,9 @@ class ListingTest(unittest.TestCase):
     @pytest.mark.udf
     def test_list(self) -> None:
         # Ensure that we have at least one asset registered (a UDF)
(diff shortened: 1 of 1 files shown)


tests/test_asset.py
```diff
@@ -221,7 +221,9 @@ class ListingTest(unittest.TestCase):
     @pytest.mark.udf
     def test_list(self) -> None:
         # Ensure that we have at least one asset registered (a UDF)
-        with testonly.register_udf(lambda: 1, func_name="some_lambda"):
+        with testonly.register_udf(
+            lambda: 1, func_name=testonly.random_name("some_lambda")
+        ):
```
sgillies, line 226: This is follows usage in other related tests.

spencerseale (approved):

PR #711, opened 2025-02-20
Remove upper limit for tblib requirement spec

### LINKED ISSUE #709: is this compatible with tblib above 1.7.0?
Seems to not be, yet it's at 3.0 already...

tblib==1.7.0
  - tiledb-cloud==0.13.0 [requires: tblib~=1.7]
    - tiledb-vector-search==0.11.0 [requires: tiledb-cloud>=0.11]

sgillies:
Resolves #709.

See also #405.

I've confirmed that tblib version 3.0 can unpickle data pickled by tblib >= 1.7.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,7 +17,7 @@ dependencies = [
     "six>=1.10",
     # Not directly used on the client, but some server-side environments have
     # tblib transitively enabled, so this is needed for unpickling.
-    "tblib",
+    "tblib>=1.7",
     "tiledb~=0.30,!=0.33.1,!=0.33.2",
     "typing-extensions",
     "urllib3>=1.26",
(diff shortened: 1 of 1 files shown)


ktsitsi (approved):

pyproject.toml
```diff
@@ -17,7 +17,7 @@ dependencies = [
     "six>=1.10",
     # Not directly used on the client, but some server-side environments have
     # tblib transitively enabled, so this is needed for unpickling.
-    "tblib~=1.7",
+    "tblib",
```
ihnorton, line 20: Based on the timeline, I think we need to treat `1.7` as a lower bound.
```suggestion
    "tblib>=1.7",
```

ihnorton (approved):
LGTM w/ the lower bound suggestion inline.

PR #712, opened 2025-02-22
Permit null concat with non-null in distributed vcf query

spencerseale:
https://app.shortcut.com/tiledb-inc/story/63653/permit-null-concat-with-non-null-in-distributed-vcf-query

Defaulting to False for now, but eventually should be considered standard and removed from being an arg. 

We currently support pyarrow 11.0 and the arg is `promote`: https://arrow.apache.org/docs/11.0/python/generated/pyarrow.concat_tables.html

But the stable pyarrow versions use: `promote_options`:
https://arrow.apache.org/docs/python/generated/pyarrow.concat_tables.html

Not sure how we should handle that as when we bump, it will break. 11.0 is so old i'm not sure if concat accepts `promote` to be backwards compatible.

### DIFF
--- a/src/tiledb/cloud/vcf/query.py
+++ b/src/tiledb/cloud/vcf/query.py
@@ -83,6 +83,7 @@ def vcf_query_udf(
     memory_budget_mb: int = 1024,
     af_filter: Optional[str] = None,
     transform_result: Optional[Callable[[pa.Table], pa.Table]] = None,
+    promote_null: bool = False,
     log_uri: Optional[str] = None,
     log_id: str = "query",
     verbose: bool = False,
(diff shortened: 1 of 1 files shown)


src/tiledb/cloud/vcf/query.py
```diff
@@ -83,6 +83,7 @@ def vcf_query_udf(
     memory_budget_mb: int = 1024,
     af_filter: Optional[str] = None,
     transform_result: Optional[Callable[[pa.Table], pa.Table]] = None,
+    promote_null: bool = False,
```
sgillies, line 86: @spencerseale Optionally, we could just call it `promote` like Pyarrow does. I'm not opposed to `promote_null`.

sgillies (approved):

src/tiledb/cloud/vcf/query.py
spencerseale: was trying to be more neutral between the `promote` and `promote_options` discrep in pyarrow versions.

gspowley (approved):
Thanks @spencerseale!

spencerseale:
Added new `_concat_tables` and unittests

sgillies (approved):
@spencerseale Excellent!

PR #721, opened 2025-04-30
Resolve sc-49443 for workaround on member-list performance

johnkerl:
This is an old follow-up from [sc-49443](https://app.shortcut.com/tiledb-inc/story/49443/performance-issue-in-soma-group-member-retrieval), as descrbed in the comments in the affected code.


sgillies (approved):
👍 @johnkerl

PR #723, opened 2025-06-14
Add support for VCF ingestion failure

alancleary:
This PR updates VCF ingestion to support resuming failed ingestions. This is done by using a new "ready" status in the VCF manifest array.

**Previous ingestion process:**
1. Given a set of VCF URIS, add URIs that aren't already in the manifest
    - Samples that are ready to be ingested should now have `"ok"` or `"missing index"`
2. Ingest URIs from the manifest array with status `"ok"` or `"missing index"` that aren't already in the dataset

Notes:
- This process prevents URIs from ever being ingested if they don't have the `"ok"` or `"missing index"` status when added to the manifest
- This process prevents resuming ingestions that got far enough to add samples to the dataset but then failed

**New ingestion process:**
1. Given a set of VCF URIs, add URIs that aren't already in the manifest
    - Samples that are ready to be ingested should now have `"ready"` or `"missing index"`
2. Ingest URIs from the manifest array with status `"ready"` or `"missing index"`
    - If the sample for a URI is already in the dataset, it will only be ingested if the `resume` ingestion parameter is `True`
3. change sample status to `"ok"` after successful ingestion

Notes:
- This workflow allows resuming partial/failed ingestions
- This workflow is backwards compatible with existing manifests
    - Previously ingested entries with status `"ok"` are assumed to be ingested
    - Previously ingested entries with status `"missing index"` can be "resumed" and their status will changed to `"ok"`
    - Previously failed ingestions with status `"missing index"` that got far enough to have an entry in the dataset can now be resumed

**Edge cases:**
- Samples already in the manifest with status `"ok"` whose ingestion failed after the sample was added to the dataset will need to be manually removed from the manifest
  - There is no simple way to detect these samples programmatically so manual intervention is required

### DIFF
--- a/src/tiledb/cloud/vcf/ingestion.py
+++ b/src/tiledb/cloud/vcf/ingestion.py
@@ -80,7 +80,8 @@ class Status(str, enum.Enum):
     """
     The ingestion status of samples in the manifest.
 
-    OK = the VCF file can be ingested
+    READY = the VCF file can be ingested
+    OK = the VCF file was successfully ingested
     MISSING_INDEX = the VCF file does not have a corresponding index file
     MISSING_SAMPLE_NAME = the VCF file does not have a sample name
     MULTIPLE_SAMPLES = the VCF file has multiple sample names
@@ -88,6 +89,7 @@ class Status(str, enum.Enum):
     BAD_INDEX = the VCF index file could not be properly read
     """
 
+    READY = "ready"
     OK = "ok"
     MISSING_INDEX = "missing index"
     MISSING_SAMPLE_NAME = "missing sample name"
@@ -611,6 +613,7 @@ def filter_samples_udf(
     dataset_uri: str,
     *,
     config: Optional[Mapping[str, Any]] = None,
+    resume: bool = True,
     verbose: bool = False,
 ) -> Sequence[str]:
     """
@@ -618,6 +621,7 @@ def filter_samples_udf(
 
     :param dataset_uri: dataset URI
     :param config: config dictionary, defaults to None
+    :param resume: enable resume ingestion mode, defaults to True
     :param verbose: verbose logging, defaults to False
     :return: sample URIs
     """
@@ -633,29 +637,38 @@ def filter_samples_udf(
                 dataset_uri,
                 cfg=tiledbvcf.ReadConfig(tiledb_config=config),
             )
-            existing_samples = set(ds.samples())
+            dataset_samples = set(ds.samples())
 
-            # Read all samples in the manifest with status == "ok"
+            # Read all samples in the manifest and samples ready for ingestion
             group = tiledb.Group(dataset_uri)
             manifest_uri = group[MANIFEST_ARRAY].uri
 
             with tiledb.open(manifest_uri) as A:
-                manifest_df = A.query(
-                    cond=f"status == '{Status.OK}' or status == '{Status.MISSING_INDEX}'"
+                manifest_df = A.df[:]
+                ingest_df = A.query(
+                    cond=f"status == '{Status.READY}' or status == '{Status.MISSING_INDEX}'"
                 ).df[:]
+            manifest_samples = set(manifest_df.sample_name)
 
             # Sort manifest by sample_name
-            manifest_df = manifest_df.sort_values(by=["sample_name"])
+            ingest_df = ingest_df.sort_values(by=["sample_name"])
+            ingest_samples = set(ingest_df.sample_name)
 
-            # Find samples that have not already been ingested
-            manifest_samples = set(manifest_df.sample_name)
-            new_samples = manifest_samples.difference(existing_samples)
-            manifest_df = manifest_df[manifest_df.sample_name.isin(new_samples)]
-            result = manifest_df.vcf_uri.to_list()
+            # Find samples that have already been ingested or failed to ingest
+            existing_samples = dataset_samples.difference(ingest_samples)
+            failed_samples = dataset_samples.intersection(ingest_samples)
 
+            # Finalize samples to ingest
+            if not resume:
+                ingest_df = ingest_df[~ingest_df.sample_name.isin(failed_samples)]
+            result = ingest_df.vcf_uri.to_list()
+
+            logger.info("%d samples in the dataset.", len(dataset_samples))
+            logger.info("%d samples fully ingested.", len(existing_samples))
+            logger.info("%d samples partially ingested.", len(failed_samples))
             logger.info("%d samples in the manifest.", len(manifest_samples))
-            logger.info("%d samples already ingested.", len(existing_samples))
-            logger.info("%d new samples to ingest.", len(result))
+            logger.info("%d samples ready to be ingested.", len(ingest_samples))
+            logger.info("%d samples will be ingested.", len(result))
             prof.write("count", len(result))
 
         return result
@@ -699,7 +712,7 @@ def ingest_manifest_udf(
                 values = defaultdict(list)
 
                 for vcf_uri in sample_uris:
-                    status = Status.OK
+                    status = Status.READY
 
                     # Check for sample name issues
                     try:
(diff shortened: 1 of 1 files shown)


spencerseale (approved):
Looks good! Thanks @alancleary! 

I think the edge case won't be too big of a concern, unless someone did not realize their dataset failed ingestions. I don't think much of that is out there in our customers thankfully!

leipzig:
Can we get some clarification on the behavior when people try to ingest the same sample split into multiple chromosomes?

alancleary:
> Can we get some clarification on the behavior when people try to ingest the same sample split into multiple chromosomes - both during the same ingest attempt and in staggered attempts?
> 
> No expectations here but this is one case where even an unfair policy is better than an unclear or inconsistent one.

This PR preserves the existing behavior, i.e. a sample name cannot be ingested from more than one VCF. Implementation details aside, this is basically because once a sample is in the dataset you can't tell which VCF it came from.

I'll be sure to update the documentation to convey this, as we discussed.

jp-dark:
Do we need to worry about duplicate data if there is a failure between successfully writing a sample to the dataframe and updating the status in the manifest to `"ok"`?

alancleary:
> Do we need to worry about duplicate data if there is a failure between successfully writing a sample to the dataframe and updating the status in the manifest to `"ok"`?

No. If `resume=False` then the data won't be loaded because it's already in the dataset. If `resume=True` then the underlying call TileDB-VCF ingest will be run with `resume=True` as well, resulting in no data duplication. I've tested the latter.

src/tiledb/cloud/vcf/ingestion.py
```diff
@@ -611,29 +635,37 @@ def filter_samples_udf(
...
+            # Read all samples in the manifest and samples ready for ingestion
             group = tiledb.Group(dataset_uri)
             manifest_uri = group[MANIFEST_ARRAY].uri
 
+            cond = f"status == '{Status.READY}' or status == '{Status.MISSING_INDEX}'"
             with tiledb.open(manifest_uri) as A:
-                manifest_df = A.query(
-                    cond="status == 'ok' or status == 'missing index'"
-                ).df[:]
+                manifest_df = A.df[:]
```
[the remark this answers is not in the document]
alancleary: I haven't benchmarked this but I'm guessing not. The same pattern is used [when filtering the URIs to ingest](https://github.com/TileDB-Inc/TileDB-Cloud-Py/blob/1cdbb7834aa005d1dbb59b8585d6ee4b89f4426f/src/tiledb/cloud/vcf/ingestion.py#L596), which could be done with a `Query` to avoid loading the entire manifest into memory. Also, I think in practice manifests will be no larger than hundreds of thousands of entries (someone please correct me if this is wrong).

That said, we could revise this to only fetch these data if the logger is set to a level that will report it, or just omit it altogether.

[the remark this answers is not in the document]
Shelnutt2: > Also, I think in practice manifests will be no larger than hundreds of thousands of entries (someone please correct me if this is wrong).

Your generally correct. We've got a couple use cases in 1-10M sample range but 100k-1M is typically the upper bound for a lot of use cases. Usually when there are more samples involved multiple datasets start coming into play. Its not a TileDB limitation but more often geographic boundaries and/or other compliance or regulations tend for push users to split the datasets up and you never end up with 100s of millions of samples in a single TIleDB-VCF data.

src/tiledb/cloud/vcf/ingestion.py
```diff
@@ -611,29 +635,37 @@ def filter_samples_udf(
...
+            # Find samples that have already been ingested or failed to ingest
+            existing_samples = dataset_samples.difference(ingest_samples)
+            failed_samples = dataset_samples.intersection(ingest_samples)
+
+            # Finalize samples to ingest
+            if not resume:
+                ingest_df = ingest_df[~ingest_df.sample_name.isin(failed_samples)]
+            result = ingest_df.vcf_uri.to_list()
 
+            logger.info("%d samples in the dataset.", len(dataset_samples))
```
sgillies, line 663: @alancleary what would you think about making the log messages more structured. I don't mean using [structlog](https://www.structlog.org/en/stable/) necessarily, but changing the messages from being prose to being `event` + `data`. For example: `logger.info("Ingestion iteration completed: successful_samples=%r, failed_samples=%r", ...)`. Everything on a single line.

Messages like these are easy to scan for and can be more easily parsed and turned into data.

I know that's not the prevailing style in our Python code, but I'm trying to nudge new work in the structured direction.

src/tiledb/cloud/vcf/ingestion.py
```diff
@@ -707,14 +739,14 @@ def file_size(uri: str) -> int:
                     # Check for index issues
                     index_uri = find_index(vcf_uri)
                     if not index_uri:
-                        status = "" if status == "ok" else status + ","
-                        status += "missing index"
+                        status = "" if status == Status.READY else status + ","
+                        status += Status.MISSING_INDEX
```
sgillies, line 743: Seeing this, I wonder if status shouldn't be a stack (list) instead of a string of comma-separated values.

sgillies (changes requested):
@alancleary I made one request to structure the logging and made a minor suggestion about the state of ingestion that may or may not be useful to you.

src/tiledb/cloud/vcf/ingestion.py
alancleary: Thanks for the suggestion @sgillies. That makes sense to me. I'll make the change.

src/tiledb/cloud/vcf/ingestion.py
alancleary: Yeah, I don't think string is quite right here. I would like to revisit the whole manifest at some point but this PR is just a bug fix!

src/tiledb/cloud/vcf/ingestion.py
sgillies: Here are a couple of examples:

* https://github.com/TileDB-Inc/Internal-Cloud-Py/blob/7a3f33097ee05ab14508ea0c4eb06d9fb5120d8f/src/tiledb/cloud/taskgraphs/server_executor/impl.py#L213
* https://github.com/TileDB-Inc/Internal-Cloud-Py/blob/7a3f33097ee05ab14508ea0c4eb06d9fb5120d8f/src/tiledb/cloud/soma/ingest.py#L112

alancleary: Thanks @sgillies. Looking at the logging code now, reworking all the messages in ingestion would require touching a bit of code unrelated to this PR so I think I'll do it as its own PR or as part of a manifest rework. Are there guidelines anywhere for this that I can reference in an issue?

sgillies: @alancleary I'm only asking about the new logging you've added. We can revisit the existing code later, I agree.

We don't have any guidelines about logging, but I'll add something to a contributing doc.

sgillies: @alancleary Yes, like that. In looking around for good examples on the web, I came across this one from Stripe: https://brandur.org/canonical-log-lines#what-are-they. "Literally one big log line".

I think you could leverage f-strings to simplify the above.

```python
logger.info(f"Filtering sample URIs: {len(dataset_samples)=}, {len(ingested_samples)=}, {len(partial_samples)=}, {len(manifest_samples)=}, {len(ready_samples)=}, {len(queued_samples)=}")
```

With the `=` syntax, you get, for example, `"Filtering sample URIs: len(dataset_samples)=100000, ..."`.

alancleary: > "Literally one big log line"

I like it! Also, I haven't encountered `=` syntax with f-strings before. Thanks for sharing!

src/tiledb/cloud/vcf/ingestion.py
```diff
@@ -1306,9 +1314,13 @@ def ingest_samples_dag(
...
 
-    logger.debug("partitions=%d, consolidates=%d", num_partitions, num_consolidates)
-    logger.debug("ingest_resources=%s", ingest_resources)
-    logger.debug("consolidate_resources=%s", CONSOLIDATE_RESOURCES)
+    logger.debug(
+        "Ingesting samples DAG configuration: "
+        f"{num_partitions=}, "
+        f"{num_consolidates=}, "
+        f"{ingest_resources=}, "
+        f"{CONSOLIDATE_RESOURCES=}"
```
sgillies, line 1322: Thank you @alancleary !

sgillies:
There are some conflicts to be resolved, too. After which, I'm fine with a merge. I've already approved.

PR #724, opened 2025-06-23
Set VCF record count to 0 when the index is bad

alancleary:
Previously the record value was left as `None` when the VCF index is bad. The `None` value is not compatible with the manifest's "records" attribute, resulting in an error when writing:
```console
ValueError: NumPy array conversion check failed for attr 'records'
```

### DIFF
--- a/src/tiledb/cloud/vcf/ingestion.py
+++ b/src/tiledb/cloud/vcf/ingestion.py
@@ -715,6 +715,7 @@ def ingest_manifest_udf(
                         if records is None:
                             status = "" if status == "ok" else status + ","
                             status += "bad index"
+                            records = 0
 
                     keys.append(sample_name)
                     values["status"].append(status)


src/tiledb/cloud/vcf/ingestion.py
```diff
@@ -715,6 +715,7 @@ def file_size(uri: str) -> int:
                         if records is None:
                             status = "" if status == "ok" else status + ","
                             status += "bad index"
+                            records = 0
```
sgillies, line 718: @alancleary I may be missing something, but is this working around a defect of `get_record_count()`? Should that function be returning `0` instead of `None`?

alancleary: The return description of `get_record_count()` is "record count or None if there is an error". So `None` is expected since the bad index prevents the function from running correctly. I think that makes sense because there's clear a distinction between valid outputs and errors, which is used to set the manifest status appropriately. That said, I can't imagine a user will ever have data where the record count is 0!

sgillies: I get it.

sgillies (approved):

PR #729, opened 2025-08-25
Add consolidation and manifest resource parameters for VCF ingestion

Shelnutt2:
This adds two parameters to the VCF ingestion call stack for handling overriding the default consolidation and manifest resources. Previously these were module level defaults. It was unexpected to users that we didn't have parameters for them to pass directly.


alancleary (changes requested):
New parameters need to be added to the docstring comment for `ingest_manifest_dag(...)`. Otherwise everything looks good.

Shelnutt2:
> New parameters need to be added to the docstring comment for `ingest_manifest_dag(...)`. Otherwise everything looks good.

Oops, good catch. Docstrings added!

alancleary (approved):

PR #730, opened 2025-09-05
Add support for passing resources for all tasks in VCF Ingestion

Shelnutt2:
This allows users to override resources for every task in the VCF ingestion process. We've found at scale some different task can fail with OOM. Specifically we latest ran into OOM with filter VCF URIs because the S3 path had a significant amount of files that we had to filter out.


ckrilow (approved):
Looks good to me, thanks!

spirpinias (approved):

alancleary (approved):
Looks good. Thanks for the changes!

PR #736, opened 2026-03-12
VCF-54: Backport TileDB-Client-Py #161: Load samples names via Graph/UDF

alancleary:
This has yet to be tested but the code is essentially the same; the only explicit differences are:
1. Changing teamspaces (Client-Py) to namespaces (Cloud-Py) and
2.  Dropping the `image_name` parameters (Client-Py)

### DIFF
--- a/src/tiledb/cloud/vcf/query.py
+++ b/src/tiledb/cloud/vcf/query.py
@@ -2,6 +2,7 @@
 
 import functools
 import logging
+import uuid
 from math import ceil
 from typing import Any, Callable, List, Mapping, Optional, Sequence, Tuple, Union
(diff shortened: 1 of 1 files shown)


sgillies (commented):
> Changing teamspaces (Client-Py) to workspaces (Cloud-Py) and

@alancleary I didn't see any such change in the changeset itself. Is that expected? Cloud-py doesn't have workspaces, it has namespaces.

leipzig (approved):

spirpinias (approved):
looks fine. will test whenever made available.

sgillies (approved):

